PR #5, opened 2025-03-19
feat: Logger

chris-4chain:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter


cmd/infra/main.go
```diff
@@ -1,23 +1,29 @@
...
 
 	"github.com/4chain-ag/go-wallet-toolbox/pkg/infra"
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/logging"
 )
 
 func main() {
+	rootLogger := logging.New().
+		WithLevel(slog.LevelDebug).
+		WithHandler(logging.TextHandler, os.Stdout).
+		Logger()
```
dorzepowski, line 15: I thought that what you're going to do, is to configure logger based on the configuration provided by user, not manually.

pkg/logging/configurer.go
```diff
@@ -0,0 +1,82 @@
...
+type HandlerType string
+
+// Supported handler types.
+const (
+	JSONHandler HandlerType = "json"
+	TextHandler HandlerType = "text"
+)
+
+// Configurer is the main interface for configuring a logger.
+type Configurer interface {
```
dorzepowski, line 18: What I'm missing is simple function to configure logger based on the Logger config, rather then some special builder.
Eventually you can add to config possibility to set custom handler.

pkg/infra/server.go
```diff
@@ -39,25 +45,29 @@ func NewServer(opts ...InitOption) (*Server, error) {
...
-		Addr:              fmt.Sprintf(":%d", s.Config.HTTPConfig.Port),
+		Addr:              fmt.Sprintf(":%d", port),
 		Handler:           mux,
 		ReadHeaderTimeout: 3 * time.Second,
 		ReadTimeout:       10 * time.Second,
 		WriteTimeout:      10 * time.Second,
 		IdleTimeout:       30 * time.Second,
 	}
 
+	logging.Sprintf(s.logger, slog.LevelInfo, "Listening on :%d", port)
```
dorzepowski, line 70: I'm not convinced that it's much more readable then
```suggestion
	s.logger.Info(fmt.Sprintf("http server started on port: %d", port))
```

pkg/infra/server.go
```diff
@@ -39,25 +45,29 @@ func NewServer(opts ...InitOption) (*Server, error) {
...
 		Config: cfg,
+
+		logger: logger,
 	}, nil
 }
 
 // ListenAndServe starts the JSON-RPC server
 func (s *Server) ListenAndServe() error {
-	rpcServer := server.NewRPCHandler()
+	rpcServer := server.NewRPCHandler(logging.Child(s.logger, "rpc_server"))
```
dorzepowski, line 55: IMHO this is a responsibility of NewXyz to create a Child, this way, it can name it as allways the same, despite of the place where it's used.

pkg/logging/utils.go
```diff
@@ -0,0 +1,40 @@
...
+	"os"
+)
+
+const (
+	// ServiceKey is the key used to store the service name in the logger attrs.
+	ServiceKey = "service"
+)
+
+// Child returns a new logger with the given service name added to the logger attrs.
+func Child(logger *slog.Logger, serviceName string) *slog.Logger {
```
dorzepowski, line 16: I thought that maybe it is not needed,
But if we add a line:
`logger := DefaultIfNil(logger)`

Then it would be much more useful util.

pkg/logging/utils.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+// Sprintf logs a message with the given level and arguments.
+func Sprintf(logger *slog.Logger, level slog.Level, format string, args ...any) {
+	logger.Log(context.Background(), level, fmt.Sprintf(format, args...))
+}
+
+// Fatalf logs the error and exits the program.
+func Fatalf(logger *slog.Logger, err error, msg string, args ...any) {
+	logger.Error("Fatal error: "+fmt.Sprintf(msg, args...), slog.String("error", err.Error()))
```
dorzepowski, line 29: maybe we could have slog.String("error", err.Error()) wrapped as function `logging.Error(err)`

pkg/server/rpc_server.go
```diff
@@ -10,25 +11,32 @@ import (
...
+	)
 
 	// create a handler instance and register it
 	serverHandler := &Handler{}
 	rpcServer.Register("Handler", serverHandler)
 
 	return &RPCServer{
 		Handler: rpcServer,
+
+		logger: logger,
```
dorzepowski, line 32: why to put this logger here?

pkg/infra/server.go
```diff
@@ -21,6 +25,8 @@ func NewServer(opts ...InitOption) (*Server, error) {
 		option(&options)
 	}
 
+	logger := logging.NopIfNil(options.Logger)
```
dorzepowski, line 28: Here we should use, logging Child

pkg/infra/options.go
```diff
@@ -1,16 +1,22 @@
...
+
 // Options is the parameters for initializing the "infra" server
 type Options struct {
 	EnvPrefix  string
 	ConfigFile string
+	Logger     *slog.Logger
 }
 
 // DefaultOptions returns the default parameters to initialize the "infra" server
 func DefaultOptions() Options {
```
dorzepowski, line 15: I know that it's not from this PR, but does it need to be public?

cmd/infra/main.go
chris-4chain: Check it out right now :)

pkg/logging/utils.go
```diff
@@ -0,0 +1,40 @@
...
+	os.Exit(1)
+}
+
+// NopIfNil returns a new NOP logger if the given logger is nil, otherwise returns the given logger.
+// NOTE: NOP logger discards all logs.
+func NopIfNil(logger *slog.Logger) *slog.Logger {
```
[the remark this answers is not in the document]
chris-4chain: ✅ Done

pkg/logging/configurer.go
chris-4chain: Check it out right now.

pkg/infra/server.go
chris-4chain: I've removed this function

pkg/infra/server.go
chris-4chain: ✅ Done

pkg/logging/utils.go
chris-4chain: ✅ Done

pkg/server/rpc_server.go
chris-4chain: No reason - removed ✅

pkg/infra/server.go
chris-4chain: Check it out, how it looks like right now 🤞

pkg/infra/options.go
chris-4chain: It doesn't, I've made it not-exported ✅

cmd/infra/main.go
```diff
@@ -14,8 +12,6 @@ func main() {
 		panic(err)
 	}
 
-	fmt.Printf("Listening on :%d", server.Config.HTTPConfig.Port)
```
dorzepowski, line 17: Maybe this wasn't that bad, just use Logger instead

pkg/infra/server.go
```diff
@@ -37,31 +42,52 @@ func NewServer(opts ...InitOption) (*Server, error) {
...
 
 	return nil
 }
+
+func makeLogger(cfg *Config, options *Options) *slog.Logger {
+	if options.Logger != nil {
+		return options.Logger
+	}
+
+	if !cfg.Logging.Enabled {
```
dorzepowski, line 85: maybe `disabled` would be better, as this would make the zero value also our suggested default.

Or maybe this isn't necessary at all, as you can simply provide a logger in options.

pkg/internal/logging/utils.go
```diff
@@ -0,0 +1,46 @@
...
+// Child returns a new logger with the given service name added to the logger attrs.
+func Child(logger *slog.Logger, serviceName string) *slog.Logger {
+	return DefaultIfNil(logger).With(
+		slog.String(ServiceKey, serviceName),
+	)
+}
+
+func Errorf(logger *slog.Logger, err error, format string, args ...any) {
+	logger.Error(fmt.Sprintf(format, args...), slog.String("error", err.Error()))
+}
```
dorzepowski, line 32: This is not what I meant. 
My suggestion was rather to provide wrapper over slog.String("error", err) 
something like:
```suggestion
func Error(err error) Args {
	return slog.String(ErrorKey, err.Error()))
}
```

dorzepowski (approved):

pkg/infra/server.go
chris-4chain: There is a default config which makes logger enabled. 
So it would not behave as you described (about zero-value)

pkg/internal/logging/utils.go
chris-4chain: ✅ Done

PR #48, opened 2025-03-27
feat: storage.createAction - validate args

chris-4chain:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter


pkg/storage/internal/actions/create.go
```diff
@@ -22,6 +22,14 @@ type FundingResult struct {
 	Fee            uint64
 }
 
+type CreateActionParams struct {
+}
+
+func FromValidCreateActionArgs(args *wdk.ValidCreateActionArgs) CreateActionParams {
+	// TODO: use only the necessary fields (no redundant fields)
+	return CreateActionParams{}
```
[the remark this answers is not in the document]
chris-4chain: We will do it gradually in subsequent PRs

pkg/internal/validate/valid_create_action_args.go
```diff
@@ -0,0 +1,126 @@
...
+	if args.IsSignAction != deducedIsSignAction {
+		return fmt.Errorf("inconsistent IsSignAction with IsNewTx and Options.SignAndProcess and Inputs.UnlockingScript")
+	}
+
+	deducedIsDelayed := bool(args.Options.AcceptDelayedBroadcast)
+	if args.IsDelayed != deducedIsDelayed {
+		return fmt.Errorf("inconsistent IsDelayed with Options.AcceptDelayedBroadcast")
+	}
+
+	deducedIsNoSend := bool(args.Options.NoSend)
```
dorzepowski, line 39: ### minor
because BooleanDefaultTrue and BooleanDefaultFalse aren't the types that would bring any benefit of being standalone types, maybe it would be better to change them to type aliases:
```go
// BooleanDefaultTrue is a boolean with a default value of true
type BooleanDefaultTrue = bool

// BooleanDefaultFalse is a boolean with a default value of false
type BooleanDefaultFalse = bool
```

this way, the casting to bool would be redundant.

pkg/internal/validate/valid_create_action_args_test.go
```diff
@@ -0,0 +1,216 @@
...
+	tests := map[string]struct {
+		args wdk.ValidCreateActionArgs
+	}{
+		"IsSendWith is set even though there is no 'send-with' txs in options": {
+			args: func() wdk.ValidCreateActionArgs {
+				args := fixtures.DefaultValidCreateActionArgs()
+				args.IsSendWith = true
+				args.Options.SendWith = []wdk.TXIDHexString{}
+				return args
+			}(),
```
dorzepowski, line 35: 😕 eeee what's the point of having this in function that is called just after declaration?

Maybe a better idea would be to accept an args as and input and call this function with the default args in the test?

```suggestion
			args: func(args wdk.ValidCreateActionArgs) wdk.ValidCreateActionArgs {
				args.IsSendWith = true
				args.Options.SendWith = []wdk.TXIDHexString{}
				return args
			},
```

dorzepowski (changes requested):

pkg/internal/validate/valid_create_action_args.go
dzolt-4chain: Now that you mention it I think we don't need it anyways? In TS the way it works it will default to true if undefined but in golang I belie we won't get the "undefined" or nil unless its a pointer?

pkg/internal/validate/valid_create_action_args_test.go
chris-4chain: They were just generator functions which produced args - they were capable of generating "any" args.
- Your suggestion modifies the default args, provided once in the test code. 
- Because - for now - all off the tests rely on the default config - I've applied your suggestion ✅ 
- But I don't think that it was serious enough to hit "change request" - just another approach.

dorzepowski (approved):

PR #60, opened 2025-04-02
feat: handle funding without change cases in funder

dorzepowski:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter


pkg/storage/internal/actions/funder/fee_calculator.go
```diff
@@ -0,0 +1,48 @@
...
+	return &feeCalc{
+		value: feeValue,
+		bytes: 1000,
+	}
+}
+
+func (f *feeCalc) Calculate(txSize uint64) (uint64, error) {
+	size, err := to.Float64FromUnsigned(txSize)
+	if err != nil {
+		return 0, fmt.Errorf("invalid transaction size: %s", err.Error())
```
dzolt-4chain:
```suggestion
		return 0, fmt.Errorf("invalid transaction size: %s", err)
```

Shouldn't it be like this here?

pkg/storage/internal/database/models/user_utxo.go
```diff
@@ -0,0 +1,17 @@
...
+	UserID   int    `gorm:"primaryKey;uniqueIndex:idx_window,sort:asc,priority:1"`
+	TxID     string `gorm:"primaryKey;uniqueIndex:idx_window,sort:asc,priority:4"`
+	Vout     uint32 `gorm:"primaryKey;uniqueIndex:idx_window,sort:asc,priority:5"`
+	Satoshis uint64
+	// EstimatedInputSize is the estimated size increase when adding and unlocking this UTXO to a transaction.
+	EstimatedInputSize uint64
+	Basket             string
+	CreatedAt          time.Time `gorm:"uniqueIndex:idx_window,sort:asc,priority:3"`
+	// TouchedAt is the time when the UTXO was last touched (selected for preparing transaction outline) - used for prioritizing UTXO selection.
+	TouchedAt time.Time `gorm:"uniqueIndex:idx_window,sort:asc,priority:2"`
```
dzolt-4chain: isn't it the same as updated_at, or can't we use gorms updated at field?

pkg/storage/internal/actions/funder/fee_calculator.go
dorzepowski: my bad, it should be %w

pkg/storage/internal/database/models/user_utxo.go
dorzepowski: this is still work in progress,
I will remove unneeded fields later.
I would like to ask you, to not bother to much about this struct for now.

As a side note:
I'm still considering introduction of the Funder with single sql query,
but first I want to finish in code version, and then I will have a bunch of tests that would give me the answer that the Plain SQL version is also valid, and those fields was introduced in sql version.

dorzepowski: Or maybe you're right, I will remove unneeded fields

pkg/storage/internal/actions/funder/sql.go
```diff
@@ -32,5 +44,132 @@ func NewSQL(logger *slog.Logger, db *gorm.DB, feeModel defs.FeeModel) *SQL {
...
+		return nil, fmt.Errorf("failed to fund transaction: %w", err)
+	}
+	return result, nil
+
+}
+
+func (f *SQL) loadUTXOs(ctx context.Context, userID int) iter.Seq2[*models.UserUTXO, error] {
+	batches := seqerr.ProduceWithArg(
+		func(page *paging.Page) ([]*models.UserUTXO, *paging.Page, error) {
+			utxos, err := f.utxoRepository.FindAllUTXOs(ctx, userID, page)
```
chris-4chain, line 72: What if GORM returns an error that there is empty result?

pkg/storage/internal/actions/funder/sql.go
```diff
@@ -32,5 +44,132 @@ func NewSQL(logger *slog.Logger, db *gorm.DB, feeModel defs.FeeModel) *SQL {
...
+			if err != nil {
+				return nil, nil, fmt.Errorf("failed to load utxos: %w", err)
+			}
+			page.Next()
+			return utxos, page, nil
+		},
+		&paging.Page{
+			Limit:  1000,
```
chris-4chain: It could be a constant

pkg/storage/internal/actions/funder/sql.go
```diff
@@ -32,5 +44,132 @@ func NewSQL(logger *slog.Logger, db *gorm.DB, feeModel defs.FeeModel) *SQL {
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to allocate utxo: %w", err)
+		}
+
+		if collector.IsFunded() {
+			break
+		}
+	}
+
+	return collector.GetResult()
```
chris-4chain, line 104: Consider this - maybe it would more straightforward. (And it doesn't need to call IsFunded twice)
```suggestion
		if collector.IsFunded() {
			return collector.result, nil
		}
	}

	return nil, errfunder.NotEnoughFunds
```

pkg/storage/internal/actions/funder/sql.go
```diff
@@ -32,5 +44,132 @@ func NewSQL(logger *slog.Logger, db *gorm.DB, feeModel defs.FeeModel) *SQL {
...
+
+func (c *utxoCollector) Allocate(utxo *models.UserUTXO) (err error) {
+	c.addToAllocated(utxo)
+
+	err = c.increaseSize(utxo.EstimatedInputSize)
+	if err != nil {
+		return fmt.Errorf("failed to increase tx size: %w", err)
+	}
+
+	c.increaseValue(utxo.Satoshis)
```
chris-4chain, line 139: For this one-liner, I think, the version without additional function would be more readable:

```suggestion
	c.satsCovered += satoshis
```

pkg/storage/internal/actions/funder/sql_test.go
```diff
@@ -0,0 +1,138 @@
...
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/storage/internal/testabilities/testusers"
+)
+
+const smallTransactionSize = 44
+const desiredUTXONumberToPreferSingleChange = 1
+const testDesiredUTXOValue = 1000
+
+var ctx = context.Background()
+
+func TestFunderSQLFundSuccessResult(t *testing.T) {
```
chris-4chain, line 17: Do you plan to test a case in which:
- adding change output increases the fee and the change output 
    - a) decreases its value
    - b) decreases to zero

pkg/storage/internal/database/models/consts.go
```diff
@@ -0,0 +1,9 @@
+package models
+
+// EstimatedInputSizeForP2PKH is the estimated size increase when adding and unlocking P2PKH input to transaction.
+// 32 bytes txID
+// + 4 bytes vout index
+// + 1 byte script length
+// + 107 bytes script pub key
+// + 4 bytes nSequence
+const EstimatedInputSizeForP2PKH = 148
```
chris-4chain, line 9: Shouldn't it be in `wdk` package?

pkg/storage/internal/actions/funder/sql.go
dorzepowski: it's weird but it doesn't, the tests are covering this situation.

pkg/storage/internal/actions/funder/sql.go
dorzepowski: I already have next steps, and later I need to calculate change and change count, I'm not sure yet if I want to do that during processing, that's why I prefer to keep this method.

pkg/storage/internal/actions/funder/sql.go
dorzepowski: What you propose is mixing the levels of abstraction,
that's why I introduce this method, and I would rather keep it, as is

pkg/storage/internal/actions/funder/sql_test.go
dorzepowski: As I said, it's still work in progress,
there are still many test cases to check, and several of them are about creation of changes.
So yes, I'm going to have such test cases.

pkg/storage/internal/database/models/consts.go
dorzepowski: We will see, when we will really store it in db. For now, it's in the same place as it was in spv-wallet.

pkg/storage/internal/actions/funder/sql.go
chris-4chain: I think, this is because you use (on the repo level) GORM method: `Find` which does not return error for empty result.

chris-4chain (approved):

dzolt-4chain (approved):

PR #61, opened 2025-04-02
feat: add validation for certificates actions

dzolt-4chain:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter


pkg/internal/validate/validate_insert_certificate_auth_args.go
```diff
@@ -0,0 +1,55 @@
+package validate
+
+import (
+	"fmt"
+
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/wdk"
+)
+
+func ValidateInsertCertificateAuthArgs(args *wdk.TableCertificateX) error {
```
chris-4chain: Name starts with the package name 

For my args I named the function as the argument name.
So in this case it would be:
```golang
package validate 

func TableCertificateX(args *wdk.TableCertificateX) error

// so from outside the package it would be called as:
func main() {
   //...
   err := validate.TableCertificateX(args)
}
```

pkg/wdk/primitives.go
```diff
@@ -23,18 +26,83 @@ func (d String5to2000Bytes) Validate() error {
...
+	// Step 1: Check if the string's length is divisible by 4 (Base64 requirement)
+	if len(s)%4 != 0 {
+		return fmt.Errorf("base64 string length must be divisible by 4")
+	}
+
+	// Step 2: Validate padding
+	paddingCount := strings.Count(string(s), "=")
+	if paddingCount > 2 || (paddingCount > 0 && !strings.HasSuffix(string(s), strings.Repeat("=", paddingCount))) {
+		return fmt.Errorf("invalid base64 padding")
+	}
```
dorzepowski: hmmm, wouldn't it be enough to just do it, this way:
```suggestion
	if strings.HasSuffix(string(s), "===") {
		return fmt.Errorf("invalid base64 padding")
	}
```

if it has padding of 3 or more "=" then it should have suffix of "===" so it looks like good enough check 🙂

pkg/wdk/primitives.go
```diff
@@ -23,18 +26,83 @@ func (d String5to2000Bytes) Validate() error {
...
+
+	// Step 3: Check if the string is valid Base64
+	_, err := base64.StdEncoding.DecodeString(string(s))
+	if err != nil {
+		return fmt.Errorf("invalid base64 string")
+	}
```
chris-4chain, line 53: Wouldn't be enough to just try to decode? (and skip step 1 and 2)

... or since this is only validation, and it could be memory consuming, we should not decode it at all 🤔 - and rely on step 1 and 2 (or maybe additional regex)

pkg/wdk/primitives.go
```diff
@@ -23,18 +26,83 @@ func (d String5to2000Bytes) Validate() error {
...
+		return fmt.Errorf("no more than 50 length")
+	}
+	return nil
+}
+
 // PubKeyHex is a compressed DER secp256k1 public key, exactly 66 hex characters (33 bytes) in length.
 type PubKeyHex HexString
 
+// Validate checks if the string is valid pubkey hexadecimal string
+func (pkh PubKeyHex) Validate() error {
```
chris-4chain, line 103: To be honest, I would remove the `PubKeyHex` at all - and change all its occurrences to jus `HexString`

pkg/wdk/primitives.go
```diff
@@ -87,6 +155,10 @@ const MaxSatoshis = 2100000000000000
 
 // Validate checks if the value is less than the maximum number of Satoshis
 func (s SatoshiValue) Validate() error {
+	if s < 1 {
+		return fmt.Errorf("at least %d", MaxSatoshis)
```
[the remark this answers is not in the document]
chris-4chain: Why? I think sometimes `0` satoshi is totally valid.

Despite this, I think the message is wrong - as AI found 😄 
Good Job Copilot 🙈

pkg/internal/fixtures/default_valid_insert_certificate_auth_action_args.go
chris-4chain, line 1: I know that this is internal (and it does not have a requirement for a comment)
But... could you adjust in a way to resolve my IDE errors 🙈 ?

image

pkg/wdk/primitives.go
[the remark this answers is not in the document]
dzolt-4chain: Idk but in their code satoshivalue is minimum 1 max the large number so I will leave it as 1 min

pkg/internal/validate/validate_insert_certificate_auth_args.go
dzolt-4chain: Allright, applied changes

pkg/wdk/primitives.go
```diff
@@ -120,6 +192,23 @@ type BEEF []byte
...
+// Validate checks if the string is proper outpoint string and contains outpoint index after "."
+func (s OutpointString) Validate() error {
+	split := strings.Split(string(s), ".")
+
+	if len(split) != 2 {
+		return fmt.Errorf("txid as hexstring and numeric output index joined with '.'")
+	}
+
+	// check if after decimal point there is an outpoint index
+	_, err := strconv.Atoi(split[1])
```
dorzepowski:
```suggestion
	_, err := to.Uint64FromString(split[1])
```
hopefully this will also ensure it's not negative number (If not, then we will need to fix need to fix it in `common` 😄

pkg/wdk/primitives.go
dzolt-4chain: Hmm yeah I tried and test pass without step 1 and 2, I'll remove them but i would leave the decode try as it actually handles most important logic I believe

pkg/internal/fixtures/default_valid_insert_certificate_auth_action_args.go
dzolt-4chain: sure :D

pkg/wdk/primitives.go
dzolt-4chain: I added check for pubkey length aswell and decided to keep this one

chris-4chain (approved):

pkg/wdk/primitives.go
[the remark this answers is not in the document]
chris-4chain: Show me what code you referring to ?
Initially I created this type & validation for `createAction` sake and the `min` was not set.

image

pkg/wdk/primitives.go
```diff
@@ -87,6 +160,10 @@ const MaxSatoshis = 2100000000000000
 
 // Validate checks if the value is less than the maximum number of Satoshis
 func (s SatoshiValue) Validate() error {
+	if s < 1 {
```
chris-4chain: I'm blocking the PR until we resolve above conversation about the min value ☝️

chris-4chain (changes requested):

pkg/wdk/primitives.go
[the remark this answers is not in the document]
dzolt-4chain: image
I was referring to this code

pkg/wdk/primitives.go
dzolt-4chain: removed < 1 constraint

chris-4chain (approved):

pkg/storage/methodtests/insert_certificate_auth_test.go
```diff
@@ -117,7 +117,9 @@ func TestInsertCertificateAuth(t *testing.T) {
 
 		// then:
 		require.NoError(t, err)
-		require.EqualValues(t, expectedResult, certs)
+		require.Equal(t, expectedResult.TotalCertificates, wdk.PositiveIntegerOrZero(2))
+		require.Contains(t, certs.Certificates, expectedResult.Certificates[0])
+		require.Contains(t, certs.Certificates, expectedResult.Certificates[1])
```
dorzepowski: ### minor
you can try `require.MatchElements(t, certs.Certificates, expectedResult.Certificates)

pkg/storage/methodtests/insert_certificate_auth_test.go
```diff
@@ -117,7 +117,9 @@ func TestInsertCertificateAuth(t *testing.T) {
 
 		// then:
 		require.NoError(t, err)
-		require.EqualValues(t, expectedResult, certs)
+		require.Equal(t, expectedResult.TotalCertificates, wdk.PositiveIntegerOrZero(2))
```
dorzepowski: Maybe it's better tu use `assert` here, as it will allow to check everything at once if there wasn't an error, instead of failing on this line first, and then after fixing, failing on the next line.

pkg/wdk/constants.go
```diff
@@ -3,4 +3,9 @@ package wdk
 const (
 	// BasketNameForChange is the name of the output basket that is used to store "change" outputs
 	BasketNameForChange = "default"
+
+	// PubKeyBytesLenCompressed is the length of the compressed pub key
+	PubKeyBytesLenCompressed = 33
+	// PubKeyBytesLenUncompressed is the length of the uncompressed pub key
+	PubKeyBytesLenUncompressed = 65
```
dorzepowski: those values doesn't need to be public, and actually should be placed in the same file as the validation that use it.

pkg/wdk/primitives.go
```diff
@@ -94,6 +94,10 @@ type PubKeyHex HexString
 
 // Validate checks if the string is valid pubkey hexadecimal string
 func (pkh PubKeyHex) Validate() error {
+	if len(pkh)/2 != PubKeyBytesLenCompressed && len(pkh)/2 != PubKeyBytesLenUncompressed {
```
dorzepowski: maybe calculate `len(pkh)/2` once 
and maybe add the comment why is this counted like that 🙈

dorzepowski (approved):

PR #65, opened 2025-04-03
feat: handle funding transaction by user

dorzepowski:
Next part of funder implementation.

Feature: handle situation when user will provide external inputs and will cover the transaction outputs and eventually fee.

### DIFF
--- a/pkg/storage/internal/actions/funder/sql.go
+++ b/pkg/storage/internal/actions/funder/sql.go
@@ -45,52 +45,24 @@ func NewSQL(logger *slog.Logger, utxoRepository UTXORepository, feeModel defs.Fe
 // @param numberOfDesiredUTXOs - the number of UTXOs in basket #TakeFromBasket
 // @param minimumDesiredUTXOValue - the minimum value of UTXO in basket #TakeFromBasket
 // @param userID - the user ID.
-func (f *SQL) Fund(ctx context.Context, targetSat int64, currentTxSize int64, numberOfDesiredUTXOs int, minimumDesiredUTXOValue uint64, userID int) (*actions.FundingResult, error) {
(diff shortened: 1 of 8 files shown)


dzolt-4chain (approved):

pkg/storage/internal/actions/funder/sql.go
```diff
@@ -161,17 +163,41 @@ func (c *utxoCollector) addToAllocated(utxo *models.UserUTXO) {
...
+	if err != nil {
+		return nil, fmt.Errorf("cannot convert change amount to uint64: %w", err)
+	}
+
+	return &actions.FundingResult{
+		AllocatedUTXOs: c.allocatedUTXOs,
+		Fee:            fee,
+		ChangeAmount:   changeAmount,
+		ChangeCount:    to.IfThen(changeAmount == 0, 0).ElseThen(1),
+	}, nil
```
chris-4chain, line 202: To be honest, for such a simple condition I would not introduce ternary-operator-like custom method.

```suggestion
	changeCount := 0
	if changeAmount > 0 {
		changeCount = 1
	}

	return &actions.FundingResult{
		AllocatedUTXOs: c.allocatedUTXOs,
		Fee:            fee,
		ChangeAmount:   changeAmount,
		ChangeCount:    changeCount,
	}, nil
```

chris-4chain (approved):

pkg/storage/internal/actions/funder/sql.go
dorzepowski: I will take this into account in the final solution,
for now, it's not handling splitting change, so for sure this part will change soon when I finally will tackle the problem of change splitting.

PR #72, opened 2025-04-07
feat: BsvExchangeRate from WhatsOnChain service 

dzolt-4chain:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter


pkg/wdk/services.interface.go
```diff
@@ -0,0 +1,79 @@
+package wdk
+
+import (
+	"github.com/bsv-blockchain/go-sdk/transaction"
+	"github.com/bsv-blockchain/go-sdk/transaction/chaintracker"
+)
+
+// WalletServices is an interface to access functionality implemented by external transaction processing services
+type WalletServices interface {
```
chris-4chain: Why do we need this interface?
Where will it be used?

dzolt-4chain: In TS it is used inside wallet as a services property, so I wanted to make it similiar to their code base, maybe we won't need it as an interface here. But it's implementation is inside services folder

go.mod
dorzepowski, line 1: ##### putting this comment here, just to enable option to resolve it

Please change the PR title - because it will appear in Release
(and remember to make a squash)

pkg/wdk/primitives.go
```diff
@@ -232,3 +232,10 @@ type AuthID struct {
 	UserID      *int
 	IsActive    *bool
 }
+
+// ReqHistoryNote is the history representation of the request
+type ReqHistoryNote struct {
```
dorzepowski: Is this something that the client of storage will use?
if not, then maybe services package is better place for it.

pkg/services/services_options.go
```diff
@@ -0,0 +1,164 @@
...
+
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/defs"
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/services/internal/whatsonchain"
+)
+
+// Options is a function that can be used to override services options
+type Options = func(*WalletServicesOptions)
+
+// WalletServicesOptions is a struct that has options for wallet services
+type WalletServicesOptions struct {
```
dorzepowski: I would make it a simply config struct, that can be passed to the constructor.
And I would add the chain also her.

Why?
Because we would probably like to expose those options in server config yaml.
So in current version:
we will need to parse the yaml to some struct,
then we would need to change each field from config to the Option
We've been there (spv-wallet) and it's harder to use, then simply provide a struct.

But maybe you have some idea how to do that in a smarter way?

pkg/services/services.go
```diff
@@ -0,0 +1,148 @@
...
+	return &WalletServices{
+		httpClient:   httpClient,
+		chain:        chain,
+		options:      options,
+		logger:       logger,
+		whatsonchain: whatsonchain.New(httpClient, logging.Child(logger, "service_whats_on_chain"), options.WhatsOnChainApiKey, chain),
+	}
+}
+
+// ChainTracker returns service which requires `options.chaintracks` be valid.
```
dorzepowski: ### minor
```suggestion
// ChainTracker returns service, which requires `options.chaintracks` be valid.
```

pkg/services/services.types.go
dorzepowski, line 1: ### minor
to be aligned with the go convention, the name of the file should be rather `services_types.go`

pkg/services/services_options.go
```diff
@@ -0,0 +1,164 @@
...
+		//nolint:gosec
+		taalApiKey = "testnet_0e6cf72133b43ea2d7861da2a38684e3"
+		port = 8083
+		arcUrl = "https://arc-test.taal.com/arc"
+	}
+
+	return &WalletServicesOptions{
+		TaalApiKey: taalApiKey,
+		BsvExchangeRate: whatsonchain.BSVExchangeRate{
+			Timestamp: time.Date(2023, time.December, 13, 0, 0, 0, 0, time.UTC),
```
dorzepowski: ### Just thinking.
maybe we don't need to set those values by default, as they are already deprecated.

pkg/services/services_options.go
```diff
@@ -0,0 +1,164 @@
...
+			Base:      "USD",
```
dorzepowski: Why you're not using the Currency const/type?

pkg/services/services_options.go
```diff
@@ -0,0 +1,164 @@
...
+			Rate:      47.52,
+		},
+		BsvUpdateMsecs: 1000 * 60 * 15, // 15 minutes
```
dorzepowski: maybe we could use a Duration here?

pkg/services/services_options.go
```diff
@@ -0,0 +1,164 @@
...
+		FiatExchangeRates: FiatExchangeRates{
+			Timestamp: time.Date(2023, time.December, 13, 0, 0, 0, 0, time.UTC),
+			Base:      "USD",
+			Rates: map[string]float64{
+				"USD": 1,
+				"GBP": 0.8,
+				"EUR": 0.93,
+			},
```
dorzepowski: Why not const of Currency type?

pkg/services/services_options.go
```diff
@@ -0,0 +1,164 @@
...
+		},
+		FiatUpdateMsecs:                 1000 * 60 * 60 * 24, // 24 hours
```
dorzepowski: again, maybe we could use Duration.

pkg/services/services_options.go
```diff
@@ -0,0 +1,164 @@
...
+		DisableMapiCallback:             true,                // rely on WalletMonitor by default
+		ExchangeratesApiKey:             "bd539d2ff492bcb5619d5f27726a766f",
+		ChaintracksFiatExchangeRatesUrl: fmt.Sprintf("https://npm-registry.babbage.systems:%d/getFiatExchangeRates", port),
```
dorzepowski: this looks like another component that we should rewrite to bsv-blockchain.

Can you please add an issue for checking this.

pkg/services/services_test.go
```diff
@@ -0,0 +1,127 @@
...
+	"time"
+
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/services"
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/services/internal/whatsonchain"
+	"github.com/go-resty/resty/v2"
+	"github.com/jarcoal/httpmock"
+	"github.com/stretchr/testify/assert"
+)
+
+func newMockHTTPClient(status int, content string, err error) *resty.Client {
```
dorzepowski: If you don't mind, I would prefer usage of testabilities pattern here.
Where you could have something like that:
```go
// given:
given := testabilities.Given(t)

given.WhatsOnChain().WillRespondWithRates(...)

services := given.NewServices()

// when:
result, err := services.BsvExchangeRate()

// then:
// ... assertions

```

pkg/services/services_options.go
```diff
@@ -0,0 +1,164 @@
...
+
+// Options is a function that can be used to override services options
+type Options = func(*WalletServicesOptions)
+
+// WalletServicesOptions is a struct that has options for wallet services
+type WalletServicesOptions struct {
+	TaalApiKey                      string
+	BitailsApiKey                   *string
+	WhatsOnChainApiKey              string
+	BsvExchangeRate                 whatsonchain.BSVExchangeRate
```
dorzepowski: looks like some of them are really the options, and others are rather the configurations.
This for example is option for sure, because I hope that we will remove it soon in the future.

pkg/services/services_test.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	// transport.RegisterResponder("GET", fmt.Sprintf("%s/v1/tx/%s", arcURL, minedTxID), responder(http.StatusOK, `{
+	transport.RegisterResponder("GET", "https://api.whatsonchain.com/v1/bsv/test/exchangerate", responder(status, content))
+
+	return client
+}
+
+func TestUpdateBsvExchangeRateSuccess(t *testing.T) {
+	t.Run("returns cached exchange rate if within update threshold", func(t *testing.T) {
```
dorzepowski, line 13: Please add the task for setting up some cacheing component that would handle cache in memory or with any other tool (like redis)

pkg/services/services_test.go
```diff
@@ -0,0 +1,127 @@
...
+		// given:
+		httpClient := newMockHTTPClient(200, "", nil)
```
dorzepowski: IMHO, it should return some different value then expected, so it would be clearer that it's not called here.

pkg/services/internal/whatsonchain/service.go
```diff
@@ -0,0 +1,79 @@
...
+		httpClient: httpClient,
+		apiKey:     apiKey,
+		url:        fmt.Sprintf("https://api.whatsonchain.com/v1/bsv/%s", network),
+		logger:     logger,
+	}
+}
+
+func (woc *WhatsOnChain) UpdateBsvExchangeRate(exchangeRate *BSVExchangeRate, bsvUpdateMsecs *int) (BSVExchangeRate, error) {
+	if exchangeRate != nil {
+		fifteenMinutesMs := 1000 * 60 * 15
```
dorzepowski: it's rather a const.

BTW. you could make it clearer by calculating it like that:
```suggestion
		fifteenMinutesMs := 15 * time.Minute.Milliseconds()
```

pkg/services/internal/whatsonchain/service.go
```diff
@@ -0,0 +1,79 @@
...
+func New(httpClient *resty.Client, logger *slog.Logger, apiKey string, network defs.BSVNetwork) *WhatsOnChain {
+	return &WhatsOnChain{
+		httpClient: httpClient,
+		apiKey:     apiKey,
+		url:        fmt.Sprintf("https://api.whatsonchain.com/v1/bsv/%s", network),
+		logger:     logger,
+	}
+}
+
+func (woc *WhatsOnChain) UpdateBsvExchangeRate(exchangeRate *BSVExchangeRate, bsvUpdateMsecs *int) (BSVExchangeRate, error) {
```
dorzepowski: we should probably use `int64` instead of `int` everywhere, to make the code working on both (32 and 64 bit) systems in an expected way.

pkg/services/internal/whatsonchain/service.go
```diff
@@ -0,0 +1,79 @@
...
+	apiKey     string
+	logger     *slog.Logger
+}
+
+func New(httpClient *resty.Client, logger *slog.Logger, apiKey string, network defs.BSVNetwork) *WhatsOnChain {
+	return &WhatsOnChain{
+		httpClient: httpClient,
+		apiKey:     apiKey,
+		url:        fmt.Sprintf("https://api.whatsonchain.com/v1/bsv/%s", network),
+		logger:     logger,
```
chris-4chain, line 45: It should create a child logger, like this:

```golang
func NewSQL(logger *slog.Logger, utxoRepository UTXORepository, feeModel defs.FeeModel) *SQL {
	logger = logging.Child(logger, "funderSQL")
```

pkg/services/internal/whatsonchain/service.go
chris-4chain: This cloud be moved to const

pkg/services/internal/whatsonchain/service.go
chris-4chain: Why pointer to int?
I mean, I know that you wanted to make it as "default argument", but since you use it only here:
```go
bsvExchangeRate, err := s.whatsonchain.UpdateBsvExchangeRate(
  &s.options.BsvExchangeRate,
  &s.options.BsvUpdateMsecs,
)
```
... maybe the default thing is not necessary.

pkg/services/internal/whatsonchain/service.go
```diff
@@ -0,0 +1,79 @@
...
+		// Calculate the threshold time by subtracting updateMsecs from the current time
+		thresholdTime := time.Now().Add(-time.Duration(updateMscecs) * time.Millisecond)
+
+		// Check if the rate timestamp is newer than the threshold time
+		if exchangeRate.Timestamp.After(thresholdTime) {
+			return *exchangeRate, nil
+		}
+	}
+
+	for retry := 0; retry < 2; retry++ {
```
chris-4chain: number of retries could be a const value.

pkg/services/internal/whatsonchain/service.go
```diff
@@ -0,0 +1,79 @@
...
+		var exchangeRateResponse BSVExchangeRateResponse
+		req := woc.httpClient.R().
+			SetHeader("Accept", "application/json")
+		if woc.apiKey != "" {
+			req.SetHeader("Authorization", woc.apiKey)
+		}
```
chris-4chain: This general configuration could be moved to a separate method for this and future usage.

pkg/services/internal/whatsonchain/service.go
```diff
@@ -0,0 +1,79 @@
...
+
+		res, err := req.
+			SetResult(&exchangeRateResponse).
+			Get(fmt.Sprintf("%s/exchangerate", woc.url))
+		if err != nil {
+			return BSVExchangeRate{}, fmt.Errorf("failed to fetch exchange rate: %w", err)
+		}
+
+		if res.Status() == "Too Many Requests" && retry < 2 {
```
chris-4chain: 1. Shouldn't we compare to `StatusCode` ?

2. 

image

pkg/services/internal/whatsonchain/service.go
```diff
@@ -0,0 +1,79 @@
...
+			time.Sleep(2000 * time.Millisecond)
+			continue
+		}
+
+		if res.StatusCode() != http.StatusOK {
+			return BSVExchangeRate{}, fmt.Errorf("failed to retrieve successful response from WOC. Actual status: %d", res.StatusCode())
+		}
```
chris-4chain: This doesn't give a chance to make another retry 🤔

pkg/services/internal/whatsonchain/service.go
```diff
@@ -0,0 +1,79 @@
...
+
+		if exchangeRateResponse.Currency != "USD" {
+			return BSVExchangeRate{}, fmt.Errorf("unsupported currency returned from Whats On Chain")
+		}
+
+		return BSVExchangeRate{
+			Timestamp: time.Now(),
+			Base:      "USD",
```
chris-4chain: Consider constant or enum for that "currency"

pkg/services/internal/whatsonchain/service.go
```diff
@@ -0,0 +1,79 @@
...
+		var exchangeRateResponse BSVExchangeRateResponse
+		req := woc.httpClient.R().
+			SetHeader("Accept", "application/json")
+		if woc.apiKey != "" {
+			req.SetHeader("Authorization", woc.apiKey)
+		}
+
+		res, err := req.
+			SetResult(&exchangeRateResponse).
+			Get(fmt.Sprintf("%s/exchangerate", woc.url))
```
chris-4chain: Just FYI:
Resty supports for automatic retries

pkg/services/services.go
```diff
@@ -0,0 +1,148 @@
...
+
+// Height returns the height of the active chain
+func (s *WalletServices) Height() int {
+	panic("Not implemented yet")
+}
+
+// BsvExchangeRate returns approximate exchange rate US Dollar / BSV, USD / BSV
+// This is the US Dollar price of one BSV
+func (s *WalletServices) BsvExchangeRate() (float64, error) {
+	bsvExchangeRate, err := s.whatsonchain.UpdateBsvExchangeRate(
```
chris-4chain, line 63: Don't we want to support a mechanism from TS that for certain methods there are primary and optional providers for doing something?

pkg/services/internal/whatsonchain/service.go
```diff
@@ -0,0 +1,79 @@
...
+		if err != nil {
+			return BSVExchangeRate{}, fmt.Errorf("failed to fetch exchange rate: %w", err)
+		}
+
+		if res.Status() == "Too Many Requests" && retry < 2 {
+			time.Sleep(2000 * time.Millisecond)
+			continue
+		}
```
dorzepowski: Resty can do the whole retry for you,
you could configure it already in the constructor by setting it like that:
```go
httpClient = 	woc.httpClient.Clone().
		SetRetryCount(3).
		SetRetryWaitTime(1 * time.Second).
		SetRetryMaxWaitTime(3 * time.Second).
		AddRetryCondition(func(res *resty.Response, err error) bool {
			return res.Status() == "Too Many Requests"
		})
```

or eventually skip the part with retry condition, and put it later on request like that (if that is more readable for you):
```go
req := woc.httpClient.R().
                 AddRetryCondition(func(res *resty.Response, err error) bool {
			return res.Status() == "Too Many Requests"
		}).
                SetHeader("Accept", "application/json")
```

pkg/services/services.go
```diff
@@ -0,0 +1,148 @@
...
+		&s.options.BsvExchangeRate,
+		&s.options.BsvUpdateMsecs,
+	)
+	if err != nil {
+		return 0, fmt.Errorf("error during bsvExchangeRate: %w", err)
+	}
+
+	s.options.BsvExchangeRate = bsvExchangeRate
+	return bsvExchangeRate.Rate, nil
```
dorzepowski: I don't like this approach,
I would rather prepare some even simple cache like component, which would called here and it would call the whatsonchain for BsvExchangeRates if the cached value is outdated. 

But we can leave that for later.

pkg/services/internal/whatsonchain/service.go
dzolt-4chain: Yeah I was trying to resemble "optional" parameter i will change this

pkg/services/internal/whatsonchain/service.go
dzolt-4chain: Based in their code I assumed, the status code could be 200 but the message that too many requests. If you think we should change it to status code I can change it
image

pkg/services/internal/whatsonchain/service.go
dzolt-4chain: Yes They're doing it the same way. The only time they allow a retry is when WoC responds with `Too Many Requests` So I'd leave it like this
image

dzolt-4chain: Just perhaps add these two checks aswell to check if there is a body and if the status text is not "OK"

dzolt-4chain: Actually I will leave it as it is these other checks are redundant in my opinion but make it so that resty will retry the call

pkg/services/internal/whatsonchain/service.go
dzolt-4chain: Yes, but here let's leave it as it is for now due to specific constraints as to when it should retry

pkg/services/services.go
dzolt-4chain: Yes, but it's still work in progress, I will be introducing these changes when other services will be created, for now I want to achieve "base" functionality with these functions and afterwards I will try to implement the optional or chained providers

pkg/wdk/primitives.go
dzolt-4chain: Yes, it appears that this type is being used in storage methods as well. So I would leave it for now in this package

pkg/services/services_options.go
dzolt-4chain: You're right. I'm not sure if these were created with intention or not. I will change it a little bit

pkg/services/services_options.go
dzolt-4chain: I changed it to const

pkg/services/services_options.go
dzolt-4chain: changed

pkg/services/services_options.go
dzolt-4chain: added a task for it https://github.com/4chain-ag/go-wallet-toolbox/issues/86

pkg/services/internal/whatsonchain/service.go
dzolt-4chain: But then it's not a duration, we can have `15 * time.Minute` but then it's not a miliseconds, but I'm not sure if it even needs to be ms. Maybe it doesn't

pkg/services/services_options.go
dzolt-4chain: I'm thinking of extracting these to their corresponding services, but I don't want to do that in this pr. I'd rather do it in later stage as this is still WIP and might change, after we introduce 2nd service

pkg/services/services_test.go
dzolt-4chain: done https://github.com/4chain-ag/go-wallet-toolbox/issues/89

pkg/services/services.go
dzolt-4chain: Lets leave it for another time

pkg/services/internal/whatsonchain/service.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	var exchangeRateResponse defs.BSVExchangeRateResponse
+	req := woc.httpClient.Clone().
+		SetRetryCount(defs.Retries).
+		SetRetryWaitTime(defs.RetriesWaitTime).
+		SetRetryMaxWaitTime(defs.Retries * defs.RetriesWaitTime).
+		AddRetryCondition(func(res *resty.Response, err error) bool {
+			return res.Status() == "Too Many Requests"
```
chris-4chain: We should rather check for StatusCode.

```suggestion
			return res.StatusCode() == 429
```

FYI: Status() is not what WoC returns in body. It is rather string representation of status code. 
For example, normally, we check for status 200 instead of string "OK".

pkg/services/internal/whatsonchain/service.go
```diff
@@ -0,0 +1,87 @@
...
+		thresholdTime := time.Now().Add(-updateInterval)
+
+		// Check if the rate timestamp is newer than the threshold time
+		if exchangeRate.Timestamp.After(thresholdTime) {
+			return *exchangeRate, nil
+		}
+	}
+
+	var exchangeRateResponse defs.BSVExchangeRateResponse
+	req := woc.httpClient.Clone().
```
chris-4chain: `Clone` cannot be call here - check documentation of this method:
> // Clone returns a clone of the original client.
> //
> // NOTE: Use with care:
> //   - Interface values are not deeply cloned. Thus, both the original and the
> //     clone will use the same value.
> //   - This function is not safe for concurrent use. You should only use this method
> //     when you are sure that any other goroutine is not using the client.

Better place to do se is the `New()` method.

pkg/services/internal/whatsonchain/service.go
```diff
@@ -0,0 +1,87 @@
...
+	req.SetHeader("Accept", "application/json")
+
+	if woc.apiKey != "" {
+		req.SetHeader("Authorization", woc.apiKey)
+	}
+
+	return req
+}
+
+func (woc *WhatsOnChain) UpdateBsvExchangeRate(exchangeRate *defs.BSVExchangeRate, bsvUpdateDuration *time.Duration) (defs.BSVExchangeRate, error) {
```
chris-4chain: 1. Why the `bsvUpdateDuration` is a pointer since for the only one usage of this method you dereference "normal" value.

```go
bsvExchangeRate, err := s.whatsonchain.UpdateBsvExchangeRate(
		s.config.BsvExchangeRate,
		&s.config.BsvUpdateInterval, // <------- here
	)
```

pkg/defs/constants.go
```diff
@@ -0,0 +1,14 @@
+package defs
+
+import "time"
+
+const (
+	// FifteenMinutes is a time.Duration of 15 minutes in milliseconds
+	FifteenMinutes = 15 * time.Minute
```
chris-4chain: 1. For me this value should be named after its purpose. For example:
`DefaultBSVExchangeUpdateDuration`
2. After renaming to more specific thing - I would move this to actual package (whatsonchain).... and make it private.

pkg/defs/constants.go
```diff
@@ -0,0 +1,14 @@
...
+	// TwentyFourHours is a time.Duration of 24 hours in milliseconds
+	TwentyFourHours = 24 * time.Hour
```
chris-4chain: Same comment here. In golang the duration definition is readable enough (`24 * time.Hour`) that you can define it in-place - without additional constant.
- If the constant doesn't add any other information

pkg/defs/constants.go
```diff
@@ -0,0 +1,14 @@
...
+	// Retries is the number of retries the client should make when trying to query for the chain's resource
+	Retries = 2
```
chris-4chain: `defs` package is not a good place to "place" here service specific constants. 
- I would move this to `whatsonchain` package.

pkg/defs/constants.go
```diff
@@ -0,0 +1,14 @@
...
+	// RetriesWaitTime is the duration to wait to retry a failed call to outside service
+	RetriesWaitTime = 2 * time.Second
```
chris-4chain: Same comment for this.

pkg/defs/services.go
```diff
@@ -0,0 +1,28 @@
...
+	Rate      float64             `mapstructure:"rates"`
+	Base      primitives.Currency `mapstructure:"base"`
+}
+
+// BSVExchangeRateResponse is the response from WhatsOnChain for bsv exchange range
+type BSVExchangeRateResponse struct {
+	Time     int     `json:"time" mapstructure:"time"`
+	Rate     float64 `json:"rate" mapstructure:"rate"`
+	Currency string  `json:"currency" mapstructure:"currency"`
+}
```
chris-4chain: This shouldn't be here. It is rather `whatsonchain` specific thing - and should be a private struct there.

pkg/services/services.go
```diff
@@ -0,0 +1,143 @@
...
+func (s *WalletServices) BsvExchangeRate() (float64, error) {
+	bsvExchangeRate, err := s.whatsonchain.UpdateBsvExchangeRate(
+		s.config.BsvExchangeRate,
+		&s.config.BsvUpdateInterval,
+	)
+	if err != nil {
+		return 0, fmt.Errorf("error during bsvExchangeRate: %w", err)
+	}
+
+	s.config.BsvExchangeRate = &bsvExchangeRate
```
chris-4chain, line 71: Why we use `config` as a place to hold current value. 
Some `WalletServices`-level field (variable) should be used instead.

For me - config should not be altered.

pkg/wdk/primitives/request.go
```diff
@@ -0,0 +1,8 @@
+package primitives
+
+// ReqHistoryNote is the history representation of the request
+type ReqHistoryNote struct {
+	When        *string
+	What        string
+	ExtraFields map[string]interface{}
+}
```
chris-4chain: This is not primitive - move it to `wdk`.

chris-4chain (changes requested):

pkg/wdk/primitives/currency.go
```diff
@@ -0,0 +1,11 @@
...
+
+// Currency represents supported currency types
+type Currency string
+
+// Supported currency types
+const (
+	USD Currency = "USD"
+	EUR Currency = "EUR"
+	GBP Currency = "GBP"
+)
```
chris-4chain: Please move it to `wdk`

pkg/defs/constants.go
dzolt-4chain: I've renamed but I cant change them to private as they are being used in testabilities to create a default configuration

pkg/services/services.go
dzolt-4chain: Yes I agree, in TS it's called options I named it config here, I would leave it for now, after I'm done with the whole WoC service I will create specific configurations and variables in the service when I have the whole picture with all the methods. Maybe it won't be even there if we will have some caching structure afterall

pkg/defs/services.go
```diff
@@ -0,0 +1,15 @@
...
+// 	Rates     map[wdk.Currency]float64 `mapstructure:"rates"`
+// 	Base      wdk.Currency             `mapstructure:"base"`
+// }
+//
+// // BSVExchangeRate is the rate struct for BSV exchange
+// type BSVExchangeRate struct {
+// 	Timestamp time.Time    `mapstructure:"timestamp"`
+// 	Rate      float64      `mapstructure:"rates"`
+// 	Base      wdk.Currency `mapstructure:"base"`
+// }
```
chris-4chain: Remove commented-out code

chris-4chain (approved):

PR #74, opened 2025-04-08
feat: random derivation

chris-4chain:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter

### DIFF
--- a/pkg/storage/internal/actions/create.go
+++ b/pkg/storage/internal/actions/create.go
@@ -93,7 +93,14 @@ func (c *create) Create(auth wdk.AuthID, args CreateActionParams) (*wdk.StorageC
 		return nil, fmt.Errorf("funding failed: %w", err)
 	}
 
-	return &wdk.StorageCreateActionResult{}, nil
+	derivationPrefix, err := txutils.RandomDerivation(16)
+	if err != nil {
+		return nil, fmt.Errorf("failed to generate random derivation prefix: %w", err)
+	}
+
+	return &wdk.StorageCreateActionResult{
+		DerivationPrefix: derivationPrefix,
+	}, nil
 }
 
 func (c *create) targetSat(args *CreateActionParams) (int64, error) {
--- a/pkg/storage/internal/actions/create.go
+++ b/pkg/storage/internal/actions/create.go
@@ -14,6 +14,10 @@ import (
 	"github.com/go-softwarelab/common/pkg/to"
 )
 
+const (
+	derivationPrefixLength = 16
+)
+
 type UTXO struct {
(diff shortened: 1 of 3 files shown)


pkg/storage/internal/actions/create.go
```diff
@@ -93,7 +93,14 @@ func (c *create) Create(auth wdk.AuthID, args CreateActionParams) (*wdk.StorageC
 		return nil, fmt.Errorf("funding failed: %w", err)
 	}
 
-	return &wdk.StorageCreateActionResult{}, nil
+	derivationPrefix, err := txutils.RandomDerivation(16)
```
dzolt-4chain: Maybe extract 16 as const or make a comment so it doesn't appear as a magic number 😄

pkg/storage/internal/txutils/random_derivation.go
```diff
@@ -0,0 +1,21 @@
+package txutils
+
+import (
+	"crypto/rand"
+	"encoding/base64"
+	"fmt"
+)
+
+func RandomDerivation(length uint64) (string, error) {
```
dorzepowski, line 9: ### minor
maybe the Randomizer (from change distirbution) could be an interface, 
and will handle random with max and random bytes 
🤷  just thinking out loud.

pkg/storage/internal/txutils/random_derivation_test.go
```diff
@@ -0,0 +1,16 @@
...
+)
+
+func TestRandomDerivation(t *testing.T) {
+	// when:
+	derivation, err := RandomDerivation(16)
+
+	// then:
+	require.NoError(t, err)
+	require.NotEmpty(t, derivation)
+}
```
dorzepowski, line 16: Maybe some more robust/detailed assertions/test cases?
like check it's different for each call, and if the length is really equal to expected. And that it's failing for 0 length

pkg/storage/internal/txutils/random_derivation.go
chris-4chain: I'm not sure if it's necessary for this case.

pkg/storage/internal/txutils/random_derivation_test.go
chris-4chain: ✅ Done

pkg/storage/internal/actions/create.go
chris-4chain: ✅ Done

dorzepowski (approved):

PR #75, opened 2025-04-08
feat: locking script key offset

chris-4chain:
### TS source code:
https://github.com/bitcoin-sv/wallet-toolbox/blob/master/src/storage/methods/createAction.ts#L927

## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter


dzolt-4chain (approved):

pkg/storage/internal/txutils/lock_script_key_offset_pub_key.go
```diff
@@ -0,0 +1,88 @@
...
+	pub, err := primitives.PublicKeyFromString(pubKey)
+	if err != nil {
+		return nil, "", fmt.Errorf("failed to parse public key: %w", err)
+	}
+
+	hashedSecret, keyOffset, err := l.keyOffsetToHashedSecret(pub)
+	if err != nil {
+		return nil, "", fmt.Errorf("failed to get hashed secret: %w", err)
+	}
+	newPointX, newPointY := primitives.S256().ScalarBaseMult(hashedSecret)
```
dorzepowski: add empty line before this line please.

pkg/storage/internal/txutils/lock_script_key_offset_pub_key.go
```diff
@@ -0,0 +1,88 @@
+package txutils
```
dorzepowski: I'm not sure if this is good place for this component,
It looks too generic, while the component is rather prepared to handle specific situation: deriving a key and address for fee for service provider, and util we will have real use case for making it more generic, I would rather make this a component that is generating address for service provider (so it can keep the service provider public key and can be instantiated once)

pkg/storage/internal/txutils/lock_script_key_offset_pub_key_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+	// and:
+	offsetPrivKey := "L1Yz9NvuvDru3Z7f8Kh8CK34U6UoKa6niynPEvqgHpm3AaNZig6z"
+	pubKey := "02f40c35f798e2ece03ae1ebf749545336db8402eb7e620bfe04d50da8ca8b06cc"
+
+	// and:
+	// mocking the offset private key generator
+	generator.offsetPrivGenerator = func() (*primitives.PrivateKey, error) {
+		return primitives.PrivateKeyFromWif(offsetPrivKey)
+	}
```
dorzepowski: I don't like this approach, I would prefer having the randomKeyGenerator as an argument in constructor instead, but I won't force that change.

pkg/storage/internal/txutils/lock_script_key_offset_pub_key.go
```diff
@@ -0,0 +1,88 @@
...
+	offset, err := l.offsetPrivGenerator()
+	if err != nil {
+		return nil, "", fmt.Errorf("failed to create new private key for keyOffset: %w", err)
+	}
+
+	sharedSecret, err := offset.DeriveSharedSecret(pub)
+	if err != nil {
+		return nil, "", fmt.Errorf("failed to derive shared secret: %w", err)
+	}
+	hashedSecret = crypto.Sha256(sharedSecret.ToDER())
```
dorzepowski: ### minor
add empty line before this line

pkg/storage/internal/txutils/lock_script_key_offset_pub_key.go
chris-4chain: ✅ Done - I've moved the code to `commission` package.
And prepared for usage for `createAction` - but I would like to implement actual usage in the next PR.

pkg/storage/internal/txutils/lock_script_key_offset_pub_key_test.go
chris-4chain: I've modified it and now there is a setter which can be optionally use.

```golang
type ScriptGenerator struct {
	offsetPrivGenerator func() (*primitives.PrivateKey, error)
	pubKey              string
}

// NewScriptGenerator creates a new instance of ScriptGenerator.
func NewScriptGenerator(pubKey string) *ScriptGenerator {
	return &ScriptGenerator{
		offsetPrivGenerator: randomPrivateKey,
		pubKey:              pubKey,
	}
}

// SetOffsetGenerator sets the generator function for the offset private key.
// Default is a random private key generator.
func (l *ScriptGenerator) SetOffsetGenerator(generator func() (*primitives.PrivateKey, error)) {
	l.offsetPrivGenerator = generator
}
```

pkg/storage/internal/txutils/lock_script_key_offset_pub_key.go
chris-4chain: ✅ Done

pkg/storage/internal/txutils/lock_script_key_offset_pub_key.go
chris-4chain: ✅ Done

pkg/infra/config.go
```diff
@@ -17,6 +17,7 @@ type Config struct {
 	DBConfig         defs.Database   `mapstructure:"db"`
 	HTTPConfig       HTTPConfig      `mapstructure:"http"`
 	Logging          LogConfig       `mapstructure:"logging"`
+	Commission       defs.Commission `mapstructure:"commission"`
```
dorzepowski, line 22: please apply this change on example yaml config. or wait for my changes with autogenerate for example config.

pkg/storage/internal/actions/create.go
```diff
@@ -59,18 +61,27 @@ type BasketRepo interface {
...
-		basketRepo: basketRepo,
+	c := &create{
+		logger:        logger,
+		funder:        funder,
+		basketRepo:    basketRepo,
+		commissionCfg: commissionCfg,
 	}
+
+	if commissionCfg.Enabled() {
+		c.commission = commission.NewScriptGenerator(commissionCfg.PubKeyHex)
```
dorzepowski: ### just leaving it there

maybe we could do it like that:
- if commission enabled, then use the ScriptGenerator
- if commission disabled, then apply NoCommisionScriptGenerator or something like that

And then when applying commision, first will really apply it, and the second won't do anything.
This could also enable future extensions, like making several commission outputs, or some weird scripts, etc. just by adding different commission service.

dorzepowski (approved):

pkg/storage/internal/actions/create.go
chris-4chain: In the next steps we'll discuss this 👍

dorzepowski (approved):

PR #76, opened 2025-04-08
feat: split tx funded by user into multiple change outputs.

dorzepowski:
Next part of funder implementation.

Feature: handle splitting user "top up" action into several change outputs.

### DIFF
--- a/pkg/storage/internal/actions/funder/sql.go
+++ b/pkg/storage/internal/actions/funder/sql.go
@@ -114,7 +114,8 @@ func newCollector(txSats int64, txSize uint64, numberOfDesiredUTXOs int, minimum
 	}
 
 	c.numberOfDesiredUTXOs = must.ConvertToUInt64(to.NoLessThan(numberOfDesiredUTXOs, 1))
-	c.minimumChange = minimumDesiredUTXOValue / 4
+
+	c.calculateMinimumChange()
 
 	err = c.calculateChangeOutputs()
 	if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/storage/internal/actions/funder/sql.go
```diff
@@ -84,31 +85,40 @@ func (f *SQL) loadUTXOs(ctx context.Context, userID int) iter.Seq2[*models.UserU
...
 		return nil, fmt.Errorf("failed to increase transaction size: %w", err)
 	}
 
-	if c.change() > 0 {
-		err = c.increaseSize(changeOutputSize)
-		if err != nil {
-			return nil, fmt.Errorf("failed to increase transaction size: %w", err)
-		}
+	c.numberOfDesiredUTXOs = must.ConvertToUInt64(to.NoLessThan(numberOfDesiredUTXOs, 1))
+	c.minimumChange = minimumDesiredUTXOValue / 4
```
dzolt-4chain: Why 4?

dorzepowski: I don't really know, 
but they do that in ts version, so I'm partially trying to restore this logic.

pkg/storage/internal/actions/funder/sql.go
```diff
@@ -195,10 +210,42 @@ func (c *utxoCollector) prepareResult() (*actions.FundingResult, error) {
 		return nil, fmt.Errorf("cannot convert change amount to uint64: %w", err)
 	}
 
+	// if adding change will increase the fee so there is no change anymore,
+	// we're dropping the changes and passing higher fee to miner.
```
chris-4chain:
```suggestion
	// If adding a change output would increase the fee to the point where no change remains,
	// we drop the change output and allow the higher fee to go to the miner.
```

chris-4chain (approved):

pkg/storage/internal/actions/funder/sql.go
chris-4chain: Maybe make a constant value for this 🤔 
That would make it slightly more descriptive.

dorzepowski: can you propose the name for this constant to make it more descriptive?
because it's already some kind of setting a const with name `minimumChange` by dividing desired utxo value by 4, 
so how can I name this const?
`minimumChangeDivisor` is it good name? or maybe you have better suggestion?

dzolt-4chain (approved):

PR #90, opened 2025-04-11
feat: separate whats on chain config

dzolt-4chain:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter


pkg/services/services_configuration.go
```diff
@@ -23,4 +21,6 @@ type WalletServicesConfiguration struct {
 	Chaintracks                     any // TODO: create *ChaintracksServiceClient
 	ArcUrl                          string
 	ArcConfig                       any // TODO: create *ArcConfig
+
+	WhatsOnChainConfiguration whatsonchain.WhatsOnChainConfiguration
```
chris-4chain: 1. Simple "WhatsOnChain" is enough - without "Configuration" suffix.
2. The struct you use is from `internal` package, which means - this field will not be able to be set by a "user" of this toolbox.

chris-4chain (changes requested):

pkg/services/services_configuration.go
dorzepowski: I thought about the same comments 😄 

🥇 for you

pkg/services/internal/whatsonchain/service.go
```diff
@@ -39,10 +42,12 @@ func New(httpClient *resty.Client, logger *slog.Logger, apiKey string, network d
...
-		httpClient: client,
-		apiKey:     apiKey,
-		url:        fmt.Sprintf("https://api.whatsonchain.com/v1/bsv/%s", network),
-		logger:     logger,
+		httpClient:        client,
+		apiKey:            config.ApiKey,
+		url:               fmt.Sprintf("https://api.whatsonchain.com/v1/bsv/%s", network),
+		logger:            logger,
+		bsvExchangeRate:   config.BsvExchangeRate,
+		bsvUpdateInterval: config.BsvUpdateInterval,
```
dorzepowski: set it with default so it doesn't need to be set every time we Update the exchange rate

pkg/services/internal/whatsonchain/service.go
```diff
@@ -25,9 +25,12 @@ type WhatsOnChain struct {
 	url        string
 	apiKey     string
 	logger     *slog.Logger
+
+	bsvExchangeRate   *wdk.BSVExchangeRate // TODO: possibly handle by some caching structure/redis
+	bsvUpdateInterval *time.Duration
```
dorzepowski: use value instead of pointer - because it's harder to maintain use it later as a pointer.

pkg/services/internal/whatsonchain/service.go
```diff
@@ -39,10 +42,12 @@ func New(httpClient *resty.Client, logger *slog.Logger, apiKey string, network d
...
 	return &WhatsOnChain{
-		httpClient: client,
-		apiKey:     apiKey,
-		url:        fmt.Sprintf("https://api.whatsonchain.com/v1/bsv/%s", network),
-		logger:     logger,
+		httpClient:        client,
+		apiKey:            config.ApiKey,
+		url:               fmt.Sprintf("https://api.whatsonchain.com/v1/bsv/%s", network),
+		logger:            logger,
+		bsvExchangeRate:   config.BsvExchangeRate,
```
dorzepowski: I would initialize this field with `config.BSVExchangeRate{}` this way, you won't need to check if it's nil and the timestamp will be year 1 so it's always outdated.

pkg/services/internal/whatsonchain/service.go
```diff
@@ -56,15 +62,15 @@ func (woc *WhatsOnChain) prepareRequest(req *resty.Request) *resty.Request {
 	return req
 }
```
dorzepowski, line 64: ohhh, because you're returning a value not pointer, there is no point to hold it as pointer, you can simply use value field, which will initialize to zero value by default (so the comment on line 49 will be redundant)

pkg/services/internal/whatsonchain/service.go
```diff
@@ -56,15 +61,15 @@ func (woc *WhatsOnChain) prepareRequest(req *resty.Request) *resty.Request {
...
+		updateInterval := to.IfThen(woc.bsvUpdateInterval != nil, *woc.bsvUpdateInterval).ElseThen(DefaultBSVExchangeUpdateInterval)
 		// Calculate the threshold time by subtracting updateMsecs from the current time
 		thresholdTime := time.Now().Add(-updateInterval)
 
 		// Check if the rate timestamp is newer than the threshold time
-		if exchangeRate.Timestamp.After(thresholdTime) {
-			return *exchangeRate, nil
+		if woc.bsvExchangeRate.Timestamp.After(thresholdTime) {
+			return *woc.bsvExchangeRate, nil
 		}
```
dorzepowski: ### suggestion

it's a little weird looking this `Add` with minus value, maybe something like this would be more readable?
```suggestion
		nextUpdate := woc.bsvExchangeRate.Timestamp.Add(updateInterval)
		if nextUpdate.After(time.Now()) {
			return *woc.bsvExchangeRate, nil
		}
```

dorzepowski (changes requested):

dorzepowski (approved):

pkg/services/configuration/whatsonchain.go
```diff
@@ -0,0 +1,14 @@
...
+
+import (
+	"time"
+
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/wdk"
+)
+
+// WhatsOnChain is a struct that configures WhatsOnChain service
+type WhatsOnChain struct {
+	ApiKey            string
```
chris-4chain:
```suggestion
	APIKey            string
```

pkg/services/configuration/whatsonchain.go
```diff
@@ -0,0 +1,14 @@
...
+	BsvExchangeRate   wdk.BSVExchangeRate
```
chris-4chain:
```suggestion
	BSVExchangeRate   wdk.BSVExchangeRate
```

pkg/services/configuration/whatsonchain.go
```diff
@@ -0,0 +1,14 @@
...
+	BsvUpdateInterval *time.Duration
```
chris-4chain:
```suggestion
	BSVUpdateInterval *time.Duration
```

pkg/services/configuration/services.go
```diff
@@ -7,14 +7,11 @@ import (
 	"github.com/4chain-ag/go-wallet-toolbox/pkg/wdk"
 )
 
-// WalletServicesConfiguration is a struct that has options for wallet services
-type WalletServicesConfiguration struct {
+// WalletServices is a struct that has options for wallet services
+type WalletServices struct {
```
chris-4chain, line 11: Is it meant to be used for configuration (viper / yaml)? If so, please add `mapstructure` tags.

chris-4chain (approved):

PR #95, opened 2025-04-15
feat: create-action store new outputs

chris-4chain:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter


pkg/storage/internal/actions/create.go
```diff
@@ -60,8 +61,8 @@ func FromValidCreateActionArgs(args *wdk.ValidCreateActionArgs) CreateActionPara
 		LockTime:    args.LockTime,
 		Description: string(args.Description),
 		Labels:      args.Labels,
-		Outputs:     seq.PointersFromSlice(args.Outputs),
-		Inputs:      seq.PointersFromSlice(args.Inputs),
+		Outputs:     args.Outputs,
```
dorzepowski: wouldn't it be better to make it a pointer at this point?
```suggestion
		Outputs:     slices.Map(args.Outputs, to.Ptr),
```

pkg/storage/internal/actions/create.go
```diff
@@ -145,17 +146,25 @@ func (c *create) Create(ctx context.Context, userID int, params CreateActionPara
...
+		funding.ChangeCount,
+		derivationPrefix,
+		params.Outputs,
+		commOut,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create new outputs: %w", err)
+	}
+
 	err = c.txRepo.CreateTransaction(ctx, &wdk.NewTx{
```
dorzepowski: ### just saying

I'm not sure if NewTx struct should be in package wdk,
maybe it could be some additional internal package 🤷‍♂️

pkg/storage/internal/actions/create.go
```diff
@@ -250,3 +260,112 @@ func (c *create) randomValues() (derivationPrefix string, reference string, err
...
+		if output.LockingScript != nil {
+			resultOutputs[i].LockingScript = *output.LockingScript
+		}
+
+		if output.Basket != nil {
+			resultOutputs[i].Basket = to.Ptr(primitives.StringUnder300(*output.Basket))
+		}
+	}
+
+	return resultOutputs
```
dorzepowski: ### just saying
You can use common `v0.22.0` and then:
```suggestion
	return slices.Map(newOutputs, func(output *wdk.NewOutput) wdk.StorageCreateTransactionSdkOutput {
		basket := optional.Map(optional.OfPtr(output.Basket),
			func(it string) *primitives.StringUnder300 { return to.Ptr(primitives.StringUnder300(it)) }).
			OrZeroValue()

		return wdk.StorageCreateTransactionSdkOutput{
			Vout:             output.Vout,
			ProvidedBy:       output.ProvidedBy,
			Purpose:          output.Purpose,
			DerivationSuffix: output.DerivationSuffix,
			ValidCreateActionOutput: wdk.ValidCreateActionOutput{
				Satoshis:           primitives.SatoshiValue(must.ConvertToUInt64(output.Satoshis)),
				OutputDescription:  primitives.String5to2000Bytes(output.Description),
				CustomInstructions: output.CustomInstructions,
				LockingScript:      optional.OfPtr(output.LockingScript).OrZeroValue(),
				Basket:             basket,
				// TODO: Tags
			},
		}
	})
```


the basket part could be done with simple if also if that's clearer.

pkg/storage/internal/database/models/output.go
```diff
@@ -0,0 +1,31 @@
...
+type Output struct {
+	gorm.Model
+
+	UserID        int    `gorm:"index"`
+	TransactionID uint   `gorm:"index"`
+	Vout          uint32 `gorm:"index"`
+	Satoshis      int64
+
+	LockingScript      *string
+	CustomInstructions *string
```
wregulski: Just a thought:
for `CustomInstructions` or `Description` to maybe put a `gorm:"type:text"` if we expect those fields to contain big strings.
But I am not sure if we want to support something more than sqlite for now.

pkg/storage/internal/actions/create.go
```diff
@@ -250,3 +260,112 @@ func (c *create) randomValues() (derivationPrefix string, reference string, err
 
 	return
 }
+
+func (c *create) newOutputs(
+	changDistribution iter.Seq[uint64],
```
wregulski: typo :D 
`changeDistribution`

pkg/storage/internal/database/models/labels.go
```diff
@@ -1,10 +1,16 @@
...
 
 type Label struct {
-	gorm.Model
+	CreatedAt time.Time
+	UpdatedAt time.Time
+	DeletedAt gorm.DeletedAt `gorm:"index"`
 
-	Name   string `gorm:"uniqueIndex:idx_name_userid"`
-	UserID int    `gorm:"uniqueIndex:idx_name_userid"`
+	Name   string `gorm:"primarykey"`
```
wregulski, line 14: Thought 🤔 
I don't know if this is a requirement for Name and UserID to be composed primary key.
If not, maybe we can consider adding:
```go
Name   string `gorm:"index:idx_user_label,unique"`
UserID int    `gorm:"index:idx_user_label,unique"`
```
for composite unique index.
Then it will be easier to make joins if we only have an ID as foreign key for other table.

wregulski: Oh, sorry I haven't seen that it was designed like this before	
```go
 - Name   string `gorm:"uniqueIndex:idx_name_userid"`
 - UserID int    `gorm:"uniqueIndex:idx_name_userid"`
```

so probably there is a reason for that change 😅🙈

pkg/storage/internal/actions/create.go
```diff
@@ -250,3 +260,112 @@ func (c *create) randomValues() (derivationPrefix string, reference string, err
...
+			return nil, fmt.Errorf("failed to generate random derivation suffix: %w", err)
+		}
+
+		all = append(all, &wdk.NewOutput{
+			Satoshis:         must.ConvertToInt64FromUnsigned(satoshis),
+			Basket:           to.Ptr(wdk.BasketNameForChange),
+			Spendable:        true,
+			Change:           true,
+			ProvidedBy:       wdk.ProvidedByStorage,
+			Type:             "P2PKH",
```
dorzepowski: maybe we could have an enum for that?

pkg/storage/internal/actions/create.go
```diff
@@ -250,3 +260,112 @@ func (c *create) randomValues() (derivationPrefix string, reference string, err
...
+	if commissionOutput != nil {
+		all = append(all, &wdk.NewOutput{
+			LockingScript: to.Ptr(commissionOutput.LockingScript),
+			Satoshis:      must.ConvertToInt64FromUnsigned(commissionOutput.Satoshis),
+			Basket:        nil,
+			Spendable:     false,
+			Change:        false,
+			ProvidedBy:    wdk.ProvidedByStorage,
+			Type:          "custom",
+			Purpose:       "storage-commission",
```
dorzepowski: maybe constant?

pkg/storage/internal/actions/create.go
```diff
@@ -250,3 +260,112 @@ func (c *create) randomValues() (derivationPrefix string, reference string, err
...
+		})
+	}
+
+	rand.Shuffle(len(all), func(i, j int) {
```
dorzepowski: I'm not sure, but I think that I saw that, the shuffling was done when some config was used.

pkg/storage/internal/database/models/labels.go
dorzepowski: There is a problem in gorm to make many2many relationship and have unique index.
We tried to do that yesterday, and the only good enough solution was this one (if we don't want to make struct for join table manually)

On the other hand the name and user id is really the id of the label, I'm not sure if having additional id would give any benefit.

pkg/storage/internal/database/models/output_baskets.go
```diff
@@ -14,8 +14,8 @@ type OutputBasket struct {
 
 	BasketID                int    `gorm:"primaryKey;not null"`
 	Name                    string `gorm:"type:varchar(300);not null;uniqueIndex:idx_name_user_id"`
-	NumberOfDesiredUTXOs    int64  `gorm:"not null"`
-	MinimumDesiredUTXOValue uint64 `gorm:"not null;default:10000"`
+	NumberOfDesiredUTXOs    int64  `gorm:"not null;column:number_of_desired_utxos;default:32"`
```
dorzepowski, line 17: ### just saying

In gorm, if you have field of value type (non pointer) and the default, will use the default for the value of `0`

pkg/storage/methodtests/create_action_test.go
```diff
@@ -71,10 +85,72 @@ func TestCreateActionWithCommission(t *testing.T) {
...
+	})
+	assert.Equal(t, primitives.SatoshiValue(10), commissionOutput.Satoshis)
+	assert.Nil(t, commissionOutput.Basket)
+	assert.Equal(t, wdk.ProvidedByStorage, commissionOutput.ProvidedBy)
+	assert.Nil(t, commissionOutput.DerivationSuffix)
+	assert.NotEmpty(t, commissionOutput.LockingScript)
+	assert.NoError(t, commissionOutput.LockingScript.Validate())
+	assert.Empty(t, commissionOutput.OutputDescription)
+	assert.Nil(t, commissionOutput.CustomInstructions)
+	assert.Empty(t, commissionOutput.Tags)
```
dorzepowski, line 105: Maybe having custom assertions (`then`) wouldn't be that bad 🙈

pkg/storage/internal/database/models/labels.go
chris-4chain: Yes 🙈

pkg/storage/internal/actions/create.go
chris-4chain: I'm moving it to `storage/internal/entity` package.

pkg/storage/internal/actions/create.go
chris-4chain: I'm using `optional.OfPtr` for locking script. 
But I would prefer to leave basket conversion and the mapping as is.

pkg/storage/internal/actions/create.go
chris-4chain: Nice catch 👍

pkg/storage/internal/database/models/output_baskets.go
chris-4chain: For now it works - I would leave it as is.

pkg/storage/methodtests/create_action_test.go
chris-4chain: Yes - but maybe we could think more about improving testability, and then do it in some of the next PRs

pkg/storage/internal/database/models/output.go
chris-4chain: ✅ Done

pkg/storage/internal/actions/create.go
chris-4chain: ✅ Done

pkg/storage/internal/actions/create.go
chris-4chain: No, it's not needed.

pkg/storage/internal/actions/create.go
chris-4chain: ✅ Done

pkg/storage/internal/actions/create.go
chris-4chain: ✅ Done

pkg/storage/internal/actions/create.go
chris-4chain: ✅ Done

wregulski (approved):

dorzepowski (approved):

PR #146, opened 2025-04-22
refactor: prepare abstract structure for services collection

### LINKED ISSUE #106: [FEATURE] Prepare ServiceCollection to handle fallback to other services

dorzepowski:
## Description of Changes

Prepare an abstraction for collection of services called for single method.

Current implementation allows to collect several "methods" in single queue, and later call them one by one until getting the successful response. In other words, this approach allow for calling some external service and having multiple fallback external services in case of failure.

Production example can be found in `RawTx` method in pkg/services/services.go

## Linked Issues / Tickets

Closes #106


pkg/services/internal/servicequeue/queues.go
```diff
@@ -0,0 +1,193 @@
...
+	})
+}
+
+type serv interface {
+	Name() string
+}
+
+func processOneByOne[S serv, R any](logger *slog.Logger, services []S, callService func(S) (R, error)) (R, error) {
+	if len(services) == 0 {
+		return to.ZeroValue[R](), errors.New("no services registered")
```
chris-4chain: Maybe stick to `fmt.Errorf`?

pkg/services/internal/servicequeue/queues.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+
+	results := seq.MapOrErr(seq.FromSlice(services), func(s S) (collection.Tuple2[string, R], error) {
```
chris-4chain: The name `results` suggests that it holds *all* the results. It could be misleading. Consider some other name, which tells that it is a lazy iterator.

I understand it, but - for example - AI (claude 3.7 sonnet thinking) not xD

pkg/services/internal/servicequeue/queues.go
```diff
@@ -0,0 +1,193 @@
...
+		res, err := callService(s)
+		if err != nil {
+			err = fmt.Errorf("service %s responds with error: %w", s.Name(), err)
+		}
+		return collection.NewTuple2(s.Name(), res), err
+	})
+
+	results = takeUntilHaveResult(results)
+
+	results = seq2.Each(results, func(serviceResult collection.Tuple2[string, R], err error) {
```
chris-4chain: Tuples are good for prototyping, but I would prefer to use some actual structs (generic in this use-case) that describes what it holds.

pkg/services/internal/servicequeue/queues.go
```diff
@@ -0,0 +1,193 @@
...
+		}
+	})
+
+	result := seq2.Reduce(results,
+		func(res collection.Tuple2[R, error], serviceResult collection.Tuple2[string, R], err error) collection.Tuple2[R, error] {
+			ret := collection.NewTuple2[R, error](serviceResult.B, errors.Join(res.B, err))
+			return ret
+		},
+		collection.Tuple2[R, error]{},
+	)
```
chris-4chain: It looks quite complex - and uses external methods that a reader needs to know what they do.

Consider more straight-forward way.
```suggestion
	result := collection.Tuple2[R, error]{} // tuple could be also changed to actual struct
	for serviceResult, err := range takeUntilHaveResult(results) {
		if err != nil {
			logger.Warn("error when calling service",
				slog.String("service.name", serviceResult.A),
				logging.Error(err),
			)
		}
		result = collection.NewTuple2[R, error](serviceResult.B, errors.Join(result.B, err))
	}
```

pkg/services/internal/whatsonchain/service.go
```diff
@@ -83,8 +81,16 @@ func (woc *WhatsOnChain) RawTx(txID string, chain defs.BSVNetwork) (*internal.Ra
...
 
-	result.RawTx = txHexDecoded
-	return result, nil
+	txIDFromRawTx := txutils.TransactionIDFromRawTx(txHexDecoded)
+	if txID != txIDFromRawTx {
+		return nil, fmt.Errorf("computed txid %s doesn't match requested value %s", txIDFromRawTx, txID)
+	}
+
+	return &wdk.RawTxResult{
+		Name:  "WoC",
```
chris-4chain: Here the name is `WoC` but for service name is `WhatsOnChain` - consider a single name for it. Maybe some const ? :)

pkg/services/services.go
```diff
@@ -45,75 +40,31 @@ func New(httpClient *resty.Client, logger *slog.Logger, config configuration.Wal
...
-			continue
-		}
-
-		txIDFromRawTx := txutils.TransactionIDFromRawTx(resp.RawTx)
-		if txID != txIDFromRawTx {
-			lastErr = fmt.Errorf("computed txid %s doesn't match requested value %s", txIDFromRawTx, txID)
-			s.rawTxServices.Next()
-			continue
+func (s *WalletServices) RawTx(txID string) (wdk.RawTxResult, error) {
+	result, err := s.rawTxServices.OneByOne(context.TODO(), txID)
```
chris-4chain, line 60: About the context - maybe the `servicequeue` could handle timeouts out of the box?

pkg/services/internal/servicequeue/queues.go
dorzepowski: It's a matter of taste, and in my opinion your suggestion aren't that much clearer or something.

> and uses external methods that a reader needs to know

The reduce method is popular in many languages as a part of filter/map/reduce functional operations, 
and for me, introducing iter.Seq in go, opens the doors to use this pattern, therefore I don't see any reason to not use this pattern just because a reader need to know the method Reduce, or something.

pkg/services/services.go
dorzepowski: can you elaborate a little bit more, what do you mean? Resty is handling timeout,how servicequeue should also handle this timeout?
you mean, that it should check the context.Done before next calls or something?

pkg/services/internal/servicequeue/queues.go
dorzepowski: But I play with the last part and error handling and can have something simpler.

chris-4chain (approved):

cain4chain (approved):
Overall looks good to me. 
Here are some thoughts during review.

- I tried to find a way to make the process more dynamic handling the params count in the services with the queue methods without needing to include the type declarations for the services. Didn't really come up with a compelling alternative though.

- Wondering about service prioritization as an optimization,  although I am not sure about the underlying process for setting service queues, and if we are adding parallelization later then would assume it would cover that aspect.

- I make also the assumption that all services successfully return a non nil response in the takeUntilHaveResult  method <`if is.Nil(result.Result)`> in which case this would still be fine.

Thanks!

PR #147, opened 2025-04-23
feat: create action returns inputs

chris-4chain:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter


pkg/internal/satoshi/satoshi.go
```diff
@@ -15,6 +15,21 @@ func (v Value) Int64() int64 {
 	return int64(v)
 }
 
+func (v Value) Int() (int, error) {
+	if v < 0 {
+		return 0, fmt.Errorf("cannot convert negative satoshi to int")
+	}
```
dorzepowski: hmmm... it's a little bit inconsistent with Int64 method
Maybe at least add documentation for that 🤷 
or it should be PostiveInt 🤷

pkg/storage/internal/repo/outputs.go
```diff
@@ -0,0 +1,69 @@
...
+
+func (o *Outputs) FindOutputs(ctx context.Context, outputIDs iter.Seq[uint]) ([]*wdk.TableOutput, error) {
+	count := seq.Count(outputIDs)
+	if count == 0 {
+		return nil, nil
+	}
+	idsClause := make([]any, 0, count)
+	for outputID := range outputIDs {
+		idsClause = append(idsClause, outputID)
+	}
```
dorzepowski: this looks simpler and would have only one iteration

```suggestion
	if seq.IsEmpty(outputIDs) {
		return nil, nil
	}
	
	idsClause := seq.Collect(outputIDs)
```

pkg/storage/internal/testabilities/fixture_faucet.go
```diff
@@ -1,28 +1,81 @@
...
-	funderFixture testabilities.FunderFixture
-	basket        *wdk.TableOutputBasket
+	t        testing.TB
+	user     testusers.User
+	db       *database.Database
+	basketID int
+	index    int
 }
 
 func (f *faucetFixture) TopUp(satoshis satoshi.Value) {
```
dorzepowski, line 31: Nice ❤️

pkg/storage/internal/actions/create.go
```diff
@@ -204,12 +213,18 @@ func (c *create) Create(ctx context.Context, userID int, params CreateActionPara
...
+		return nil, err
+	}
+
 	return &wdk.StorageCreateActionResult{
 		Reference:        reference,
 		Version:          params.Version,
 		LockTime:         params.LockTime,
 		DerivationPrefix: derivationPrefix,
 		Outputs:          c.resultOutputs(newOutputs),
+		Inputs:           resultInputs,
```
dorzepowski, line 227: maybe in a test we could check if the same utxo aren't selected twice as their should be reserved. Or maybe in `TestReservedUTXO` we should check for concrete error that the user has not enough funds?

pkg/internal/satoshi/satoshi.go
chris-4chain: The function is not used anymore - so I'm removing it.

pkg/storage/internal/actions/create.go
chris-4chain: I'm adding following check for `TestReservedUTXO`
```golang
require.ErrorIs(t, err, errfunder.NotEnoughFunds)
```

pkg/storage/internal/repo/outputs.go
chris-4chain: ✅ Done

pkg/storage/internal/actions/create.go
chris-4chain: ✅ Done

dorzepowski (approved):

pkg/storage/internal/actions/create.go
```diff
@@ -93,23 +92,36 @@ type TxRepo interface {
...
+) *create {
 	logger = logging.Child(logger, "createAction")
 	c := &create{
 		logger:        logger,
 		funder:        funder,
 		basketRepo:    basketRepo,
 		txRepo:        txRepo,
 		commissionCfg: commissionCfg,
+		outputRepo:    outputRepo,
 	}
```
mgosek-4chain, line 125: Could you please explain in which layer the arguments of `newCreateAction` are validated?

pkg/storage/internal/actions/create.go
```diff
@@ -385,6 +400,43 @@ func (c *create) resultOutputs(newOutputs []*entity.NewOutput) []wdk.StorageCrea
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to find allocated outputs: %w", err)
+	}
+	if len(utxos) != len(allocatedUTXOs) {
+		return nil, fmt.Errorf("expected %d outputs, got %d", len(allocatedUTXOs), len(utxos))
+	}
+
+	resultInputs := make([]wdk.StorageCreateTransactionSdkInput, len(allocatedUTXOs))
+	for i, utxo := range utxos {
+		if utxo.Txid == nil {
```
mgosek-4chain, line 416: Is it possible to get a nil `utxo`?

pkg/storage/internal/actions/create.go
```diff
@@ -385,6 +400,43 @@ func (c *create) resultOutputs(newOutputs []*entity.NewOutput) []wdk.StorageCrea
...
+			SourceSatoshis:        utxo.Satoshis,
+			SourceLockingScript:   *utxo.LockingScript,
+			UnlockingScriptLength: txutils.P2PKHUnlockingScriptLength,
+			ProvidedBy:            wdk.ProvidedByStorage,
+			Type:                  utxo.Type,
+			DerivationPrefix:      utxo.DerivationPrefix,
+			DerivationSuffix:      utxo.DerivationSuffix,
+
+			// TODO raw source tx when handle isSignAction
+		}
```
mgosek-4chain, line 435: Based on the previous comments, it might be worth creating a 'constructor' function that accepts `utxo` and `i` as arguments during call. Inside the function, you can define all the validation rules and return an error if any of them fail. This would make the codebase more cohesive.

pkg/storage/internal/methodtests/create_action_test.go
```diff
@@ -73,6 +74,19 @@ func TestCreateActionHappyPath(t *testing.T) {
...
+	assert.Equal(t, 0, input.Vin)
+	assert.NotEmpty(t, input.SourceTxid)
+	assert.Equal(t, uint32(0), input.SourceVout)
+	assert.Equal(t, int64(100_000), input.SourceSatoshis)
+	assert.NotEmpty(t, input.SourceLockingScript)
+	assert.Nil(t, input.SourceTransaction)
+	assert.Equal(t, wdk.ProvidedByStorage, input.ProvidedBy)
+	assert.Equal(t, string(wdk.OutputTypeP2PKH), input.Type)
+	assert.NotEmpty(t, input.DerivationSuffix)
+	assert.NotEmpty(t, input.DerivationPrefix)
```
mgosek-4chain: This approach can lead to false positives. If you define an extra field and forget to add a corresponding assertion, the test will still pass. Instead, I'd suggest defining the expected state of the structure and using a simple `assert.Equal` to compare the actual and expected values.

wregulski (approved):

pkg/storage/internal/database/models/output.go
```diff
@@ -28,4 +28,6 @@ type Output struct {
 	ProvidedBy  string
 	Purpose     string
 	Type        string
+
+	Transaction *Transaction `gorm:"foreignKey:TransactionID;references:ID"`
```
mgosek-4chain, line 32: Could you explain please why the previous fields don't contain any `gorm` tags?

pkg/storage/internal/repo/outputs.go
```diff
@@ -0,0 +1,66 @@
...
+			ProvidedBy:         output.ProvidedBy,
+			Purpose:            output.Purpose,
+			Type:               output.Type,
+			Txid:               output.Transaction.TxID,
+			DerivationPrefix:   output.DerivationPrefix,
+			DerivationSuffix:   output.DerivationSuffix,
+			CustomInstructions: output.CustomInstructions,
+			LockingScript:      output.LockingScript,
+		}
+	}), nil
```
mgosek-4chain, line 65: Have you considered using `FindInBatches` and performing necessary mapping within the defined function for batch processing?

pkg/storage/internal/actions/create.go
chris-4chain: https://github.com/bsv-blockchain/go-wallet-toolbox/blob/193bbf2ece03e716435d94c7b1038940eb3ac4f3/pkg/storage/provider.go#L255-L259

pkg/storage/internal/actions/create.go
chris-4chain: No

pkg/storage/internal/actions/create.go
chris-4chain: No, this is only one place where it is done, it's internal and it's separated in a function (you can call it "constructor" if you will).
- So I would like to leave it as is.

pkg/storage/internal/methodtests/create_action_test.go
chris-4chain: This is the expected state definition - moving it to other struct adds only one extra layer and doesn't solve the problem.

pkg/storage/internal/database/models/output.go
chris-4chain: Because they don't need it - GORM default behavior is applied. 
This one needs.

pkg/storage/internal/repo/outputs.go
chris-4chain: Yes I considered, and decided not to use it. The expected size of a row and count of rows won't exceed a level when it would be beneficial. 

Besides - we will identify bottlenecks in the future when it starts to work. Then we'll adjust such functions if needed.

pkg/storage/internal/methodtests/create_action_test.go
chris-4chain: I've added additional checks for randomized DerivationPrefix and DerivationSuffix.

PR #156, opened 2025-04-28
feat: add function to run all services from queue in parallel.

dorzepowski:
The goal of this PR is to prepare a method on servicequeue that would allow to call all the services in parallel and receive their results. This will be used in Wallet Services to broadcast transaction to all registered services.

### DIFF
--- a/pkg/services/internal/parallel.go
+++ b/pkg/services/internal/parallel.go
@@ -20,10 +20,11 @@ func MapParallel[E any, R any](ctx context.Context, sequence iter.Seq[E], runner
 
 		ctx, cancel := context.WithCancel(ctx)
 
+	startGoRoutines:
 		for v := range sequence {
 			select {
 			case <-ctx.Done():
-				break
+				break startGoRoutines
 			default:
 				wg.Add(1)
 				go func(v E) {
--- a/pkg/services/internal/parallel.go
+++ b/pkg/services/internal/parallel.go
@@ -1,0 +1,69 @@
+package internal
+
+import (
+	"context"
+	"iter"
+	"sync"
+)
+
+func MapParallel[E any, R any](ctx context.Context, sequence iter.Seq[E], runner func(context.Context, E) R) iter.Seq[R] {
+	if sequence == nil {
+		return func(yield func(R) bool) {}
+	}
+
+	return func(yield func(R) bool) {
+		wg := &sync.WaitGroup{}
+
+		results := make(chan R, 100)
+
+		ctx, cancel := context.WithCancel(ctx)
+
+		for v := range sequence {
+			wg.Add(1)
+			go func(v E) {
+				defer wg.Done()
+
+				var result R
+				select {
+				case <-ctx.Done():
+					return
+				default:
+					result = runner(ctx, v)
+				}
+
+				select {
+				case <-ctx.Done():
+					return
+				default:
+					results <- result
+				}
+
+			}(v)
+		}
+
+		go func() {
+			wg.Wait()
+			close(results)
+		}()
+
+		for {
+			select {
+			case <-ctx.Done():
+				cancel()
+				return
+			case res, ok := <-results:
+				if !ok {
+					cancel()
+					return
+				}
+				if !yield(res) {
+					cancel()
+					for range results {
+						// drain the channel to avoid memory leaks
+					}
+					return
+				}
+			}
+		}
+	}
+}
--- a/pkg/services/internal/parallel.go
+++ b/pkg/services/internal/parallel.go
@@ -6,6 +6,8 @@ import (
 	"sync"
 )
 
+const defaultChannelBuffer = 100
+
 func MapParallel[E any, R any](ctx context.Context, sequence iter.Seq[E], runner func(context.Context, E) R) iter.Seq[R] {
 	if sequence == nil {
 		return func(yield func(R) bool) {}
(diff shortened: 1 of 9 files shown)


pkg/services/internal/parallel.go
```diff
@@ -0,0 +1,69 @@
...
+func MapParallel[E any, R any](ctx context.Context, sequence iter.Seq[E], runner func(context.Context, E) R) iter.Seq[R] {
+	if sequence == nil {
+		return func(yield func(R) bool) {}
+	}
+
+	return func(yield func(R) bool) {
+		wg := &sync.WaitGroup{}
+
+		results := make(chan R, 100)
+
```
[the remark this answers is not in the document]
dorzepowski: this is intentional approach because at this point we don't know how many services could be in the sequence, and if we really would receive the result from all of them.
IMHO, this parameter could be adjusted in the future if we would find any issue with it.

pkg/services/internal/servicequeue/queues.go
```diff
@@ -131,6 +161,41 @@ type serv interface {
...
 
+func processParallel[S serv, R any](ctx context.Context, logger *slog.Logger, services []S, callService func(context.Context, S) (R, error)) ([]*NamedResult[R], error) {
+	if len(services) == 0 {
+		return nil, ErrNoServicesRegistered
+	}
+
+	results := internal.MapParallel(ctx, seq.FromSlice(services), func(ctxParallel context.Context, s S) (result *NamedResult[R]) {
+		defer func() {
+			if r := recover(); r != nil {
+				err := fmt.Errorf("%v", r)
```
chris-4chain: You could add more context to the error message.

pkg/services/internal/parallel.go
```diff
@@ -0,0 +1,69 @@
...
+		ctx, cancel := context.WithCancel(ctx)
+
+		for v := range sequence {
+			wg.Add(1)
+			go func(v E) {
```
chris-4chain: Just a thought. Shouldn't you check for ctx.Done before spawning new goroutine?

pkg/services/internal/parallel.go
```diff
@@ -0,0 +1,69 @@
...
+
+func MapParallel[E any, R any](ctx context.Context, sequence iter.Seq[E], runner func(context.Context, E) R) iter.Seq[R] {
+	if sequence == nil {
+		return func(yield func(R) bool) {}
+	}
+
+	return func(yield func(R) bool) {
+		wg := &sync.WaitGroup{}
+
+		results := make(chan R, 100)
```
chris-4chain: Please make constant for it.

pkg/services/internal/parallel.go
```diff
@@ -0,0 +1,69 @@
...
+			wg.Add(1)
+			go func(v E) {
+				defer wg.Done()
+
+				var result R
+				select {
+				case <-ctx.Done():
+					return
+				default:
+					result = runner(ctx, v)
```
chris-4chain: How about the situation when some `runner` runs indefinitely?

dorzepowski: This is the responsibility of the "adopter"/client to pass the context with timeout.
Then when it's timed out the line 51 will handle it, and finish this method.

BUT I don't know about any method to kill the goroutine from outside, so if it's hanging, then it will hang for ever probably, but this is the adopter/client problem, that he is passing something that doesn't check the context and then work indefinitely and burn resources.

pkg/services/internal/parallel.go
dorzepowski: I thought about it already, but 
It already looks quite complex - and this additional check would make it even more complex, while I don't even know how to test it, to ensure that actually it's working as expected.

But maybe it's better to have this event if not confirmed if it's working then not having it at all 🤔 

So ok, I will change this.

pkg/services/internal/servicequeue/queues_test.go
```diff
@@ -225,47 +246,399 @@ func TestQueue(t *testing.T) {
...
+			// given:
+			services := slices.Map(test.services, func(s TestService) *servicequeue.Service2[string, int, *TestServiceResult] {
+				service := s.NewTest(t)
+				return servicequeue.NewService2(service.Name, service.Do2)
+			})
+
+			// and:
+			queue := servicequeue.NewQueue2(
+				logging.NewTestLogger(t),
+				"Do3",
```
chris-4chain: Most probably "Do2" :)

pkg/services/internal/servicequeue/queues_test.go
```diff
@@ -225,47 +246,399 @@ func TestQueue(t *testing.T) {
...
+				services...,
+			)
+
+			// when:
+			results, err := queue.All(context.Background())
+
+			// debug: show example of error from panicing service
+			slices.ForEach(results, func(result *servicequeue.NamedResult[*TestServiceResult]) {
+				if result.IsError() {
+					fmt.Println(result.GetError())
```
chris-4chain: 1. Should this block remain? 
2. Consider `t.Logf`

pkg/services/internal/servicequeue/queues_test.go
```diff
@@ -225,47 +246,399 @@ func TestQueue(t *testing.T) {
...
+			queue := servicequeue.NewQueue(
+				logging.NewTestLogger(t),
+				"Do",
+				services...,
+			)
+
+			// when:
+			results, err := queue.All(context.Background())
+
+			// debug: show example of error from panicing service
```
chris-4chain:
```suggestion
			// debug: show example of error from panicking service
```

Check other occurrences.

pkg/services/internal/servicequeue/queues_test.go
dorzepowski: 1. I would keep this, to show how the error can look like (and that it contains the stacktrace
2. Right: I will use `t.Log`

chris-4chain (approved):

wregulski (approved):

PR #178, opened 2025-04-30
feat: ProvenTxReq with history tracking & raw tx storage

chris-4chain:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter


pkg/storage/internal/database/models/proven_tx_req.go
```diff
@@ -0,0 +1,43 @@
...
+	}
+
+	note := HistoryNote{
+		When:  when,
+		What:  what,
+		Attrs: attrs,
+	}
+	history.Notes = append(history.Notes, note)
+
+	data, _ := json.Marshal(history)
```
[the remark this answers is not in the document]
chris-4chain: Don't know if this is necessary. The history notes are not crucial for the process. Besides, we would need to pass the logger instance up here.

pkg/storage/internal/testabilities/fixture_faucet.go
```diff
@@ -77,7 +87,9 @@ func (f *faucetFixture) TopUp(satoshis satoshi.Value) (txtestabilities.Transacti
 		Output: output,
 	}
 
-	f.db.DB.WithContext(f.t.Context()).Create(utxo)
+	tx := f.db.DB.WithContext(f.t.Context())
```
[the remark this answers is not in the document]
chris-4chain: Not necessary, since it's only test fixture.

pkg/storage/internal/repo/proven_tx_req.go
```diff
@@ -0,0 +1,56 @@
...
+
+func (p *ProvenTxReq) UpsertProvenTxReq(ctx context.Context, req *entity.UpsertProvenTxReq, historyNote string, historyAttrs map[string]any) error {
+	err := p.db.WithContext(ctx).Transaction(func(tx *gorm.DB) error {
+		var model models.ProvenTxReq
+		err := tx.First(&model, "tx_id = ? ", req.TxID).Error
+		if err != nil && !errors.Is(err, gorm.ErrRecordNotFound) {
+			return fmt.Errorf("cannot upsert proven tx req: %w", err)
+		}
+
+		model.Status = req.Status // TODO: Shouldn't we check the status first? Only if it's higher than the current one, we should update it.
```
wregulski, line 30: Maybe we can define statuses like a map with a proper order (if it's applicable):
```go
var statusRank = map[wdk.ProvenTxReqStatus]int{
	ProvenTxStatusSending: 0,
         ...
}

func rankOf(status wdk.ProvenTxReqStatus) int {
	if r, ok := statusRank[status]; ok {
		return r
	}

	return -1
}

if rankOf(req.Status) < rankOf(model.Status) {
	return "something..."
}

```

or with "iota"? 🤔

pkg/storage/internal/database/models/proven_tx_req.go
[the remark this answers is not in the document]
dorzepowski: I have an idea how to improve it, so you can leave it as is.

pkg/storage/internal/methodtests/create_action_test.go
```diff
@@ -94,6 +95,33 @@ func TestCreateActionHappyPath(t *testing.T) {
 	// TODO: Test DB state: but after we make actual getter methods, like ListActions
 }
 
+func TestCreateActionIsSignActionHappyPath(t *testing.T) {
```
dorzepowski: ### minor
```suggestion
func TestCreateActionWithSignActionHappyPath(t *testing.T) {
```

pkg/storage/internal/actions/internalize.go
```diff
@@ -62,6 +59,17 @@ func (in *internalize) Internalize(ctx context.Context, userID int, args *wdk.In
 		return nil, fmt.Errorf("failed to generate random reference: %w", err)
 	}
 
+	// TODO: Don't upsert ProvenTxReq if the transaction is already known in ProvenTx (not *Req)
+	err = in.provenTxRepo.UpsertProvenTxReq(ctx, &entity.UpsertProvenTxReq{
+		TxID:      txID,
+		RawTx:     tx.Bytes(),
+		InputBeef: args.Tx,
+		Status:    wdk.ProvenTxStatusUnmined,
+	}, entity.InternalizeActionHistoryNote, entity.UserIDHistoryAttr(userID))
```
dorzepowski: maybe package name `history` would be clearer

chris-4chain: I think so, but I would leave it as is because we will rework it anyway

dorzepowski: I know, but it's just about a package name, which change should be simple at this stage (and maybe we just move later this package and that's all).

pkg/storage/internal/repo/proven_tx_req.go
chris-4chain: Yes, we could implement it that way - but... in the TS version, they don't do such a thing at all - so, for now, I wouldn't like to do it either.

wregulski: Ok👍

wregulski (approved):

dorzepowski (approved):

PR #191, opened 2025-05-06
feat: processAction validate args and preprocessing (more tests will be added in upcoming PRs)

chris-4chain:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter


pkg/storage/internal/actions/process.go
```diff
@@ -0,0 +1,113 @@
...
+
+	return nil, nil
+}
+
+func (p *process) processNewTx(ctx context.Context, userID int, args *wdk.ProcessActionArgs) error {
+	tx, err := transaction.NewTransactionFromBytes(args.RawTx)
+	if err != nil {
+		return fmt.Errorf("failed to build transaction object from raw tx bytes: %w", err)
+	}
+	txID := tx.TxID().String()
```
dorzepowski: ### minor
empty line before this one please 🙈

pkg/storage/internal/actions/process.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+	if !tableTx.IsOutgoing {
+		return fmt.Errorf("transaction with reference (%s) is not outgoing", reference)
+	}
+	if len(tableTx.InputBEEF) == 0 {
+		return fmt.Errorf("transaction with reference (%s) has no inputBEEF", reference)
+	}
+	if tableTx.Status != wdk.TxStatusUnsigned && tableTx.Status != wdk.TxStatusUnprocessed {
+		return fmt.Errorf("transaction with reference (%s) is not in a valid status for processing", reference)
+	}
```
dorzepowski: ### minor
add empty line between `if` statements.

pkg/storage/internal/repo/transactions.go
```diff
@@ -177,20 +177,40 @@ func (txs *Transactions) FindTransactionByUserIDAndTxID(ctx context.Context, use
...
-		LockTime:      to.Ptr(transaction.LockTime),
-		TxID:          transaction.TxID,
-		InputBEEF:     transaction.InputBeef,
-	}, nil
+	return txs.mapModelToTableTransaction(&transaction), nil
+
+}
 
+func (txs *Transactions) FindTransactionByReference(ctx context.Context, userID int, reference string) (*wdk.TableTransaction, error) {
+	transaction := models.Transaction{}
```
dorzepowski, line 185: Just thinking out loud

maybe we could make a join with outputs and inputs here and load all in one go?

pkg/internal/validate/validate_process_action_args.go
```diff
@@ -0,0 +1,29 @@
+package validate
+
+import (
+	"fmt"
+
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/wdk"
+)
+
+func ProcessActionArgs(args *wdk.ProcessActionArgs) error {
+	if args.IsNewTx {
```
dorzepowski, line 10: 🤔 hmmm.. if it's not new, then all those can be nil? or it's just a partial implementation?

pkg/storage/internal/actions/process.go
```diff
@@ -0,0 +1,113 @@
...
+
+	// TODO: Services::nLockTimeIsFinal(tx)
+
+	tableTx, err := p.txRepo.FindTransactionByReference(ctx, userID, *args.Reference)
+	if err != nil {
+		return fmt.Errorf("failed to find transaction by reference: %w", err)
+	}
+
+	err = p.validateNewTxTable(*args.Reference, tableTx)
```
dorzepowski: ### minor
maybe `validateStateOfTableTx`

pkg/storage/internal/actions/process.go
```diff
@@ -0,0 +1,113 @@
...
+				return fmt.Errorf("output index %d is out of range of provided tx outputs count %d", voutInt, len(tx.Outputs))
+			}
+			fromDB := *output.LockingScript
+			providedInArgs := tx.Outputs[voutInt].LockingScript.String()
+			if providedInArgs != fromDB {
+				return fmt.Errorf("locking script mismatch as vout: %d, provided %s, calculated from raw tx: %s", voutInt, providedInArgs, fromDB)
+			}
+		}
+	}
+	return nil
```
dorzepowski:
```suggestion
	for _, output := range outputs {
		if output.LockingScript != nil {
			continue
		}

		voutInt := must.ConvertToIntFromUnsigned(output.Vout)
		if voutInt >= len(tx.Outputs) {
			return fmt.Errorf("output index %d is out of range of provided tx outputs count %d", voutInt, len(tx.Outputs))
		}

		fromDB := *output.LockingScript
		providedInArgs := tx.Outputs[voutInt].LockingScript.String()
		if providedInArgs != fromDB {
			return fmt.Errorf("locking script mismatch as vout: %d, provided %s, calculated from raw tx: %s", voutInt, providedInArgs, fromDB)
		}
	}
	return nil
```

pkg/storage/internal/actions/process.go
```diff
@@ -0,0 +1,113 @@
...
+	if tableTx.Status != wdk.TxStatusUnsigned && tableTx.Status != wdk.TxStatusUnprocessed {
+		return fmt.Errorf("transaction with reference (%s) is not in a valid status for processing", reference)
+	}
+	return nil
+}
+
+func (p *process) validateNewTxOutputs(tx *transaction.Transaction, outputs []*wdk.TableOutput) error {
+	for _, output := range outputs {
+		voutInt := must.ConvertToIntFromUnsigned(output.Vout)
+		if output.LockingScript != nil {
```
dorzepowski: I'm not sure why we have this if?
I mean, it seems like some workaround or hidden deduction.

If my assumption is valid, if it is nil, then it's custom output,
if so, then maybe we should simply check if output.Change is true?
And if not, then maybe having nil Locking script is also wrong 🤷

pkg/storage/internal/actions/process.go
```diff
@@ -0,0 +1,113 @@
...
+func (p *process) processNewTx(ctx context.Context, userID int, args *wdk.ProcessActionArgs) error {
+	tx, err := transaction.NewTransactionFromBytes(args.RawTx)
+	if err != nil {
+		return fmt.Errorf("failed to build transaction object from raw tx bytes: %w", err)
+	}
+	txID := tx.TxID().String()
+
+	if txID != string(*args.TxID) {
+		return fmt.Errorf("txID mismatch: provided %s, calculated from raw tx: %s", *args.TxID, txID)
+	}
```
dorzepowski: This part looks rather like a part of validating args

pkg/storage/internal/actions/process.go
chris-4chain: I don't think it's necessary here. But sure, I can do it 🤷

pkg/storage/internal/repo/transactions.go
chris-4chain: I thought about that, but 
- outputs are not needed for tx-validation process. So outputs are loaded only if validation passes.
- the code with josin is more complex 
- in `ts` they also load tx and then outputs separately

So, I would do such optimisations when we have actual feedback/metrics - not now.

pkg/internal/validate/validate_process_action_args.go
chris-4chain: > then all those can be nil?

Yes

pkg/storage/internal/actions/process.go
chris-4chain: Nice catch :) 
Of course I should have used "continue" 👍 
But, there is a mistake in your suggestion. It should be `output.LockingScript == nil` (not `!=` 😄)

pkg/storage/internal/actions/process.go
chris-4chain: 1. This is the `logic` from TS version
2. Your logic is `not` valid (I think) - if locking script is nil - it's a change output generated during `createAction`

I would leave it as is, at least for now, until I learned more how the process action works.

pkg/storage/internal/actions/process.go
chris-4chain: I didn't want to put it there, because in `validation` I make a trivial checking. 

1. Here - it requires parsing the raw tx - and it's rather additional validation for consistency. 
2. Simpli, I didn't want to parse the tx twice.

pkg/storage/internal/actions/process.go
dorzepowski: yes, sorry, I was doing it in gh comment text area 😄

pkg/storage/internal/actions/process.go
dorzepowski: 1. I understand that it is from ts, but I don't understand this logic, as I wrote it looks like a workaround.
2. I believe that this is a goal of the field Change to say, that this output is a change generated in createAction - or in other words, that this output is something that can be handled by the system. So if it is not a change then it should have LockingScript, and all the validation for locking script should be applied.

pkg/storage/internal/actions/process.go
dorzepowski: I don't understand the point 2, why would you parse tx twice? I don't yet see any reason to parse it, apart from validating the txID. But I don't know whole logic.

chris-4chain: Hmm, agree - for now the `tx` object is made only for validating. But I'm 88% sure 😄 it will be necessary in the future. 
Let's see... Can we agree that I'll move this "validation" logic to `validation` package when I'll be sure and know the further logic?

dorzepowski (approved):

PR #195, opened 2025-05-08
feat(#161): get status of broadcasted transactions; test: handling non success response on broadcasting

dorzepowski:


pkg/services/internal/arc/arc_broadcast.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+func (s *Service) broadcast(ctx context.Context, txHex string) (*TXInfo, error) {
+	result := &TXInfo{}
+	arcErr := &APIError{}
+
+	headers := httpx.NewHeaders().
+		Set("X-CallbackUrl").IfNotEmpty(s.config.CallbackURL).
+		Set("X-CallbackToken").IfNotEmpty(s.config.CallbackToken).
+		Set("X-WaitFor").IfNotEmpty(s.config.WaitFor)
```
chris-4chain: Could it be configured once at NewARCService level?

pkg/services/internal/arc/arc_service.go
```diff
@@ -56,45 +56,128 @@ func NewARCService(logger *slog.Logger, httpClient *resty.Client, config Config)
...
-	// for each txids
-	// 	if txid == broadcasted tx.ID
-	// 		get tx
+}
+
+func (s *Service) getTxIDResults(ctx context.Context, txInfo *TXInfo, txIDs []string) iter.Seq[results.PostTxID] {
+	subjectTxResult := internal.NewNamedResult(txInfo.TxID, types.SuccessResult(txInfo))
+
+	txIDsToGetStatus := seq.Filter(seq.FromSlice(txIDs), func(txID string) bool {
+		return txInfo.TxID != txID
```
chris-4chain, line 99: Hmm.... This and the `getTxIDResults` could be described by some comment, that explains why we need to request ARC for txIDs (except the txID)

pkg/services/internal/arc/arc_service.go
```diff
@@ -56,45 +56,128 @@ func NewARCService(logger *slog.Logger, httpClient *resty.Client, config Config)
...
+	})
+
+	txsData := internal.MapParallel(ctx, txIDsToGetStatus, s.getTransactionData)
 
+	txsData = seq.Prepend(txsData, subjectTxResult)
```
chris-4chain, line 105: The `subjectTxResult` could be defined just before this line.

pkg/services/internal/arc/arc_service.go
```diff
@@ -56,45 +56,128 @@ func NewARCService(logger *slog.Logger, httpClient *resty.Client, config Config)
...
+
+	resultsForTxID := seq.Map(txsData, toResultForPostTxID)
+	return resultsForTxID
```
chris-4chain:
```suggestion
	return seq.Map(txsData, toResultForPostTxID)
```

pkg/services/internal/arc/arc_service.go
```diff
@@ -56,45 +56,128 @@ func NewARCService(logger *slog.Logger, httpClient *resty.Client, config Config)
...
+		merklePath, err := transaction.NewMerklePathFromHex(info.MerklePath)
+		if err != nil {
+			result.Error = err
+			result.Result = results.ResultStatusError
+		} else {
+			result.MerklePath = merklePath
+		}
+	}
+
+	dataBytes, err := json.Marshal(info)
```
chris-4chain, line 140: Why result.Bytes is a JSON string? (Instead of just TXInfo object)?

pkg/services/internal/arc/arc_service_test.go
```diff
@@ -29,7 +30,43 @@ func TestPostBEEFWithARCService(t *testing.T) {
...
+		txID := tx.TxID().String()
+
+		// when:
+		res, err := service.PostBeef(context.Background(), beef, nil)
+
+		// then:
+		assert.NoError(t, err)
+		assert.NotNil(t, res)
+
+		exp := []results.PostTxID{
```
chris-4chain:
```suggestion
		expected := []results.PostTxID{
```

pkg/services/internal/testabilities/fixture_arc.go
```diff
@@ -88,23 +127,49 @@ func (f *arcFixture) IsUpAndRunning() {
...
+		merklePath := optional.OfPtr(tx.Transaction.MerklePath)
+
+		return &knownTransaction{
+			txid:        txID,
+			status:      to.IfThen(merklePath.IsEmpty(), "SEEN_ON_NETWORK").ElseThen("MINED"),
+			blockHeight: optional.Map(merklePath, func(it sdk.MerklePath) uint32 { return it.BlockHeight }).OrZeroValue(),
+			blockHash: optional.Map(merklePath, func(it sdk.MerklePath) string {
+				// it's not that important so, let's ignore the error here
+				root, _ := it.ComputeRootHex(&txID)
+				return root
```
chris-4chain:
```suggestion
				root, err := it.ComputeRootHex(&txID)
				assert.NoError(f, err, "Your message, describing why it's not important :D")
				return root
```

pkg/services/internal/arc/arc_broadcast.go
dorzepowski: 🤔 maybe

pkg/services/internal/arc/arc_service.go
dorzepowski: because TxInfo is internal, and data is something that probably someone could somehow look into for some reason, and if this would be an internal struct then he has very little possibilities to handle this.

pkg/services/internal/arc/arc_service.go
dorzepowski: Haha, I moved it earlier on another branch, and I was confused that you're proposing this change, as I already did that 😅

pkg/services/internal/arc/arc_service.go
dorzepowski: I changed the name of the variable, IMHO now it's more understandable without comment.

pkg/services/internal/arc/arc_service_test.go
dorzepowski: I inlined this, hope this is ok.

pkg/services/internal/testabilities/fixture_arc.go
```diff
@@ -88,23 +127,49 @@ func (f *arcFixture) IsUpAndRunning() {
...
+	f.transport.RegisterResponder("GET", "=~"+ArcURL+"/v1/tx/.*", func(req *http.Request) (*http.Response, error) {
+		txid := req.URL.String()[len(ArcURL+"/v1/tx/"):]
+		return f.knownTransactions[txid].toResponse()
+	})
+}
 
-		responder, err := httpmock.NewJsonResponder(200, content)
-		require.NoError(f, err)
+func (f *arcFixture) TxInfoJSON(id string) string {
+	_, content := f.knownTransactions[id].toResponseContent()
```
wregulski: What do you think about safe check here?
```go
tx, ok := f.knownTransactions[id]
require.True(f, ok, "known transaction not found: %s", id)
```

pkg/services/internal/arc/arc_service_test.go
```diff
@@ -38,15 +75,170 @@ func TestPostBEEFWithARCService(t *testing.T) {
...
+			given.ARC().IsUpAndRunning()
+
+			// and:
+			service := given.Services().NewArcService()
+
+			// when:
+			res, err := service.PostBeef(context.Background(), test.BEEF(t), nil)
+
+			// then:
+			assert.Error(t, err)
```
wregulski, line 188: What do you think about:
`require.Error(t, err)` here and even checking:

`assert.Contains(t, err.Error(), "some expected error message")`?

As far as I know the test won't stop if you won't have error on line 188.

wregulski (approved):

pkg/services/internal/arc/arc_service_test.go
dorzepowski: I choose assert instead of require intentionally, because if there isn't an error, maybe there is something in result that would give me the clue why the test is failing (why there wasn't an error)

When it comes to error message checking, I don't like this approach, but good comment to think about.

Unfortunately I cannot use error.Is here as I need/want to wrap the ARC error inside it, and I can't wrap 2 errors. This is something I need to rethink entirely for this service, but it's rather a scope of different PR.

pkg/services/internal/testabilities/fixture_arc.go
dorzepowski: good point, maybe it's better to fail here if we cannot find it.

chris-4chain (approved):

pkg/services/internal/testabilities/fixture_arc.go
```diff
@@ -48,6 +59,34 @@ func (f *arcFixture) HttpClient() *resty.Client {
 	return client
 }
 
+func (f *arcFixture) WillAlwaysReturnStatus(httpStatus int) {
+	var details string
+	switch httpStatus {
+	case 400:
```
mgosek-4chain: To keep things consistent, I'd suggest to using status codes from the `http` package instead of defining them explicitly.

pkg/services/internal/arc/arc_broadcast.go
```diff
@@ -0,0 +1,54 @@
...
+
+	req := s.httpClient.R().
+		SetContext(ctx).
+		SetHeaders(s.broadcastHeaders).
+		SetResult(result).
+		SetError(arcErr)
+
+	req.SetBody(broadcastRequestBody{
+		RawTx: txHex,
+	})
```
mgosek-4chain: Is there a reason why `SetBody` wasn't included in the previous chain?

pkg/services/internal/arc/arc_broadcast.go
```diff
@@ -0,0 +1,54 @@
...
+
+	response, err := req.Post(s.broadcastURL)
+
```
mgosek-4chain: Redundant empty line.

pkg/services/internal/arc/arc_query_transaction.go
```diff
@@ -0,0 +1,46 @@
...
+	result := &TXInfo{}
+	arcErr := &APIError{}
+	req := s.httpClient.R().
+		SetContext(ctx).
+		SetResult(result).
+		SetError(arcErr).
+		SetPathParam("txID", txID)
+
+	response, err := req.Get(s.queryTxURL)
+
```
mgosek-4chain: Redundant empty line.

pkg/services/internal/arc/arc_broadcast.go
```diff
@@ -0,0 +1,54 @@
...
+		return result, nil
+	case http.StatusUnauthorized, http.StatusForbidden, http.StatusNotFound:
+		return nil, fmt.Errorf("arc returned unauthorized: %w", arcErr)
+	case StatusNotExtendedFormat:
+		return nil, fmt.Errorf("arc expects transaction in extended format: %w", arcErr)
+	case StatusFeeTooLow, StatusCumulativeFeeValidationFailed:
+		return nil, fmt.Errorf("arc rejected transaction because of wrong fee: %w", arcErr)
+	default:
+		return nil, fmt.Errorf("arc cannot process provided transaction: %w", arcErr)
+	}
```
mgosek-4chain, line 51: It's up to you, but in my opinion an empty line after each `return` statement makes the code a bit more readable.

```
	switch response.StatusCode() {
	case http.StatusOK:
		return result, nil

	case http.StatusUnauthorized, http.StatusForbidden, http.StatusNotFound:
		return nil, fmt.Errorf("arc returned unauthorized: %w", arcErr)

	case StatusNotExtendedFormat:
		return nil, fmt.Errorf("arc expects transaction in extended format: %w", arcErr)

	case StatusFeeTooLow, StatusCumulativeFeeValidationFailed:
		return nil, fmt.Errorf("arc rejected transaction because of wrong fee: %w", arcErr)

	default:
		return nil, fmt.Errorf("arc cannot process provided transaction: %w", arcErr)
	}
```

pkg/services/internal/testabilities/fixture_arc.go
```diff
@@ -48,6 +59,34 @@ func (f *arcFixture) HttpClient() *resty.Client {
...
+		details = "The request seems to be malformed and cannot be processed"
+	case 401:
+		details = "The request is not authorized"
+	case 403:
+		details = "The request is not authorized"
+	case 404:
+		details = "The requested resource could not be found"
```
mgosek-4chain, line 73: Why not just use `http.StatusText` instead of defining the results explicitly?

pkg/services/internal/testabilities/fixture_arc.go
```diff
@@ -48,6 +59,34 @@ func (f *arcFixture) HttpClient() *resty.Client {
...
+	case 500:
+		details = "The server encountered an internal error and was unable to complete your request"
+	}
+
+	f.transport.RegisterResponder("POST", "=~"+ArcURL+"/v1/tx.*", func(req *http.Request) (*http.Response, error) {
```
mgosek-4chain: `http.MethodPost`

pkg/services/internal/testabilities/fixture_known_transaction.go
```diff
@@ -0,0 +1,52 @@
...
+	return 200, map[string]any{
+		"blockHash":    t.blockHash,
+		"blockHeight":  t.blockHeight,
+		"competingTxs": nil,
+		"extraInfo":    "",
+		"merklePath":   t.merklePath,
+		"timestamp":    timestamp,
+		"txStatus":     t.status,
+		"txid":         t.txid,
+	}
```
mgosek-4chain: Same suggestions like in the previous comments about using status codes from the `http` pkg.

mgosek-4chain (approved):

pkg/services/internal/arc/arc_broadcast.go
dorzepowski: I thought that this increase readability - as it separates some steps of making request, but I'll change it.

pkg/services/internal/testabilities/fixture_arc.go
dorzepowski: Those are the details that are really returned from arc, so I want to be close to it's behaviour.

pkg/services/internal/arc/arc_broadcast.go
dorzepowski: It's a matter of taste (and actually I don't see a benefit from those additional lines. 
Sure, it splits visually each result, but also visually I feel like I'm loosing a context that it is switch-case

PR #200, opened 2025-05-09
feat: handle error in postBEEF on getting status of single tx.

dorzepowski:


wregulski (approved):

cain4chain (approved):

pkg/services/internal/arc/arc_service.go
```diff
@@ -107,6 +111,14 @@ func (s *Service) getTxIDResults(ctx context.Context, txInfo *TXInfo, txIDs []st
 	return seq.Map(txsData, toResultForPostTxID)
 }
 
+func (s *Service) getTxIDResults(ctx context.Context, txIDs []string) iter.Seq[results.PostTxID] {
+	txIDsWithMissingTxInfo := seq.FromSlice(txIDs)
+
+	txsData := internal.MapParallel(ctx, txIDsWithMissingTxInfo, s.getTransactionData)
```
chris-4chain, line 117: 🚀

pkg/services/internal/testabilities/fixture_arc.go
```diff
@@ -174,6 +176,90 @@ func (f *arcFixture) store(txHex string) {
...
+
+type arcQueryFixture struct {
+	testing.TB
+	parent *arcFixture
+	txID   string
+}
+
+func (a *arcQueryFixture) WillReturnDifferentTxID() {
+	tx := a.knownTransaction()
+	tx.txid = a.rotatedTxIdByNumberOfChars(7)
```
chris-4chain, line 193: Why 7?

pkg/services/internal/testabilities/fixture_arc.go
```diff
@@ -174,6 +176,90 @@ func (f *arcFixture) store(txHex string) {
...
+// rotatedTxIdByNumberOfChars will return rotated txid by number of chars
+// for example:
+// txid: 1234567890
+// rotatedTxIdByNumberOfChars(3) will return 4567890123
+func (a *arcQueryFixture) rotatedTxIdByNumberOfChars(number int) string {
+	start := a.txID[number:]
+	end := a.txID[:number]
+	rotated := start + end // storing in variable for easier debugging
+	return rotated
+}
```
chris-4chain, line 205: Maybe something simpler? 🤔 

```golang
func differentTxID(txID string) string {
	return "d1ff" + txID[4:]
}
```

chris-4chain (approved):

pkg/services/internal/testabilities/fixture_arc.go
dorzepowski: I don't know, I just pick some number, it doesn't matter, as this is only generating some different txId, 
I just choose a little bit longer number but it also could be 6, 3, or even 1, but probably the higher number the lower probability of making exactly the same txID

pkg/services/internal/testabilities/fixture_arc.go
dorzepowski: Maybe I'm wrong, but by rotating by 7 by which we cannot divide 64. I'm limiting possible collisions (that the result will be the same as input, to only 17 cases (where all the characters in the txID are the same).

PR #210, opened 2025-05-12
refactor: service config for infra

chris-4chain:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter


infra-config.example.yaml
```diff
@@ -54,7 +58,6 @@ wallet_services:
             USD: 1
         timestamp: {}
```
chris-4chain, line 59: Because `go generate` regenerates the file **and** `ToYAML` doesn't work for time objects, I cannot override it.
As we discussed with @dorzepowski - this is not crucial and we will fix it in the future.

wregulski (approved):

pkg/defs/arc.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+func (arc *ARC) validateCallbackURL() error {
+	if arc.CallbackURL == "" {
+		return nil
+	}
+
+	callbackURLString := arc.CallbackURL
+
+	if !explicitHTTPURLRegex.MatchString(callbackURLString) {
```
dorzepowski: Why not simly use url.Parse without additional regex check?
The same could be achieved by checking the schema in parsed url.

pkg/defs/services.go
```diff
@@ -0,0 +1,111 @@
...
+		ChaintracksFiatExchangeRatesUrl: "",  // TODO: implement me
+		Chaintracks:                     nil, // TODO: implement me
+		ArcConfig: ARC{
+			URL:   arcUrl,
+			Token: arcApiKey,
+		},
+	}
+}
+
+func networkSpecific(chain BSVNetwork) (taalApiKey, arcURL string) {
```
dorzepowski, line 101: I would rather do it in opposite way,

1. Create WalletServices config with network agnostic values
2. check the network
3. use setDefaultConfigForMainNet() or setDefaultConfigForTestNet()

pkg/storage/provider.go
```diff
@@ -64,10 +64,17 @@ func NewGORMProvider(logger *slog.Logger, config GORMProviderConfig, opts ...Pro
 		random = randomizer.New()
 	}
 
+	var services wdk.Services
+	if options.services == nil {
+		logger.Warn("services is not set, some actions may not work")
```
cain4chain, line 69: Wondering if we should have something more specific here? Or maybe add a note to later provide reference to some documentation?

cain4chain (approved):
Nice job, tests passing confirmed.

pkg/infra/config_test.go
```diff
@@ -16,6 +17,7 @@ func TestCaseInsensitiveEnums(t *testing.T) {
 	t.Setenv("TEST_BSV_NETWORK", "MAIN")
 	t.Setenv("TEST_LOGGING_LEVEL", "DeBug")
 	t.Setenv("TEST_LOGGING_HANDLER", "Text")
+	t.Setenv("TEST_WALLET_SERVICES_WHATS_ON_CHAIN_BSV_EXCHANGE_RATE_BASE", "euR")
```
cain4chain, line 20: "euR" -> "EUR" ?

pkg/infra/server.go
```diff
@@ -56,7 +60,7 @@ func NewServer(opts ...InitOption) (*Server, error) {
 		Chain:      cfg.BSVNetwork,
 		FeeModel:   cfg.FeeModel,
 		Commission: cfg.Commission,
-	})
+	}, storage.WithServices(activeServices))
```
dorzepowski: maybe it should be part of the `storage.GORMProviderConfig`,
now, it looks a little confusing, that part of config is passed as config and part is through options.

pkg/storage/provider.go
chris-4chain: Hmmm ... TS code doesn't have any warning. It just throws errors when service is needed but it's not set.

pkg/infra/config_test.go
chris-4chain: It's done by purpose. Here I test if provided enum strings are correctly set even though the casing is weird.

pkg/defs/arc.go
chris-4chain: Good point 👍 
FYI: I just copied the solution from spv-wallet 🙈

pkg/infra/server.go
chris-4chain: Cool 👍

pkg/defs/services.go
chris-4chain: As discussed - we (or you) will adjust it in the next PR(s)

pkg/defs/arc.go
dorzepowski: I thought that I already seen this unneeded check 😄

dorzepowski (approved):

PR #215, opened 2025-05-14
feat: aggregate post beef

chris-4chain:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter


pkg/wdk/services_aggregated_postebeef_result.go
```diff
@@ -0,0 +1,111 @@
...
+	aggregatedTxs := make(AggregatedPostBEEF)
+
+	for _, result := range results {
+		if !result.Success() {
+			continue
+		}
+
+		mapped := make(map[string]*PostedTxID)
+		for _, txIDResult := range result.PostedBEEFResult.TxIDResults {
+			mapped[txIDResult.TxID] = &txIDResult
```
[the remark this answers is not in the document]
chris-4chain: AI responds AI 🙈 
From perplexity:

>Go 1.22 introduced a significant change to for loop variable semantics, making loop variables have per-iteration scope instead of per-loop scope. This change means that each iteration of a loop creates new variables rather than reusing the same variable across all iterations.

cain4chain (approved):

wregulski (approved):

pkg/storage/internal/actions/process.go
```diff
@@ -226,3 +237,58 @@ func (p *process) getSendStatus(ctx context.Context, txID string) (wdk.SendWithR
...
+	err error,
+) {
+	aggregated := broadcastResults.Aggregated([]string{txID})
+	aggBroadcastResult, ok := aggregated[txID]
+	if !ok {
+		err = fmt.Errorf("failed to find aggregated result for txID %s", txID)
+		return
+	}
+
+	reviewActionResult := wdk.ReviewActionResult{
```
dorzepowski, line 254: what is ReviewAction?

pkg/wdk/services_aggregated_postebeef_result.go
```diff
@@ -0,0 +1,111 @@
...
+		}
+
+		for _, txid := range txids {
+			txIDResult, ok := mapped[txid]
+			if !ok {
+				continue
+			}
```
dorzepowski, line 84: hmmm... shouldn't we actually return an error, or panic - as it is unexpected behaviour I belive.

pkg/storage/internal/actions/process.go
chris-4chain: It's a part of public interface - it contains status, txID and compeetingtxs (in doubleSpend)
The horse as it is, everyone sees.
😄

pkg/wdk/services_aggregated_postebeef_result.go
chris-4chain: No it's not an error case. It could happen that some tx is not in the response. The `summarize` step, after checking all the results, will also create aggregated results for such transactions.

dorzepowski (approved):

PR #217, opened 2025-05-15
feat: create skeleton of the BRC-100 wallet.

dorzepowski:


chris-4chain (approved):

pkg/internal/sdk/key_deriver.go
```diff
@@ -0,0 +1,243 @@
...
+// It supports deriving public and private keys, symmetric keys, and revealing key linkages.
+type KeyDeriver struct {
+	rootKey *ec.PrivateKey
+}
+
+// NewKeyDeriver creates a new KeyDeriver instance with a root private key.
+// The root key can be either a specific private key or the special 'anyone' key.
+func NewKeyDeriver(privateKey *ec.PrivateKey) *KeyDeriver {
+	if privateKey == nil {
+		privateKey, _ = AnyoneKey()
```
wregulski, line 32: Maybe let's add an error check here with a panic?
```go
if err != nil {
	panic(err) // or return nil
}
```

pkg/internal/sdk/proto_wallet.go
```diff
@@ -0,0 +1,352 @@
...
+type ProtoWalletArgs struct {
+	Type       ProtoWalletArgsType
+	PrivateKey *ec.PrivateKey
+	KeyDeriver *KeyDeriver
+}
+
+// NewProtoWallet creates a new ProtoWallet from a private key or KeyDeriver
+func NewProtoWallet(rootKeyOrKeyDeriver ProtoWalletArgs) (*ProtoWallet, error) {
+	switch rootKeyOrKeyDeriver.Type {
+	case ProtoWalletArgsTypeKeyDeriver:
```
wregulski, line 39: I am not sure about this:
In `NewProtoWallet`, when handling the `ProtoWalletArgsTypeKeyDeriver` case, you're still passing `PrivateKey` into `NewKeyDeriver`, which seems unintended:

```go
case ProtoWalletArgsTypeKeyDeriver:
    return &ProtoWallet{
        keyDeriver: NewKeyDeriver(rootKeyOrKeyDeriver.PrivateKey), // 👈 not sure if it's intended??
    }, nil
```
If the caller sets Type to "keyDeriver", I’d expect it to actually use the KeyDeriver field instead of ignoring it.

wregulski (approved):

pkg/internal/sdk/key_deriver.go
dorzepowski: I forgot to write (those are the files copied from go-sdk - so they are rather out of the scope)

Secondary I made the same mistake, when reading the original code,
the AnyoneKey is returning (privKey, pubKey) so there is no error.

> Great minds think alike

pkg/internal/sdk/proto_wallet.go
dorzepowski: I forgot to write (those are the files copied from go-sdk - so they are rather out of the scope)

Secondary I found the same in original code and [write a comment in PR to go-sdk](https://github.com/bsv-blockchain/go-sdk/pull/170#discussion_r2090813563)
> Great minds think alike

PR #224, opened 2025-05-19
feat: monitor with distributed scheduler support

chris-4chain:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter


pkg/storage/monitor.go
```diff
@@ -0,0 +1,131 @@
...
+		Cronjob:  job,
+	}
+
+	d.logger.Info("Starting a task", slog.Any("task", taskName), slog.Any("interval", interval))
+	return nil
+}
+
+func (d *Daemon) singleTaskRunner(taskName defs.MonitorTask) func() {
+	return func() {
+		activeTask, ok := d.activeTasks[taskName]
```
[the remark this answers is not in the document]
chris-4chain: No, because tasks are run **after** Start is called.

pkg/storage/internal/tasks/check_for_proofs.go
```diff
@@ -0,0 +1,23 @@
...
+
+func NewCheckForProofsTask(logger *slog.Logger) TaskInterface {
+	return &CheckForProofsTask{
+		logger:  logging.Child(logger, "check_for_proofs"),
+		counter: 0,
+	}
+}
+
+func (t *CheckForProofsTask) Run() {
+	// TODO: implement
```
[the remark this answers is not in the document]
chris-4chain: Will be done in the next PR

pkg/storage/internal/tasks/check_for_proofs_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+	// when:
+	err := activeStorage.Monitor.Start(map[defs.MonitorTask]time.Duration{
+		defs.CheckForProofsMonitorTask: taskInterval,
+	})
+	require.NoError(t, err)
+
+	time.Sleep(3 * taskInterval)
+
+	activeTask, ok := activeStorage.Monitor.Get(defs.CheckForProofsMonitorTask)
```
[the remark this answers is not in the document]
chris-4chain: Good point 👍

pkg/defs/monitor.go
```diff
@@ -0,0 +1,116 @@
...
+
+// ParseMonitorTaskStr parses a string to a MonitorTask or returns an error
+func ParseMonitorTaskStr(task string) (MonitorTask, error) {
+	return parseEnumCaseInsensitive(task, CheckForProofsMonitorTask)
+}
+
+// TaskConfig defines configuration parameters for a monitoring task
+type TaskConfig struct {
+	Enabled         bool `mapstructure:"enabled"`
+	IntervalSeconds uint `mapstructure:"interval_seconds"`
```
dorzepowski, line 29: Can't it be simply a Duration?
Maybe we could use [ISO 8601 string](https://en.wikipedia.org/wiki/ISO_8601#Durations) in yaml for duration

pkg/defs/monitor.go
```diff
@@ -0,0 +1,116 @@
...
+
+// TasksConfig is a map of monitoring tasks with their configuration parameters
+// This is a struct that contains fields for each MonitorTask/
+// Make sure each field has a mapstructure tag that matches the MonitorTask enum value.
+type TasksConfig struct {
+	CheckForProofs TaskConfig `mapstructure:"check_for_proofs"`
+}
+
+// All returns a map where each MonitorTask key is paired with its corresponding TaskConfig from the TasksConfig struct.
+func (t *TasksConfig) All() map[MonitorTask]TaskConfig {
```
dorzepowski: you could return iter.Seq2 so it would process it, in one run, without temporary storage in map

dorzepowski: Does it need to be public?

jakubmkowalski (approved):

pkg/infra/config_test.go
```diff
@@ -207,3 +207,42 @@ func TestValidTimestamp(t *testing.T) {
 	expected := time.Date(2023, time.December, 13, 0, 0, 0, 0, time.UTC)
 	require.Equal(t, expected, infraSrv.Config.Services.WhatsOnChain.BSVExchangeRate.Timestamp)
 }
+
+func TestValidMonitor(t *testing.T) {
+	// given:
+	t.Setenv("TEST_SERVER_PRIVATE_KEY", fixtures.StorageServerPrivKey)
```
dorzepowski: do you need this for this line for those new tests?

pkg/storage/monitor.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	scheduler, err := gocron.NewScheduler(gocron.WithDistributedLocker(locker))
+	if err != nil {
+		return nil, fmt.Errorf("failed to create scheduler: %w", err)
+	}
+
+	return &Daemon{
+		scheduler:   scheduler,
+		logger:      logging.Child(logger, "monitor"),
```
dorzepowski: please add worker name to logger args

pkg/storage/monitor.go
```diff
@@ -0,0 +1,131 @@
...
+
+// Get retrieves the active monitoring task associated with the given name.
+// Returns the ActiveTask pointer and true if found, otherwise nil and false.
+func (d *Daemon) Get(name defs.MonitorTask) (*ActiveTask, bool) {
+	task, ok := d.activeTasks[name]
+	return task, ok
+}
+
+func (d *Daemon) initializeTask(taskName defs.MonitorTask, interval time.Duration) error {
+	taskCreator, ok := tasks.All[taskName]
```
dorzepowski: ### minor
taskFactory

pkg/storage/monitor.go
dorzepowski: Why do we need this at all?
why not pass the ActiveTask here as an argument. We could even add a taskName to ActiveTask struct, to have single argument.

pkg/storage/provider.go
```diff
@@ -20,7 +20,8 @@ import (
 
 // Provider is a storage provider.
 type Provider struct {
-	Chain defs.BSVNetwork
+	Chain   defs.BSVNetwork
+	Monitor *Daemon
```
dorzepowski: I'm pretty sure that in ts, they don't bind monitor to Provider
it's rather opposite way - Monitor has Storage instance to work with.

pkg/storage/internal/tasks/check_for_proofs_test.go
```diff
@@ -0,0 +1,51 @@
...
+
+	// and:
+	activeTask, ok := activeStorage.Monitor.Get(defs.CheckForProofsMonitorTask)
+	require.True(t, ok)
+
+	// then:
+	ensureTaskExecutedInTime(t, activeTask, taskInterval)
+}
+
+func ensureTaskExecutedInTime(t testing.TB, activeTask *storage.ActiveTask, taskInterval time.Duration) {
```
dorzepowski: this looks like a great candidate for custom assertions with Then component.

pkg/storage/internal/tasks/check_for_proofs_test.go
```diff
@@ -0,0 +1,51 @@
...
+	zeroTime := time.Time{}
+	timeout := time.Now().Add(5 * taskInterval)
+	for time.Now().Before(timeout) {
```
dorzepowski: ### just saying, to share other possible options

the same could be achieved with time.Ticker
or with `seq.Tick` + `seq.TakeWhile` 😄

pkg/storage/internal/tasks/check_for_proofs_test.go
```diff
@@ -0,0 +1,51 @@
...
+		defs.CheckForProofsMonitorTask: taskInterval,
+	})
+	require.NoError(t, err)
+
+	// and:
+	activeTask, ok := activeStorage.Monitor.Get(defs.CheckForProofsMonitorTask)
+	require.True(t, ok)
+
+	// then:
+	ensureTaskExecutedInTime(t, activeTask, taskInterval)
```
dorzepowski: I'm not sure if this is what we want to check, 
for this PR, it could be enough, but in the future PR,
I would replace it with ensuring that the state was changed by monitor (for example, that the transaction changed status, etc.)

pkg/defs/monitor.go
chris-4chain: I thought it is simpler to limit "a user" to just seconds. 
I think, this is more self-documenting and everyone will understand what to set.

Besides, this implicitly limit a minimum interval to one second - which is good, because monitor tasks won't work well in "small" periods (e.g. 100ms)

pkg/defs/monitor.go
chris-4chain: Good point 👍

chris-4chain: No - changed to not-exported

pkg/infra/config_test.go
chris-4chain: Yes, because otherwise validation fails.
But it's a fair point - I'm adding following helper function:
```golang
// setRequiredEnvs sets necessary environment variables for test configuration.
// It ensures TEST_SERVER_PRIVATE_KEY is set with a valid private key value for proper test initialization.
func setRequiredEnvs(t *testing.T) {
	t.Setenv("TEST_SERVER_PRIVATE_KEY", fixtures.StorageServerPrivKey)
}
```

pkg/storage/monitor.go
chris-4chain: I would argue with that, but I don't want to 🙈 
Check out current implementation 😄

pkg/storage/provider.go
chris-4chain: It's not such simple - it needs to be discussed.

pkg/storage/internal/tasks/check_for_proofs_test.go
chris-4chain: see your last comment 👇

pkg/storage/internal/tasks/check_for_proofs_test.go
chris-4chain: I've made following implementation. But, for me, it's harder to read than my initial implementation, so I'll go with that.

```go
	seq.Collect(
		seq.TakeWhile(seq.Tick(taskInterval/10), func(tim time.Time) bool {
			require.True(t, time.Now().Before(timeout), "timeout reached")

			lastRun, err := activeTask.Cronjob.LastRun()
			require.NoError(t, err)

			if lastRun == zeroTime {
				return true
			}

			require.True(t, lastRun.After(time.Now().Add(-2*taskInterval)))
			return false
		}),
	)
```

pkg/storage/internal/tasks/check_for_proofs_test.go
chris-4chain: Yes, for this PR, I wanted to check something, to make sure if it works.
In the next PR we will adjust this and provide better tailored testabilities and assertions.

pkg/storage/provider.go
chris-4chain: It wasn't simple, because of GORM lock but I managed to decoupled strage and monitor. 
Check it out :)

pkg/infra/server.go
```diff
@@ -65,16 +67,30 @@ func NewServer(opts ...InitOption) (*Server, error) {
 		return nil, fmt.Errorf("failed to create storage provider: %w", err)
 	}
 
-	_, err = activeStorage.Migrate(context.Background(), cfg.Name, storageIdentityKey)
+	_, err = activeStorage.Migrate(context.TODO(), cfg.Name, storageIdentityKey)
```
dorzepowski, line 70: why you changed context?

chris-4chain: Because I think, the context should be provided as `NewServer` param.

dorzepowski (approved):

PR #234, opened 2025-05-22
[FEATURE] listOutputs on sql storage

### LINKED ISSUE #24: [FEATURE] listOutputs on sql storage

ac4ch:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter

### DIFF
--- a/pkg/internal/validate/list_outputs_args.go
+++ b/pkg/internal/validate/list_outputs_args.go
@@ -1,0 +1,38 @@
+package validate
+
+import (
+	"fmt"
+
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/wdk"
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/wdk/primitives"
+)
+
+const (
+	MaxPaginationLimit  = 10000
+	MaxPaginationOffset = 1_000_000
+	MinPaginationLimit  = 1
+)
+
+func ListOutputsArgs(args *wdk.ListOutputsArgs) error {
+	if args == nil {
+		return fmt.Errorf("args cannot be nil")
+	}
+
+	if args.Limit < MinPaginationLimit {
+		return fmt.Errorf("limit must be greater than 0")
+	}
+	if args.Limit > MaxPaginationLimit {
+		return fmt.Errorf("limit exceeds max allowed value of %d", MaxPaginationLimit)
+	}
+	if args.Offset > MaxPaginationOffset {
+		return fmt.Errorf("offset is too large")
+	}
+
+	for _, txid := range args.KnownTxids {
+		if err := primitives.TXIDHexString(txid).Validate(); err != nil {
+			return fmt.Errorf("invalid txid: %w", err)
+		}
+	}
+
+	return nil
+}
--- a/pkg/storage/internal/methodtests/list_outputs_test.go
+++ b/pkg/storage/internal/methodtests/list_outputs_test.go
@@ -1,0 +1,119 @@
+package methodtests
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/internal/fixtures/testusers"
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/randomizer"
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/storage/internal/testabilities"
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/wdk"
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/wdk/primitives"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestListOutputs_EmptyByDefault(t *testing.T) {
+	// given:
+	ctx := t.Context()
+	given := testabilities.Given(t)
+	activeStorage := given.Provider().GORM()
+
+	args := wdk.ListOutputsArgs{
+		Basket: "",
+		Limit:  10,
+		Offset: 0,
+	}
+
+	// when:
+	actualResult, err := activeStorage.ListOutputs(ctx, testusers.Alice.AuthID(), args)
+
+	expectedResult := &wdk.ListOutputsResult{
+		TotalOutputs: 0,
+		BEEF:         nil,
+		Outputs:      []*wdk.WalletOutput{},
+	}
+	// then:
+	require.NoError(t, err)
+	require.NotNil(t, actualResult)
+	require.Equal(t, expectedResult, actualResult)
+}
+
+func TestListOutputs_WithKnownTxidsAndIncludeTransactions(t *testing.T) {
+	// Given:
+	ctx := t.Context()
+	given := testabilities.Given(t)
+	activeStorage := given.Provider().WithRandomizer(randomizer.NewTestRandomizer()).GORM()
+
+	actionResult, signedTx := given.ActionCreatedAndSigned(activeStorage)
+	txid := signedTx.TxID().String()
+
+	processArgs := wdk.ProcessActionArgs{
+		IsNewTx:    true,
+		IsSendWith: false,
+		IsNoSend:   false,
+		IsDelayed:  false,
+		Reference:  &actionResult.Reference,
+		TxID:       (*primitives.TXIDHexString)(&txid),
+		RawTx:      signedTx.Bytes(),
+		SendWith:   []string{},
+	}
+	_, err := activeStorage.ProcessAction(ctx, testusers.Alice.AuthID(), processArgs)
+	require.NoError(t, err)
+
+	listArgs := wdk.ListOutputsArgs{
+		Basket:              "",
+		Limit:               10,
+		Offset:              0,
+		KnownTxids:          []string{txid},
+		IncludeTransactions: true,
+	}
+
+	// When:
+	actualResult, err := activeStorage.ListOutputs(ctx, testusers.Alice.AuthID(), listArgs)
+
+	// Then:
+	require.NoError(t, err)
+	require.NotNil(t, actualResult)
+	require.NotEmpty(t, actualResult.Outputs)
+	require.NotNil(t, actualResult.BEEF)
+
+	expectedOutpoints := map[string]bool{}
+	for _, out := range actualResult.Outputs {
+		parts := strings.Split(string(out.Outpoint), ".")
+		require.Len(t, parts, 2, "Outpoint format should be <txid>.<vout>")
+		require.Equal(t, txid, parts[0], "Output txid should match known txid")
+		expectedOutpoints[string(out.Outpoint)] = true
+	}
+
+	for _, out := range actualResult.Outputs {
+		assert.Truef(t, expectedOutpoints[string(out.Outpoint)], "Unexpected outpoint: %s", out.Outpoint)
+	}
+}
+
+func TestListOutputs_FilterByBasketName(t *testing.T) {
+	// Given:
+	ctx := t.Context()
+	given := testabilities.Given(t)
+	activeStorage := given.Provider().WithRandomizer(randomizer.NewTestRandomizer()).GORM()
+
+	user := testusers.Alice
+	faucet := given.Faucet(activeStorage, user)
+	_, _ = faucet.TopUp(1000)
+
+	basketName := wdk.BasketNameForChange
+
+	// When:
+	listArgs := wdk.ListOutputsArgs{
+		Basket: primitives.StringUnder300(basketName),
+		Limit:  10,
+		Offset: 0,
+	}
+	actualResult, err := activeStorage.ListOutputs(ctx, user.AuthID(), listArgs)
+
+	// Then:
+	require.NoError(t, err)
+	require.NotNil(t, actualResult)
+	require.NotEmpty(t, actualResult.Outputs, "Expected outputs for basket %s", basketName)
+	assert.Greater(t, int(actualResult.TotalOutputs), 0, "Expected totalOutputs > 0 for basket %s", basketName)
+}
--- a/pkg/storage/internal/repo/all.go
+++ b/pkg/storage/internal/repo/all.go
@@ -1,6 +1,8 @@
 package repo
 
-import "gorm.io/gorm"
+import (
+	"gorm.io/gorm"
+)
 
 type Repositories struct {
 	*Migrator
--- a/pkg/storage/internal/repo/outputs.go
+++ b/pkg/storage/internal/repo/outputs.go
@@ -6,6 +6,7 @@ import (
 	"iter"
 
 	"github.com/4chain-ag/go-wallet-toolbox/pkg/storage/internal/database/models"
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/storage/internal/entity"
 	"github.com/4chain-ag/go-wallet-toolbox/pkg/wdk"
 	"github.com/go-softwarelab/common/pkg/seq"
 	"github.com/go-softwarelab/common/pkg/slices"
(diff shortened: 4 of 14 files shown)


pkg/storage/internal/database/models/transaction.go
```diff
@@ -24,3 +24,7 @@ type Transaction struct {
 	Labels        []*Label    `gorm:"many2many:transaction_labels;"`
 	ReservedUtxos []*UserUTXO `gorm:"foreignKey:ReservedByID"`
 }
+
+func (Transaction) TableName() string {
```
dorzepowski: Why this change?

ac4ch:
```go
if len(filter.KnownTxids) > 0 {
		query = query.Joins("JOIN "+(models.Transaction{}).TableName()+" ON "+
			(models.Transaction{}).TableName()+".id = bsv_outputs.transaction_id").
			Preload("Transaction").
			Where((models.Transaction{}).TableName()+".tx_id IN ?", filter.KnownTxids)
	}
```
so i can use it here instead of string

pkg/storage/internal/repo/all.go
```diff
@@ -29,3 +35,11 @@ func NewSQLRepositories(db *gorm.DB) *Repositories {
...
 	return repositories
 }
+
+func (r *Repositories) ListAndCountOutputs(ctx context.Context, userID int, filter ListOutputsFilter) ([]wdk.TableOutput, int, error) {
+	return r.Outputs.ListAndCountOutputs(ctx, userID, filter)
+}
+
+func (r *Repositories) GetBEEFForTxids(ctx context.Context, txids []string) (*primitives.BEEF, error) {
+	return r.ProvenTxReq.GetBEEFForTxids(ctx, txids)
+}
```
chris-4chain: Please don't add methods on `Repositories` aggregation struct.

You can call instead.
```golang
	outputModels, totalCount, err := p.repo.Outputs.ListAndCountOutputs(ctx, *auth.UserID, filterOptions)
```

pkg/storage/provider.go
```diff
@@ -295,3 +296,43 @@ func (p *Provider) ProcessAction(ctx context.Context, auth wdk.AuthID, args wdk.
...
 	return res, nil
 }
+
+// ListOutputs will list outputs with provided args
+func (p *Provider) ListOutputs(ctx context.Context, auth wdk.AuthID, args wdk.ListOutputsArgs) (*wdk.ListOutputsResult, error) {
+	if auth.UserID == nil {
+		return nil, fmt.Errorf("access is denied due to an authorization error")
+	}
+
+	filterOptions := listOutputsArgsToFilterParams(args)
```
chris-4chain: Please add validation of args to `validation` package (with appropriate unit tests)

See for reference, how to do it: `CreateAction` or `ProcessAction` - above in this file

pkg/storage/internal/repo/outputs.go
```diff
@@ -96,3 +96,52 @@ func (o *Outputs) mapModelToTableOutput(model *models.Output) *wdk.TableOutput {
...
+	IncludeTransactions       bool
+	IncludeCustomInstructions bool
+	IncludeTags               bool
+	IncludeLabels             bool
+	Limit                     int
+	Offset                    int
+	KnownTxids                []string
+}
+
+func (o *Outputs) ListAndCountOutputs(ctx context.Context, userID int, filter ListOutputsFilter) ([]wdk.TableOutput, int, error) {
```
chris-4chain: If your repository method needs to get an object instead of "atomic" arguments, please place the "model" into: `/storage/internal/entity` package.

At least, there is such convention in this project

pkg/storage/internal/repo/outputs.go
```diff
@@ -96,3 +96,52 @@ func (o *Outputs) mapModelToTableOutput(model *models.Output) *wdk.TableOutput {
...
+	var rows []*models.Output
+
+	query := o.db.WithContext(ctx).Model(&models.Output{}).Where("bsv_outputs.user_id = ?", userID)
```
chris-4chain: Please make list and count queries in one DB-transaction:

```golang
err := txs.db.WithContext(ctx).Transaction(func(tx *gorm.DB) (err error) {
		// your code here
	})
	if err != nil {
		return fmt.Errorf("failed to <sth>: %w", err)
	}
```

pkg/storage/internal/repo/outputs.go
```diff
@@ -96,3 +96,52 @@ func (o *Outputs) mapModelToTableOutput(model *models.Output) *wdk.TableOutput {
...
+		query = query.Joins("JOIN output_baskets ON bsv_outputs.basket_id = output_baskets.basket_id").
+			Where("output_baskets.name = ?", filter.Basket)
+	}
+
+	if len(filter.KnownTxids) > 0 {
+		query = query.Joins("JOIN "+(models.Transaction{}).TableName()+" ON "+
+			(models.Transaction{}).TableName()+".id = bsv_outputs.transaction_id").
+			Preload("Transaction").
+			Where((models.Transaction{}).TableName()+".tx_id IN ?", filter.KnownTxids)
+	}
```
chris-4chain: I don't like these joins. Could you try to do it more elegant?
Or, please tell me, what you want to achieve - I could help.

pkg/storage/internal/repo/outputs.go
```diff
@@ -96,3 +96,52 @@ func (o *Outputs) mapModelToTableOutput(model *models.Output) *wdk.TableOutput {
...
+	if err := query.Limit(filter.Limit).Offset(filter.Offset).Order("bsv_outputs.id ASC").Find(&rows).Error; err != nil {
+		return nil, 0, fmt.Errorf("query failed: %w", err)
+	}
+
+	mapped := slices.Map(rows, o.mapModelToTableOutput)
+	outputs := slices.Map(mapped, func(p *wdk.TableOutput) wdk.TableOutput {
+		return *p
+	})
+
+	return outputs, int(total), nil
```
chris-4chain: General comment.

We already have some tooling to do pagination stuff. 
Could you try to align with this?

See:
- `ListAndCountCertificates()`
- `scopes.Paginate(page)`

pkg/storage/internal/repo/outputs.go
```diff
@@ -96,3 +96,52 @@ func (o *Outputs) mapModelToTableOutput(model *models.Output) *wdk.TableOutput {
...
+	}
+
+	if err := query.Limit(filter.Limit).Offset(filter.Offset).Order("bsv_outputs.id ASC").Find(&rows).Error; err != nil {
+		return nil, 0, fmt.Errorf("query failed: %w", err)
+	}
+
+	mapped := slices.Map(rows, o.mapModelToTableOutput)
+	outputs := slices.Map(mapped, func(p *wdk.TableOutput) wdk.TableOutput {
+		return *p
+	})
```
chris-4chain: Hmm, I think, you could return a slice of pointers (`[]*wdk.TableOutput`) instead of a slice of values (`[]wdk.TableOutput`).

Thanks to this, the mapping wouldn't be necessary and, maybe, it would be more performant.

pkg/storage/internal/repo/proven_tx_req.go
```diff
@@ -168,3 +169,19 @@ func (p *ProvenTxReq) recursiveBuildValidBEEF(ctx context.Context, depth int, me
 	// Result is in mergeToBeef
 	return nil
 }
+
+func (p *ProvenTxReq) GetBEEFForTxids(ctx context.Context, txids []string) (*primitives.BEEF, error) {
```
chris-4chain: In repository level we try to avoid alias types that are in directory (`wdk` or `primitives`).

If its `[]byte` just return `[]byte`.

chris-4chain (changes requested):

pkg/storage/internal/repo/outputs.go
ac4ch: are u referring here to `ListOutputsFilter` to be moved under entity package  ?

chris-4chain: yes

pkg/storage/internal/repo/outputs.go
ac4ch: I used `join` to filter outputs by `tx_id` which isn't accessible directly on `bsv_outputs` table so needed `txid` to construct the `Outpoint` which is only available through the `Transaction`

I added `.Preload("Transaction")` to join that why I can filter while still loading related `Transaction` needed for result 

I started adjusting this when tests failed to compare `txid`

pkg/internal/validate/list_outputs_args.go
```diff
@@ -0,0 +1,33 @@
...
+	if args.Limit == 0 {
+		return fmt.Errorf("limit must be greater than 0")
+	}
+	if args.Limit > 10000 {
+		return fmt.Errorf("limit exceeds max allowed value of 10000")
+	}
+
+	if args.Offset > 1_000_000 {
+		return fmt.Errorf("offset is too large")
+	}
```
chris-4chain: Please add some constants for these magic numbers. Most probably we should use them also for other pagination endpoints.

pkg/storage/internal/database/models/output.go
```diff
@@ -37,3 +40,40 @@ type Output struct {
...
 }
+
+type ListOutputsParams struct {
+	UserID        int
+	BasketName    string
+	KnownTxids    []string
+	Limit, Offset int
+}
+
+func ListAndCountOutputs(ctx context.Context, db *gorm.DB, params ListOutputsParams) ([]*Output, int64, error) {
```
chris-4chain: This logic should be in the repository package.

pkg/storage/internal/database/models/output.go
```diff
@@ -37,3 +40,40 @@ type Output struct {
...
+	if err := db.WithContext(ctx).Transaction(func(tx *gorm.DB) error {
+		query := tx.Model(&Output{}).Where("bsv_outputs.user_id = ?", params.UserID)
+
+		if params.BasketName != "" {
+			query = query.Joins("JOIN output_baskets ON bsv_outputs.basket_id = output_baskets.basket_id").Where("output_baskets.name = ?", params.BasketName)
+		}
+
+		if len(params.KnownTxids) > 0 {
+			query = query.Joins("JOIN bsv_transactions ON bsv_transactions.id = bsv_outputs.transaction_id").Preload("Transaction").Where("bsv_transactions.tx_id IN ?", params.KnownTxids)
+		}
```
chris-4chain: Consider sth like this:

```golang
		if params.BasketName != "" {
			query = query.Joins("Basket", db.Where(&OutputBasket{Name: params.BasketName}))
		}

		if len(params.KnownTxids) > 0 {
			query = query.Where(
				"transaction_id IN (?)",
				db.Model(&Transaction{}).
					Select("id").
					Where("tx_id IN ?", params.KnownTxids),
			).Preload("Transaction")
		}
```

Tests pass :) But I'm not sure about `BasketName` because I couldn't find a test for that.

pkg/storage/internal/methodtests/list_outputs_test.go
```diff
@@ -0,0 +1,92 @@
...
+		Reference:  &actionResult.Reference,
+		TxID:       (*primitives.TXIDHexString)(&txid),
+		RawTx:      signedTx.Bytes(),
+		SendWith:   []string{},
+	}
+	_, err := activeStorage.ProcessAction(ctx, testusers.Alice.AuthID(), processArgs)
+	require.NoError(t, err)
+
+	listArgs := wdk.ListOutputsArgs{
+		Basket:              "",
```
chris-4chain, line 65: I can't see a test to check filtering for 'BasketName'

pkg/storage/internal/database/models/output.go
```diff
@@ -37,3 +40,40 @@ type Output struct {
...
+	KnownTxids    []string
+	Limit, Offset int
+}
+
+func ListAndCountOutputs(ctx context.Context, db *gorm.DB, params ListOutputsParams) ([]*Output, int64, error) {
+	var outputs []*Output
+	var total int64
+
+	if err := db.WithContext(ctx).Transaction(func(tx *gorm.DB) error {
+		query := tx.Model(&Output{}).Where("bsv_outputs.user_id = ?", params.UserID)
```
dorzepowski: why you use table name not simply `user_id = ?`?

pkg/storage/internal/database/models/output.go
dorzepowski: I prefer this solution, hope it will work for basket name 🤞

pkg/storage/provider.go
```diff
@@ -295,3 +296,48 @@ func (p *Provider) ProcessAction(ctx context.Context, auth wdk.AuthID, args wdk.
 	}
 	return res, nil
 }
+
+// ListOutputs will list outputs with provided args
+func (p *Provider) ListOutputs(ctx context.Context, auth wdk.AuthID, args wdk.ListOutputsArgs) (*wdk.ListOutputsResult, error) {
+	if auth.UserID == nil {
+		return nil, fmt.Errorf("access is denied due to an authorization error")
```
dorzepowski: ### minor

maybe it's good ideat to use the same message as in others

pkg/storage/mappings.go
```diff
@@ -80,3 +85,63 @@ func certificateModelFieldsToFieldsResult(fields []*models.CertificateField) map
...
+		Offset:                    offset,
+		KnownTxids:                args.KnownTxids,
+	}
+}
+
+func outputModelToResult(m *wdk.TableOutput) *wdk.WalletOutput {
+	satoshis := uint64(0)
+	if m.Satoshis > 0 {
+		satoshis = uint64(m.Satoshis)
+	}
```
dorzepowski: ### small suggestion

```suggestion
	satoshis := to.NoLessThan(m.Satoshis, 0)
```

pkg/storage/mappings.go
```diff
@@ -80,3 +85,63 @@ func certificateModelFieldsToFieldsResult(fields []*models.CertificateField) map
...
+		Satoshis:  primitives.SatoshiValue(satoshis),
+		Spendable: m.Spendable,
+	}
+
+	if m.TxID != nil {
+		outpoint := fmt.Sprintf("%s.%d", *m.TxID, m.Vout)
+		result.Outpoint = primitives.OutpointString(outpoint)
+	}
+
+	if m.CustomInstructions != nil {
```
dorzepowski: is this `if` actually needed?

pkg/storage/mappings.go
```diff
@@ -80,3 +85,63 @@ func certificateModelFieldsToFieldsResult(fields []*models.CertificateField) map
...
+		result.CustomInstructions = m.CustomInstructions
+	}
+
+	if m.LockingScript != nil {
+		s := primitives.HexString(*m.LockingScript)
+		result.LockingScript = &s
```
dorzepowski: ### minor suggestion
```suggestion
		result.LockingScript = to.Ptr(primitives.HexString(*m.LockingScript))
```

pkg/storage/internal/database/models/output.go
chris-4chain: Maybe even without `.Preload("Transaction")` 🤔

pkg/storage/mappings.go
```diff
@@ -80,3 +85,63 @@ func certificateModelFieldsToFieldsResult(fields []*models.CertificateField) map
...
+		limit = int(limitVal)
+	}
+
+	var offset int
+	offsetVal := uint64(args.Offset)
+	if offsetVal > uint64(math.MaxInt) {
+		offset = math.MaxInt
+	} else {
+		offset = int(offsetVal)
+	}
```
dorzepowski: ### minor suggestion
```suggestion
	limit := must.ConvertToIntFromUnsigned(to.NoMoreThan(args.Limit, 10000))
	offset := must.ConvertToIntFromUnsigned(to.NoMoreThan(args.Offset, math.MaxInt))
```

pkg/storage/internal/database/models/output.go
ac4ch: without `.Preload("Transaction")` test is failing 
```
Error Trace:    /Users/augustynchmiel/4chain/spv/toolbox/go-wallet-toolbox/pkg/storage/internal/methodtests/list_outputs_test.go:84
Error:          "[]" should have 2 item(s), but has 1
Test:           TestListOutputs_WithKnownTxidsAndIncludeTransactions
Messages:       Outpoint format should be <txid>.<vout>
```

pkg/storage/provider.go
ac4ch: as in others errors in this method you mean ?

pkg/storage/mappings.go
ac4ch: i think it needed as this can be `nil` `field CustomInstructions *string 'json:"customInstructions,omitempty"'`

pkg/storage/provider.go
```diff
@@ -131,7 +138,7 @@ func (p *Provider) MakeAvailable(ctx context.Context) (*wdk.TableSettings, error
 // InsertCertificateAuth inserts certificate to the database for authenticated user
 func (p *Provider) InsertCertificateAuth(ctx context.Context, auth wdk.AuthID, certificate *wdk.TableCertificateX) (uint, error) {
 	if auth.UserID == nil || certificate.UserID != *auth.UserID {
-		return 0, fmt.Errorf("access is denied due to an authorization error")
+		return 0, errors.New(ErrMsgAccessDenied)
```
dorzepowski: instead of just message, create a whole error
and use `fmt.Errorf` because we agreed that we will try to use only `fmt.Errorf` for errors creation even without any format arguments
so please, try to avoid using errors.New

pkg/storage/mappings.go
dorzepowski: but you're assigning exactly CustomInstructions to the result.CustomInstructions - no additional conversion or something.
In other words, if m.CustomInstructions is nil
This: 
```go
if m.CustomInstructions != nil {
  result.CustomInstructions = m.CustomInstructions
}
```

And This:

```go
result.CustomInstructions = m.CustomInstructions
```

Will have the same result:
```go
result.CustomInstructions == nil // true
```

that's why IMHO you should set the CustomInstructions directly from `m` during result initialization (so the line, lets say 117)

pkg/storage/internal/repo/outputs.go
```diff
@@ -96,3 +97,50 @@ func (o *Outputs) mapModelToTableOutput(model *models.Output) *wdk.TableOutput {
...
+
+func listAndCountOutputs(ctx context.Context, db *gorm.DB, params entity.ListOutputsParams) ([]*models.Output, int64, error) {
+	var outputs []*models.Output
+	var total int64
+
+	if err := db.WithContext(ctx).Transaction(func(tx *gorm.DB) error {
+		query := tx.Model(&models.Output{}).Where("user_id = ?", params.UserID)
+
+		if params.BasketName != "" {
+			query = query.Where("basket_id IN (?)", db.Model(&models.OutputBasket{}).Select("basket_id").Where("name = ?", params.BasketName))
```
chris-4chain: There could be several rows with the same name - but we are interested in provided-user baskets.
The output is the same (I believe), but there are less "unnecessary" IDs provided to `IN` clause. 

```suggestion
			query = query.Where("basket_id IN (?)", db.Model(&models.OutputBasket{}).Select("basket_id").Where("name = ? and user_id = ?", params.BasketName, params.UserID))
```

chris-4chain (approved):

dorzepowski (approved):

PR #235, opened 2025-05-22
feat: check for proofs task

chris-4chain:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter

### DIFF
--- a/pkg/monitor/interface.go
+++ b/pkg/monitor/interface.go
@@ -2,7 +2,7 @@ package monitor
 
 import "context"
 
-// MinimalStorageInterface defines the minimum storage functionality used by the monitor.
-type MinimalStorageInterface interface {
+// MonitoredStorage defines the minimum storage functionality used by the monitor.
+type MonitoredStorage interface {
 	SynchronizeTransactionStatuses(ctx context.Context) error
 }
--- a/pkg/monitor/interface.go
+++ b/pkg/monitor/interface.go
@@ -1,0 +1,8 @@
+package monitor
+
+import "context"
+
+// MinimalStorageInterface defines the minimum storage functionality used by the monitor.
+type MinimalStorageInterface interface {
+	SynchronizeTransactionStatuses(ctx context.Context) error
+}
--- a/pkg/monitor/internal/testabilities/fixture_monitor.go
+++ b/pkg/monitor/internal/testabilities/fixture_monitor.go
@@ -17,16 +17,19 @@ type MonitorFixture interface {
 
 func Given(t testing.TB) MonitorFixture {
 	return &monitorFixture{
-		t:       t,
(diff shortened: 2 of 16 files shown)


mgosek-4chain (approved):

jakubmkowalski (approved):

pkg/monitor/interface.go
```diff
@@ -0,0 +1,8 @@
+package monitor
+
+import "context"
+
+// MinimalStorageInterface defines the minimum storage functionality used by the monitor.
+type MinimalStorageInterface interface {
```
dorzepowski: maybe `MonitorStorage` or `MonitoredStorage`,

pkg/monitor/monitor.go
```diff
@@ -130,15 +142,20 @@ func (d *Daemon) initializeTask(taskName defs.MonitorTask, interval time.Duratio
...
+			}
 			if activeTask.Cronjob == nil {
 				return
 			}
 			nextRun, _ := activeTask.Cronjob.NextRun()
 			d.logger.Info("Finish task", slog.Any("task", activeTask.TaskName), slog.Any("next_run", nextRun))
 		}()
 
-		activeTask.Instance.Run()
+		err = activeTask.Instance.Run(context.TODO())
```
dorzepowski: instead of context.TODO, you can simply return func(context.Context) from singleTaskRunner and use that context.

pkg/monitor/internal/testabilities/fixture_monitor.go
```diff
@@ -17,16 +17,19 @@ type MonitorFixture interface {
 
 func Given(t testing.TB) MonitorFixture {
 	return &monitorFixture{
-		t:       t,
-		logger:  logging.NewTestLogger(t),
-		require: require.New(t),
+		t:           t,
+		logger:      logging.NewTestLogger(t),
+		require:     require.New(t),
+		mockStorage: &MockStorage{},
```
dorzepowski, line 23: # Thoughts for future
Maybe we should define the interface for "local" storage which could have those synchronisation methods, and the storage mock that we already have in `internal/mock` could be created from it, not from the `WalletStorageWriter`

pkg/monitor/monitor.go
chris-4chain: Nope, this is not what I wanted. I was thinking about 
- the whole cancellation process (of cron) 
- and timeout for single task run

pkg/monitor/internal/testabilities/fixture_monitor.go
chris-4chain: 🤷 
Noted

dorzepowski (approved):

PR #237, opened 2025-05-22
refactor: repackage storage internals

dorzepowski:


pkg/internal/storage/funder/sql_test.go
```diff
@@ -6,7 +6,7 @@ import (
 
 	"github.com/4chain-ag/go-wallet-toolbox/pkg/internal/fixtures/testusers"
 	"github.com/4chain-ag/go-wallet-toolbox/pkg/internal/satoshi"
-	"github.com/4chain-ag/go-wallet-toolbox/pkg/storage/internal/actions/funder/testabilities"
+	testabilities2 "github.com/4chain-ag/go-wallet-toolbox/pkg/internal/storage/funder/testabilities"
```
chris-4chain: 🤔

pkg/internal/storage/funder/testabilities/fixture_utxo.go
```diff
@@ -5,8 +5,8 @@ import (
 	"time"
 
 	"github.com/4chain-ag/go-wallet-toolbox/pkg/internal/fixtures/testusers"
+	models2 "github.com/4chain-ag/go-wallet-toolbox/pkg/internal/storage/database/models"
```
chris-4chain: 🤔

pkg/internal/storage/repo/migrator.go
```diff
@@ -4,7 +4,7 @@ import (
 	"context"
 	"fmt"
 
-	"github.com/4chain-ag/go-wallet-toolbox/pkg/storage/internal/database/models"
+	models2 "github.com/4chain-ag/go-wallet-toolbox/pkg/internal/storage/database/models"
```
chris-4chain: 🤔

pkg/internal/storage/repo/transactions.go
```diff
@@ -5,10 +5,10 @@ import (
 	"errors"
 	"fmt"
 
+	models2 "github.com/4chain-ag/go-wallet-toolbox/pkg/internal/storage/database/models"
```
chris-4chain: 🤔

pkg/internal/storage/funder/result.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"fmt"
+
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/internal/satoshi"
+	"github.com/go-softwarelab/common/pkg/seq"
+)
+
+type Result struct {
+	AllocatedUTXOs []*UTXO
+	ChangeCount    uint64
```
ac4ch: `ChangeCount` could be ambiguous. Is it the number of outputs or number of UTXOs changed? Consider `ChangeOutputCount` if that’s the intent.

pkg/internal/storage/funder/result.go
```diff
@@ -0,0 +1,31 @@
...
+	ChangeAmount   satoshi.Value
+	Fee            satoshi.Value
+}
+
+func (fr *Result) TotalAllocated() (satoshi.Value, error) {
+	total, err := satoshi.Sum(seq.Map(seq.FromSlice(fr.AllocatedUTXOs), func(utxo *UTXO) satoshi.Value {
+		return utxo.Satoshis
+	}))
```
ac4ch:
```suggestion
        if fr.AllocatedUTXOs == nil {
		return 0, nil // Explicitly handle nil slice gracefully
	}
	total, err := satoshi.Sum(seq.Map(seq.FromSlice(fr.AllocatedUTXOs), func(utxo *UTXO) satoshi.Value {
		return utxo.Satoshis
	}))
```
if `fr.AllocatedUTXOs` is nil, `seq.FromSlice` will still work (returns empty slice), but it might be good to explicitly check and return early ? WDYT ?

pkg/internal/storage/funder/sql_test.go
ac4ch: think been left over from refactoring .... good catch :)

pkg/internal/testabilities/fixture_storage.go
```diff
@@ -0,0 +1,111 @@
...
+	return &faucetFixture{
+		t:        s.t,
+		user:     user,
+		db:       s.db,
+		basketID: basket.BasketID,
+	}
+}
+
+func Given(t testing.TB) StorageFixture {
+	db, _ := dbfixtures.TestDatabase(t)
```
ac4ch:
```suggestion
	db, err := dbfixtures.TestDatabase(t)
	require.NoError(t, err)
```

pkg/internal/testabilities/fixture_storage.go
```diff
@@ -0,0 +1,111 @@
...
+		network:    defs.NetworkTestnet,
+		commission: defs.Commission{},
+		feeModel:   defs.DefaultFeeModel(),
+		randomizer: randomizer.New(),
+	}
+}
+
+func (s *storageFixture) Faucet(activeStorage *storage.Provider, user testusers.User) FaucetFixture {
+	s.t.Helper()
+	ctx := context.Background()
```
ac4ch:
```suggestion
	ctx := s.t.Context()
```
would not be better to use `ctx := s.t.Context()` ?

pkg/internal/storage/funder/result.go
dorzepowski: seq.FromSlice is not returning slice, 
It is iterator, so it's simply iterating over the fr.AllocatedUTXOs
and it's working the same as:
```go
for _, v := fr.AllocatedUTXOs {
```
and if it is nil, then for loop won't start
and because there is no need to explicitly check the len of the slice before for loop, I don't think that this is needed in this case.

pkg/internal/testabilities/fixture_storage.go
dorzepowski: I don't know, I don't have opinion about that,
but I will change it.

pkg/internal/testabilities/fixture_storage.go
dorzepowski: it is not error, but cleanup function,
but I decide to return this function and use in tests

pkg/internal/storage/funder/sql_test.go
dorzepowski: Thanks, good catch

jakubmkowalski (approved):

chris-4chain (approved):

ac4ch (approved):

PR #243, opened 2025-05-27
feat(#242): initialise component wallet storage manager

dorzepowski:
Preparing a skeleton of the WalletStorageManager which is needed by a storage.
Because it is handling AuthID internally, this is an essential component for Wallet to work.
I'm focusing only on basic and essential functionality needed by a Wallet, so for example there is no handling for backup storages, or conflicting active storages, etc.


pkg/storage/storage_manager.go
```diff
@@ -0,0 +1,122 @@
...
+	"fmt"
+
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/wdk"
+	"github.com/go-softwarelab/common/pkg/is"
+	"github.com/go-softwarelab/common/pkg/to"
+)
+
+var _ wdk.WalletStorage = (*WalletStorageManager)(nil)
+
+type managedStorage struct {
```
chris-4chain: Consider to move the `managedStorage` along with its methods to a new file.

pkg/wdk/storage.interface.go
```diff
@@ -5,19 +5,50 @@ import (
...
+	// ProcessAction processes a signed transaction created by CreateAction.
+	// @Write
 	ProcessAction(ctx context.Context, auth AuthID, args ProcessActionArgs) (*ProcessActionResult, error)
 
+	// InsertCertificateAuth adds a new certificate for a user.
+	// @Write
 	InsertCertificateAuth(ctx context.Context, auth AuthID, certificate *TableCertificateX) (uint, error)
+
+	// RelinquishCertificate revokes the specified certificate from the users certificates.
+	// @Write
```
chris-4chain, line 44: Nice :) 
Consider adding some documentation what annotations are supported.

chris-4chain (approved):

pkg/wdk/storage.interface.go
dorzepowski: Any anotations are supported.
This is a responsibility of the template to handle them correctly
I could add some documentation, but I don't know where 🤷

dorzepowski: Added README to client-gen

pkg/storage/storage_manager.go
dorzepowski: I moved it to internal package (I'm not sure if this was what you want)

chris-4chain (approved):

pkg/storage/internal/managed/managed_storage.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/wdk"
+)
+
+type Storage struct {
+	wdk.WalletStorageProvider
+	Settings *wdk.TableSettings
+	User     *wdk.TableUser
+}
+
+func NewMangedStorage(storage wdk.WalletStorageProvider) *Storage {
```
ac4ch:
```suggestion
func NewManagedStorage(storage wdk.WalletStorageProvider) *Storage {
```
typo ?

tools/client-gen/metadata/methods.go
ac4ch, line 1: i would think about adding unit tests for func's like `ReturnError`, `ArgumentOfType`, `SkipTypes`, and `HasError`

tools/client-gen/metadata/methods.go
```diff
@@ -0,0 +1,99 @@
...
+type MethodInfo struct {
+	Name        string    `json:"name"`
+	Comments    []string  `json:"comment"`
+	Annotations []string  `json:"annotations"`
+	Arguments   Arguments `json:"arguments"`
+	Results     Results   `json:"results"`
+}
+
+// HasAnnotation checks if the method contains an annotation that starts with the specified annotationType prefix.
+// By annotation, we mean a method Docs that starts with "// @".
```
ac4ch:
```suggestion
// HasAnnotation returns true if the method contains an annotation comment that starts with the specified prefix.
// Annotations are method-level comments starting with "// @", such as "// @Write".
```

tools/client-gen/metadata/package_info.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+func (pkg *Package) printForSamePackage(typeName string) string {
+	return typeName
+}
+
+func (pkg *Package) printForExternal(typeName string) string {
+	if typeName == "" {
+		return typeName
+	}
```
ac4ch, line 55:
```suggestion
	if len(typeName) == 0 {
	return ""
}
```
since later u do `typeName[0] == '*'` even this early return guarantee u will never get to `typeName[0] == '*'` checking here `len` instead of  empty string `""` it's more readable WDYT?

tools/client-gen/metadata/package_info.go
ac4ch, line 1: WDYT about adding unit tests here too ?

tools/client-gen/metadata/methods.go
```diff
@@ -0,0 +1,99 @@
...
+			return "0"
+		case isPointerType(it.Type):
+			return "nil"
+		default:
+			return pkg.PrintType(it.Type) + "{}"
+		}
+	})
+
+	return "return " + strings.Join(returns, ", ")
+}
```
ac4ch, line 91:
```suggestion
func (r Results) ReturnError(pkg Package, errVarName string) string {
	returns := make([]string, len(r))

	for i, it := range r {
		returns[i] = zeroValueForType(it.Type, pkg, errVarName)
	}

	return "return " + strings.Join(returns, ", ")
}

func zeroValueForType(typ string, pkg Package, errVarName string) string {
	switch typ {
	case "error":
		return errVarName
	case "string":
		return `""`
	case "bool":
		return "false"
	}

	if isNumberType(typ) {
		return "0"
	}

	if isPointerType(typ) {
		return "nil"
	}

	return pkg.PrintType(typ) + "{}"
}
```
each case is clear and not deeply nested, not sure if u would need to handle other types here too, but this way i think is easier to add new types if needed, since i am not sure if we really need `slicesx` here ... at lease this way we have import less ;)

dorzepowski: it is a matter of taste, and I don't see a big benefits from this change.

tools/client-gen/metadata/package_info.go
dorzepowski: From my perspective comparing to empty string is more readable and concise

ac4ch (approved):

PR #249, opened 2025-05-28
feat(#183): implement wallet InternalizeAction

dorzepowski:


mgosek-4chain (approved):

jakubmkowalski (approved):

pkg/internal/validate/validate_originator.go
```diff
@@ -0,0 +1,36 @@
...
+		return fmt.Errorf("originator must be between 1 and 250 bytes, but was: %d", len(originator))
+	}
+
+	if !strings.Contains(originator, ".") {
+		return nil
+	}
+
+	for originatorPart := range strings.SplitSeq(originator, ".") {
+		if !is.Between(len(originatorPart), 1, 63) {
+			return fmt.Errorf("originator part must be between 1 and 63 bytes, but was: %d", len(originatorPart))
```
ac4ch: minor but i think looks better  
```suggestion
			return fmt.Errorf("originator part %q must be between 1 and 63 bytes, but was: %d", originatorPart, len(originatorPart))

```

pkg/internal/validate/validate_originator_test.go
```diff
@@ -0,0 +1,51 @@
...
+			originator: "part1.part2.part3",
+		},
+	}
+	for name, test := range successTestCases {
+		t.Run(name, func(t *testing.T) {
+			err := validate.Originator(test.originator)
+			require.NoError(t, err)
+		})
+	}
+}
```
ac4ch, line 51:
```suggestion
func TestValidateOriginator(t *testing.T) {
	t.Run("should return error for invalid originators", func(t *testing.T) {
		tests := map[string]struct {
			originator string
		}{
			"exceeds total length limit": {
				originator: strings.Repeat("a", 251),
			},
			"part exceeds max part length": {
				originator: "part1." + strings.Repeat("a", 64) + ".part3",
			},
			"contains empty part": {
				originator: "part1..part3",
			},
		}

		for name, tc := range tests {
			t.Run(name, func(t *testing.T) {
				// Given
				originator := tc.originator

				// When
				err := validate.Originator(originator)

				// Then
				require.Error(t, err)
			})
		}
	})

	t.Run("should pass for valid originators", func(t *testing.T) {
		tests := map[string]struct {
			originator string
		}{
			"valid short originator": {
				originator: "short",
			},
			"valid max length originator": {
				originator: strings.Repeat("a", 250),
			},
			"valid originator with multiple parts": {
				originator: "part1.part2.part3",
			},
		}

		for name, tc := range tests {
			t.Run(name, func(t *testing.T) {
				// Given
				originator := tc.originator

				// When
				err := validate.Originator(originator)

				// Then
				require.NoError(t, err)
			})
		}
	})
}
```
WDYT ? think looks better and it has GWT format ?

pkg/wallet/wallet_internalize_action_test.go
```diff
@@ -0,0 +1,147 @@
...
+		})
+	}
+}
+
+func (s *WalletTestSuite) TestWalletInternalizeAction() {
+	s.Run("wallet payment protocol", func() {
+		t := s.T()
+
+		// given:
+		given, _ := testabilities.New(t)
```
chris-4chain:
```suggestion
		given := testabilities.Given(t)
```

pkg/wallet/wallet_internalize_action_test.go
```diff
@@ -0,0 +1,147 @@
...
+		assert.NoError(t, err)
+		require.NotNil(t, result)
+		assert.True(t, result.Accepted)
+	})
+
+	s.Run("basket insertion protocol", func() {
+		t := s.T()
+
+		// given:
+		given, _ := testabilities.New(t)
```
chris-4chain:
```suggestion
		given := testabilities.Given(t)
```

chris-4chain (approved):

pkg/wallet/wallet_internalize_action_test.go
```diff
@@ -0,0 +1,147 @@
...
+		},
+	}
+	for name, test := range errorTestCases {
+		t.Run(name, func(t *testing.T) {
+			// given:
+			given, then := testabilities.New(t)
+
+			aliceWallet, cleanup := given.AliceWalletWithStorage(testabilities.StorageTypeMocked)
+			defer cleanup()
+
```
ac4ch, line 94:
```go
// when:
```

pkg/wallet/wallet_internalize_action_test.go
```diff
@@ -0,0 +1,147 @@
...
+			action, err := aliceWallet.InternalizeAction(t.Context(), test.args(), test.originator)
+
+			then.Result(action).HasError(err)
```
ac4ch, line 99:
```suggestion
                         // then:
			then.Result(action).HasError(err)
```

pkg/wallet/wallet_internalize_action_test.go
```diff
@@ -0,0 +1,147 @@
...
+		// and:
+		args := fixtures.DefaultWalletInternalizeActionArgs(t, sdk.InternalizeProtocolBasketInsertion)
+
+		// when:
+		result, err := aliceWallet.InternalizeAction(t.Context(), args, fixtures.DefaultOriginator)
+
+		// then:
+		assert.NoError(t, err)
+		require.NotNil(t, result)
+		assert.True(t, result.Accepted)
```
ac4ch: think this can be `require` it is last condition anyway ?

ac4ch (approved):

pkg/internal/validate/validate_originator_test.go
dorzepowski: Good points, BUT:

1. I would like to group same as you did, but there are limitations in Jetbrain IDE that won't allow for running single tests from such a grouping, while in the current version, I can run single test from IDE with easy - that's why I prefer convenience of usage over the small improvment in readability
2. When it comes to GWT format - in such a tests I feel like it's rather a case of style over substance. I was rather thinking about changing it to one-liner for test body (if we really want to change it)
```go
t.Run(...,func(t *testing.T) {
   require.NoError(t, validate.Originator(test.originator))
})
```

WDYT about this one-liner approach?

pkg/internal/validate/validate_originator.go
dorzepowski: Nice

pkg/wallet/wallet_internalize_action_test.go
dorzepowski: Thanks, good point.

pkg/wallet/wallet_internalize_action_test.go
dorzepowski: It doesn't matter - and it's making it easier for adding another field, because no change in this line will be required.

chris-4chain (approved):

PR #253, opened 2025-05-28
fix: hack(BOGM) - bind bumps to transactions in BEEF

dorzepowski:


ac4ch (approved):

pkg/services/internal/arc/arc_service.go
```diff
@@ -261,6 +264,27 @@ func toHex(beef *transaction.Beef) (string, error) {
 	return beefHex, nil
 }
 
+func (s *Service) bindBumpsAndTransactions(beef *transaction.Beef) {
+	for i, bump := range beef.BUMPs {
+		if bump.Path == nil || len(bump.Path) == 0 || len(bump.Path[0]) == 0 {
```
chris-4chain: > S1009: should omit nil check; len() for nil slices is defined as zero (gosimple)
```suggestion
		if len(bump.Path) == 0 || len(bump.Path[0]) == 0 {
```

chris-4chain (approved):

pkg/services/internal/arc/arc_service.go
dorzepowski: good point

PR #257, opened 2025-05-30
[FEATURE] listActions on sql storage

### LINKED ISSUE #23: [FEATURE] listActions on sql storage
## Desired Solution

## Suggested Implementation

## Alternatives Considered

## Additional Context

## Categorize Feature

ac4ch:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter

### DIFF
--- a/pkg/internal/storage/database/models/tx_labels.go
+++ b/pkg/internal/storage/database/models/tx_labels.go
@@ -1,0 +1,7 @@
+package models
+
+type TransactionLabels struct {
+	TransactionID uint   `gorm:"column:transaction_id"`
+	LabelName     string `gorm:"column:label_name"`
+	LabelUserID   int    `gorm:"column:label_user_id"`
+}
--- a/pkg/internal/storage/repo/outputs.go
+++ b/pkg/internal/storage/repo/outputs.go
@@ -210,3 +210,33 @@ func (o *Outputs) FindOutput(ctx context.Context, userID int, outpoint wdk.OutPo
 	tableOutput.TxID = &outpoint.TxID
 	return tableOutput, nil
 }
+
+// FindInputsAndOutputsWithBuckets retrieves inputs and outputs for given transaction IDs, including basket information.
+// It returns two maps: one for inputs keyed by SpentBy ID and another for outputs keyed by TransactionID.
+// Each map contains slices of TableOutput, which include basket details if available.
+func (o *Outputs) FindInputsAndOutputsWithBuckets(ctx context.Context, txIDs []uint) (inputs map[uint][]*wdk.TableOutput, outputs map[uint][]*wdk.TableOutput, err error) {
+	if len(txIDs) == 0 {
+		return
+	}
+
+	var allOutputs []*models.Output
+	if err := o.db.WithContext(ctx).
+		Preload("Basket").
+		Where("transaction_id IN ? OR spent_by IN ?", txIDs, txIDs).
+		Find(&allOutputs).Error; err != nil {
+		return nil, nil, fmt.Errorf("failed to fetch inputs/outputs: %w", err)
+	}
+
+	inputMap := make(map[uint][]*wdk.TableOutput)
+	outputMap := make(map[uint][]*wdk.TableOutput)
+
+	for _, out := range allOutputs {
+		tableOut := o.mapModelToTableOutput(out)
+		if out.SpentBy != nil {
+			inputMap[*out.SpentBy] = append(inputMap[*out.SpentBy], tableOut)
+		}
+		outputMap[out.TransactionID] = append(outputMap[out.TransactionID], tableOut)
+	}
+
+	return inputMap, outputMap, nil
+}
--- a/pkg/internal/storage/repo/proven_tx_req.go
+++ b/pkg/internal/storage/repo/proven_tx_req.go
@@ -89,6 +89,32 @@ func (p *ProvenTxReq) FindProvenTxRawTX(ctx context.Context, txID string) ([]byt
 	return model.RawTx, nil
 }
 
+func (p *ProvenTxReq) FindProvenTxRawTXs(ctx context.Context, txIDs []string) (map[string][]byte, error) {
+	if len(txIDs) == 0 {
+		return make(map[string][]byte), nil
+	}
+
+	var results []struct {
+		TxID  string
+		RawTx []byte
+	}
+
+	err := p.db.WithContext(ctx).
+		Model(&models.ProvenTxReq{}).
+		Select("tx_id, raw_tx").
+		Where("tx_id IN ?", txIDs).
+		Scan(&results).Error
+	if err != nil {
+		return nil, fmt.Errorf("failed to batch fetch raw tx: %w", err)
+	}
+
+	rawTxMap := make(map[string][]byte)
+	for _, r := range results {
+		rawTxMap[r.TxID] = r.RawTx
+	}
+	return rawTxMap, nil
+}
+
 func (p *ProvenTxReq) FindProvenTxStatus(ctx context.Context, txID string) (wdk.ProvenTxReqStatus, error) {
 	var model models.ProvenTxReq
 	err := p.db.WithContext(ctx).
--- a/pkg/internal/storage/repo/transactions.go
+++ b/pkg/internal/storage/repo/transactions.go
@@ -343,3 +343,80 @@ func (txs *Transactions) mapModelToTableTransaction(model *models.Transaction) *
 		InputBEEF:     model.InputBeef,
 	}
 }
+
+func (txs *Transactions) ListAndCountActions(ctx context.Context, userID int, filter entity.ListActionsFilter) ([]*wdk.TableTransaction, int64, error) {
+	var actions []*models.Transaction
+	var total int64
+
+	err := txs.db.WithContext(ctx).Transaction(func(tx *gorm.DB) error {
+		query := tx.Model(&models.Transaction{}).
+			Where("user_id = ?", userID)
+
+		if len(filter.Status) > 0 {
+			query = query.Where("status IN ?", filter.Status)
+		}
+
+		if len(filter.Labels) > 0 {
+			labelSubQuery := tx.Model(&models.TransactionLabels{}).
+				Select("transaction_id").
+				Where("label_name IN ?", filter.Labels).
+				Where("label_user_id = ?", userID)
+
+			if filter.IncludeAllLabels {
+				labelSubQuery = labelSubQuery.
+					Group("transaction_id").
+					Having("COUNT(DISTINCT label_name) = ?", len(filter.Labels))
+			}
+
+			query = query.Where("id IN (?)", labelSubQuery)
+		}
+
+		if err := query.Count(&total).Error; err != nil {
+			return fmt.Errorf("count failed: %w", err)
+		}
+
+		if err := query.
+			Limit(filter.Limit).
+			Offset(filter.Offset).
+			Order("id ASC").
+			Find(&actions).Error; err != nil {
+			return fmt.Errorf("query failed: %w", err)
+		}
+
+		return nil
+	})
+
+	if err != nil {
+		return nil, 0, fmt.Errorf("transaction failed: %w", err)
+	}
+
+	return slices.Map(actions, txs.mapModelToTableTransaction), total, nil
+}
+
+func (txs *Transactions) GetLabelsForTransactions(ctx context.Context, txIDs []uint) (map[uint][]string, error) {
+	if len(txIDs) == 0 {
+		return make(map[uint][]string), nil
+	}
+
+	type resultRow struct {
+		TransactionID uint
+		LabelName     string
+	}
+
+	var rows []resultRow
+	err := txs.db.WithContext(ctx).
+		Model(&models.TransactionLabels{}).
+		Select("transaction_id, label_name").
+		Where("transaction_id IN ?", txIDs).
+		Where("label_name IS NOT NULL").
+		Scan(&rows).Error
+	if err != nil {
+		return nil, fmt.Errorf("failed to batch fetch labels: %w", err)
+	}
+
+	labelsMap := make(map[uint][]string)
+	for _, row := range rows {
+		labelsMap[row.TransactionID] = append(labelsMap[row.TransactionID], row.LabelName)
+	}
+	return labelsMap, nil
+}
--- a/pkg/internal/validate/list_actions_args.go
+++ b/pkg/internal/validate/list_actions_args.go
@@ -1,0 +1,46 @@
+package validate
+
+import (
+	"fmt"
+
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/wdk"
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/wdk/primitives"
+)
+
+func ListActionsArgs(args *wdk.ListActionsArgs) error {
+	if args == nil {
+		return fmt.Errorf("args cannot be nil")
+	}
+
+	if args.LabelQueryMode != nil {
+		if err := args.LabelQueryMode.Validate(); err != nil {
+			return fmt.Errorf("invalid labelQueryMode: %s", *args.LabelQueryMode)
+		}
+	}
+
+	if args.Limit > MaxPaginationLimit {
+		return fmt.Errorf("limit must be less than or equal to %d", MaxPaginationLimit)
+	}
+	if args.Offset > MaxPaginationOffset {
+		return fmt.Errorf("offset must be less than or equal to %d", MaxPaginationOffset)
+	}
+
+	for _, label := range args.Labels {
+		if err := validateLabel(label); err != nil {
+			return fmt.Errorf("invalid label: %w", err)
+		}
+	}
+
+	if !args.SeekPermissions.Value() {
+		return fmt.Errorf("operation not allowed without permission (seekPermissions=false)")
+	}
+
+	return nil
+}
+
+func validateLabel(label primitives.StringUnder300) error {
+	if len(label) == 0 || len(label) > 300 {
+		return fmt.Errorf("label must be 1-300 characters")
+	}
+	return nil
+}
--- a/pkg/storage/internal/actions/list_actions.go
+++ b/pkg/storage/internal/actions/list_actions.go
@@ -1,0 +1,73 @@
+package actions
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/internal/logging"
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/wdk"
+	"github.com/go-softwarelab/common/pkg/must"
+)
+
+type listActions struct {
+	logger           *slog.Logger
+	transactionsRepo TransactionsRepo
+	outputsRepo      OutputRepo
+	provenTxRepo     ProvenTxRepo
+	basketRepo       BasketRepo
+}
+
+func newListActions(logger *slog.Logger, transactions TransactionsRepo, outputs OutputRepo, provenTx ProvenTxRepo, basket BasketRepo) *listActions {
+	return &listActions{
+		logger:           logging.Child(logger, "list_actions"),
+		transactionsRepo: transactions,
+		outputsRepo:      outputs,
+		provenTxRepo:     provenTx,
+		basketRepo:       basket,
+	}
+}
+
+func (l *listActions) ListActions(ctx context.Context, auth wdk.AuthID, args *wdk.ListActionsArgs) (*wdk.ListActionsResult, error) {
+	userID := *auth.UserID
+	filter, err := l.toFilterParams(userID, args)
+	if err != nil {
+		return nil, fmt.Errorf("failed to convert filter params: %w", err)
+	}
+
+	txs, total, err := l.transactionsRepo.ListAndCountActions(ctx, userID, filter)
+	if err != nil {
+		return nil, fmt.Errorf("failed to list transactions: %w", err)
+	}
+
+	transactionIDs, txIDs, actions := l.mapTransactionsToActions(txs)
+
+	inputMap, outputMap, err := l.fetchInputsOutputs(ctx, transactionIDs, args)
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch inputs/outputs: %w", err)
+	}
+
+	basketMap, err := l.buildBasketMap(ctx, outputMap)
+	if err != nil {
+		return nil, fmt.Errorf("failed to build basket map: %w", err)
+	}
+
+	labelMap, err := l.loadLabelsIfNeeded(ctx, transactionIDs, args.IncludeLabels)
+	if err != nil {
+		return nil, fmt.Errorf("failed to load labels: %w", err)
+	}
+
+	rawTxMap, err := l.loadRawTxsIfNeeded(ctx, txIDs, args.IncludeInputs)
+	if err != nil {
+		return nil, fmt.Errorf("failed to load raw transactions: %w", err)
+	}
+
+	if err := l.mapInputsOutputsLabels(actions, txs, inputMap, outputMap, labelMap, rawTxMap, basketMap, args); err != nil {
+		return nil, fmt.Errorf("failed to map inputs/outputs/labels: %w", err)
+	}
+
+	return &wdk.ListActionsResult{
+		TotalActions: must.ConvertToUInt32(total),
+		Actions:      actions,
+	}, nil
+}
--- a/pkg/storage/internal/actions/list_actions_mapping.go
+++ b/pkg/storage/internal/actions/list_actions_mapping.go
@@ -1,0 +1,272 @@
+package actions
+
+import (
+	"context"
+	"fmt"
+	"maps"
+	"slices"
+	"strings"
+
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/internal/storage/entity"
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/wdk"
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/wdk/primitives"
+	"github.com/bsv-blockchain/go-sdk/transaction"
+	"github.com/go-softwarelab/common/pkg/must"
+	"github.com/go-softwarelab/common/pkg/optional"
+	"github.com/go-softwarelab/common/pkg/seq"
+	commonslices "github.com/go-softwarelab/common/pkg/slices"
+)
+
+func (l *listActions) boolFromLabelQueryMode(labelQueryMode *primitives.LabelQueryModeString) (bool, error) {
+	if labelQueryMode == nil {
+		return false, nil
+	}
+	switch strings.ToLower(string(*labelQueryMode)) {
+	case "all":
+		return true, nil
+	case "any":
+		return false, nil
+	default:
+		return false, fmt.Errorf("invalid LabelQueryMode: %s", *labelQueryMode)
+	}
+}
+
+func (l *listActions) toFilterParams(userID int, args *wdk.ListActionsArgs) (entity.ListActionsFilter, error) {
+	labelNames := commonslices.Map(args.Labels, func(label primitives.StringUnder300) string {
+		return string(label)
+	})
+
+	statuses := []wdk.TxStatus{
+		wdk.TxStatusCompleted, wdk.TxStatusUnprocessed, wdk.TxStatusSending, wdk.TxStatusUnproven,
+		wdk.TxStatusUnsigned, wdk.TxStatusNoSend, wdk.TxStatusNonFinal,
+	}
+	labelQueryMode, err := l.boolFromLabelQueryMode(args.LabelQueryMode)
+	if err != nil {
+		return entity.ListActionsFilter{}, fmt.Errorf("failed to parse LabelQueryMode: %w", err)
+	}
+
+	return entity.ListActionsFilter{
+		UserID:           userID,
+		Labels:           labelNames,
+		Status:           statuses,
+		IncludeAllLabels: labelQueryMode,
+		Limit:            must.ConvertToIntFromUnsigned(args.Limit),
+		Offset:           must.ConvertToIntFromUnsigned(args.Offset),
+	}, nil
+}
+
+func (l *listActions) mapTransactionsToActions(txs []*wdk.TableTransaction) ([]uint, []string, []wdk.WalletAction) {
+	transactionIDs := make([]uint, len(txs))
+	var txIDs []string
+	actions := make([]wdk.WalletAction, len(txs))
+
+	for i, tx := range txs {
+		transactionIDs[i] = tx.TransactionID
+		if tx.TxID != nil {
+			txIDs = append(txIDs, *tx.TxID)
+		}
+
+		actions[i] = wdk.WalletAction{
+			Satoshis:    tx.Satoshis,
+			Status:      string(tx.Status),
+			IsOutgoing:  tx.IsOutgoing,
+			Description: tx.Description,
+			TxID:        optional.OfPtr(tx.TxID).OrZeroValue(),
+			Version:     optional.OfPtr(tx.Version).OrZeroValue(),
+			LockTime:    optional.OfPtr(tx.LockTime).OrZeroValue(),
+		}
+	}
+
+	return transactionIDs, txIDs, actions
+}
+
+func (l *listActions) fetchInputsOutputs(ctx context.Context, txIDs []uint, args *wdk.ListActionsArgs) (map[uint][]*wdk.TableOutput, map[uint][]*wdk.TableOutput, error) {
+	if args.IncludeInputs.Value() || args.IncludeOutputs.Value() {
+		inputs, outputs, err := l.outputsRepo.FindInputsAndOutputsWithBuckets(ctx, txIDs)
+		if err != nil {
+			return nil, nil, fmt.Errorf("failed to fetch inputs/outputs: %w", err)
+		}
+		return inputs, outputs, nil
+	}
+	return map[uint][]*wdk.TableOutput{}, map[uint][]*wdk.TableOutput{}, nil
+}
+
+func (l *listActions) buildBasketMap(ctx context.Context, outputMap map[uint][]*wdk.TableOutput) (map[int]primitives.StringUnder300, error) {
+	basketIDSet := map[int]struct{}{}
+	for _, outputs := range outputMap {
+		for _, out := range outputs {
+			if out.BasketID != nil {
+				basketIDSet[*out.BasketID] = struct{}{}
+			}
+		}
+	}
+
+	basketIDs := seq.ToSlice(maps.Keys(basketIDSet), make([]int, 0, len(basketIDSet)))
+
+	if len(basketIDs) == 0 {
+		return map[int]primitives.StringUnder300{}, nil
+	}
+
+	baskets, err := l.basketRepo.FindBasketsByIDs(ctx, basketIDs)
+	if err != nil {
+		return nil, fmt.Errorf("failed to load baskets: %w", err)
+	}
+
+	basketMap := make(map[int]primitives.StringUnder300, len(baskets))
+	for _, b := range baskets {
+		basketMap[b.BasketID] = b.Name
+	}
+
+	return basketMap, nil
+}
+
+func (l *listActions) loadLabelsIfNeeded(ctx context.Context, txIDs []uint, include *primitives.BooleanDefaultFalse) (map[uint][]string, error) {
+	if !include.Value() {
+		return map[uint][]string{}, nil
+	}
+
+	labelMap, err := l.transactionsRepo.GetLabelsForTransactions(ctx, txIDs)
+	if err != nil {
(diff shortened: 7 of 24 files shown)


pkg/storage/internal/actions/list_actions.go
```diff
@@ -0,0 +1,217 @@
...
+		if tx.LockTime != nil {
+			actions[i].LockTime = *tx.LockTime
+		}
+	}
+
+	basketIDSet := map[int]struct{}{}
+	allOutputs := map[uint][]*wdk.TableOutput{}
+
+	for _, tx := range txs {
+		_, outputs, err := l.outputsRepo.FindInputsAndOutputsWithBuckets(ctx, tx.TransactionID)
```
chris-4chain: **Possible Performance Issue - Multiple Queries in a Loop:**
Maybe, it's ok, and there is no other option to do it - but please re-check if more actions couldn't be done on SQL and make it more performant.

pkg/storage/internal/actions/list_actions.go
```diff
@@ -0,0 +1,217 @@
...
+			return nil, fmt.Errorf("failed to load outputs for tx %d: %w", tx.TransactionID, err)
+		}
+		allOutputs[tx.TransactionID] = outputs
+
+		for _, out := range outputs {
+			basketIDSet[out.GetBasketID()] = struct{}{}
+		}
+	}
+
+	var basketIDs []int
```
chris-4chain: The length is known so preallocate the slice:
```suggestion
basketIDs := make([]int, 0, len(basketIDSet))
```

pkg/wdk/table_output.go
```diff
@@ -25,3 +25,21 @@ type TableOutput struct {
...
+	}
+	return *o.TxID
+}
+
+// GetBasketID returns the basket ID of the output.
+// If BasketID is nil, it returns 0.
+func (o TableOutput) GetBasketID() int {
+	if o.BasketID == nil {
+		return 0
+	}
```
chris-4chain: I wouldn't do it this way - because `BasketID = 0` could ba a valid Basket (depends on DB engine)

pkg/wdk/table_output.go
```diff
@@ -25,3 +25,21 @@ type TableOutput struct {
...
 }
+
+// GetTxID returns the transaction ID of the output.
+// If TxID is nil, it returns an empty string.
+func (o TableOutput) GetTxID() string {
+	if o.TxID == nil {
+		return ""
+	}
+	return *o.TxID
+}
```
chris-4chain: I don't like this function.
Please re-check if it's really needed.

pkg/storage/internal/actions/list_actions.go
```diff
@@ -0,0 +1,217 @@
...
+		var err error
+		tx, err = transaction.NewTransactionFromBytes(rawTx)
+		if err != nil {
+			return result
+		}
+	}
+
+	for _, o := range inputs {
+		input := wdk.WalletActionInput{
+			SourceOutpoint:   fmt.Sprintf("%s.%d", o.GetTxID(), o.Vout),
```
chris-4chain: Are you sure, that - at this point - all `o` inputs have "not-nil" TxIDs?

pkg/storage/internal/actions/list_actions.go
```diff
@@ -0,0 +1,217 @@
...
+	for _, tx := range txs {
+		_, outputs, err := l.outputsRepo.FindInputsAndOutputsWithBuckets(ctx, tx.TransactionID)
+		if err != nil {
+			return nil, fmt.Errorf("failed to load outputs for tx %d: %w", tx.TransactionID, err)
+		}
+		allOutputs[tx.TransactionID] = outputs
+
+		for _, out := range outputs {
+			basketIDSet[out.GetBasketID()] = struct{}{}
+		}
```
chris-4chain: To avoid `GetBasketID` (which I recommend to remove)
- You could add to the "set" only if the `BasketID` is not nil

```suggestion
		for _, out := range outputs {
			if out.BasketID == nil {
				continue
			}
			basketIDSet[*out.BasketID] = struct{}{}
		}
```

And then, the `if id == 0` wouldn't be necessary and the preallocating would fit perfectly.

```golang
	basketIDs := make([]int, 0, len(basketIDSet))
	for id := range basketIDSet {
		basketIDs = append(basketIDs, id)
	}
```

pkg/storage/internal/actions/list_actions.go
```diff
@@ -0,0 +1,217 @@
...
+	}
+
+	var basketIDs []int
+	for id := range basketIDSet {
+		if id != 0 {
+			basketIDs = append(basketIDs, id)
+		}
+	}
```
chris-4chain, line 58: Just FYI, how it could be done differently :) 

```golang
basketIDs := seq.ToSlice(maps.Keys(basketIDSet), make([]int, 0, len(basketIDSet)))
```

pkg/storage/internal/actions/list_actions.go
```diff
@@ -0,0 +1,217 @@
...
+		provenTxRepo:     provenTx,
+	}
+}
+
+func (l *listActions) ListActions(ctx context.Context, auth wdk.AuthID, args *wdk.ListActionsArgs) (*wdk.ListActionsResult, error) {
+	filter := l.toFilterParams(*auth.UserID, args)
+
+	if l.transactionsRepo == nil {
+		return nil, fmt.Errorf("transactions repository is not initialized")
+	}
```
chris-4chain: Hmm, this `if` is not-bad,
however, since it is out internal struct/object we are responsible for this that `transactionRepo` is properly set.
- BTW, we don't do such checks for other `actions` and I don't see a reason to make an exception here.

pkg/storage/internal/actions/list_actions.go
```diff
@@ -0,0 +1,217 @@
...
+			IsOutgoing:  tx.IsOutgoing,
+			Description: tx.Description,
+		}
+		if tx.Version != nil {
+			actions[i].Version = *tx.Version
+		}
+		if tx.LockTime != nil {
+			actions[i].LockTime = *tx.LockTime
+		}
+	}
```
chris-4chain, line 41: Maybe, instead of this `GetTxID`
```golang
		if tx.TxID != nil {
			actions[i].TxID = *tx.TxID
		}
```

pkg/storage/internal/actions/list_actions.go
```diff
@@ -0,0 +1,217 @@
...
+	basketMap := map[int]string{}
+	for _, b := range baskets {
+		basketMap[b.BasketID] = b.Name
+	}
+
+	for i, tx := range txs {
+		action := &actions[i]
+
+		if args.IncludeLabels {
+			labelNames, err := l.transactionsRepo.GetLabelsForTransaction(ctx, tx.TransactionID)
```
chris-4chain: Couldn't labels be loaded along with other "data" during `ListAndCountActions` ? This would limit the number of DB queries done in a loop.

pkg/storage/internal/actions/list_actions.go
```diff
@@ -0,0 +1,217 @@
...
+			if err != nil {
+				return nil, fmt.Errorf("failed to load labels: %w", err)
+			}
+			action.Labels = make([]string, len(labelNames))
+			for j, name := range labelNames {
+				action.Labels[j] = string(primitives.StringUnder300(name))
```
chris-4chain: Unnecessary casting, just:
```suggestion
				action.Labels[j] = name
```

pkg/storage/internal/actions/list_actions.go
```diff
@@ -0,0 +1,217 @@
...
+			}
+		}
+
+		inputs, outputs, err := l.outputsRepo.FindInputsAndOutputsWithBuckets(ctx, tx.TransactionID)
```
chris-4chain: Hmmm. Above you already call `FindInputsAndOutputsWithBuckets`.
Please revisit this "Querying db twice for same ids" situation

pkg/storage/internal/actions/list_actions.go
```diff
@@ -0,0 +1,217 @@
...
+			action.Outputs = mapToWalletActionOutputs(outputs, basketMap)
+		}
+
+		if args.IncludeInputs {
+			rawTx, err := l.provenTxRepo.FindProvenTxRawTX(ctx, tx.GetTxID())
+			if err != nil {
+				return nil, fmt.Errorf("failed to find raw transaction: %w", err)
+			}
+			action.Inputs = mapToWalletActionInputs(inputs, rawTx, args.IncludeInputSourceLockingScripts, args.IncludeInputUnlockingScripts)
+		}
```
chris-4chain: It seems that if `args.IncludeOutputs` and `args.IncludeInputs` are both `false` - the DB query `FindInputsAndOutputsWithBuckets` can be skipped

pkg/storage/internal/actions/list_actions.go
```diff
@@ -0,0 +1,217 @@
...
+
+	statuses := []wdk.TxStatus{
+		wdk.TxStatusCompleted, wdk.TxStatusUnprocessed, wdk.TxStatusSending, wdk.TxStatusUnproven,
+		wdk.TxStatusUnsigned, wdk.TxStatusNoSend, wdk.TxStatusNonFinal,
+	}
+
+	return entity.ListActionsFilter{
+		UserID:         userID,
+		Labels:         labelNames,
+		Status:         slices.Map(statuses, func(s wdk.TxStatus) string { return s.String() }),
```
chris-4chain: Consider switching the `Status` field for `ListActionsFilter`:
```golang
	Status         []wdk.TxStatus
```

This way, you could skip the unnecessary mapping.

pkg/storage/internal/actions/list_actions.go
```diff
@@ -0,0 +1,217 @@
...
+
+func mapToWalletActionInputs(inputs []*wdk.TableOutput, rawTx []byte, includeSourceLockingScripts, includeUnlockingScripts bool) []wdk.WalletActionInput {
+	result := make([]wdk.WalletActionInput, 0, len(inputs))
+
+	var tx *transaction.Transaction
+	if len(rawTx) > 0 {
+		var err error
+		tx, err = transaction.NewTransactionFromBytes(rawTx)
+		if err != nil {
+			return result
```
chris-4chain: Consider returning here an error. 
BTW, this logic looks weired to me.

pkg/storage/internal/actions/list_actions.go
```diff
@@ -0,0 +1,217 @@
...
+			SequenceNumber:   0,
+		}
+
+		if includeSourceLockingScripts && o.LockingScript != nil {
+			input.SourceLockingScript = *o.LockingScript
+		}
+
+		if tx != nil && includeUnlockingScripts {
+			for _, txIn := range tx.Inputs {
+				if txIn.SourceTXID.String() == *o.TxID && txIn.SourceTxOutIndex == o.Vout {
```
chris-4chain: FYI: the `txIn.SourceTXID.String()` is not "zero-cost" and doing it inside of twice-wrapped-loops (`for { for { `) could make performance issues.

pkg/internal/storage/database/models/tx_labels.go
```diff
@@ -0,0 +1,11 @@
+package models
+
+type TxLabel struct {
+	TransactionID uint   `gorm:"column:transaction_id"`
+	LabelName     string `gorm:"column:label_name"`
+	LabelUserID   int    `gorm:"column:label_user_id"`
+}
+
+func (TxLabel) TableName() string {
+	return "bsv_transaction_labels"
```
chris-4chain: Table names should be generated automatically - along with proper prefix. 
We should avoid defining table names manually.

pkg/internal/storage/database/models/output.go
```diff
@@ -37,3 +37,10 @@ type Output struct {
 
 	UserUTXO *UserUTXO `gorm:"foreignKey:OutputID"`
 }
+
+func (o *Output) GetSpentBy() uint {
+	if o.SpentBy == nil {
+		return 0
+	}
+	return *o.SpentBy
```
chris-4chain: Same comment as for `GetBucketID`

pkg/internal/storage/repo/outputs.go
```diff
@@ -135,3 +135,47 @@ func (o *Outputs) ListAndCountOutputs(ctx context.Context, filter entity.ListOut
...
+		Find(&allOutputs).Error; err != nil {
+		return nil, nil, fmt.Errorf("failed to find inputs and outputs for transactionID %d: %w", transactionID, err)
+	}
+
+	var inputRows, outputRows []*models.Output
+	for _, o := range allOutputs {
+		if o.GetSpentBy() == transactionID {
+			inputRows = append(inputRows, o)
+		}
+		if o.TransactionID == transactionID {
```
chris-4chain: Shouldn't be `else if` ?

pkg/internal/storage/repo/outputs.go
```diff
@@ -135,3 +135,47 @@ func (o *Outputs) ListAndCountOutputs(ctx context.Context, filter entity.ListOut
...
+			outputRows = append(outputRows, o)
+		}
+	}
+
+	inputs = slices.Map(inputRows, o.mapModelToTableOutput)
+	outputs = slices.Map(outputRows, o.mapModelToTableOutput)
+	return
+}
+
+func (o *Outputs) FindBasketsByIDs(ctx context.Context, basketIDs []int) ([]models.OutputBasket, error) {
```
chris-4chain: Repo should't return models itself - mapping to `wdk.Table*` or (if used only internally) `entity`.

Explanation:
Models are GORM related and repositories should be DB agnostic.

pkg/internal/storage/repo/transactions.go
```diff
@@ -334,3 +334,65 @@ func (txs *Transactions) mapModelToTableTransaction(model *models.Transaction) *
 		InputBEEF:     model.InputBeef,
 	}
 }
+
+func (t *Transactions) ListAndCountActions(ctx context.Context, userID int, filter entity.ListActionsFilter) ([]*wdk.TableTransaction, int64, error) {
```
chris-4chain: Please match the receiver name to what was already (`txs`)

```golang
func (txs *Transactions) 
```

pkg/internal/storage/repo/transactions.go
```diff
@@ -334,3 +334,65 @@ func (txs *Transactions) mapModelToTableTransaction(model *models.Transaction) *
...
+	})
+
+	if err != nil {
+		return nil, 0, fmt.Errorf("transaction failed: %w", err)
+	}
+
+	return slices.Map(actions, t.mapModelToTableTransaction), total, nil
+}
+
+func (t *Transactions) GetLabelsForTransaction(ctx context.Context, transactionID uint) ([]string, error) {
```
chris-4chain: Please match the receiver name to what was already (`txs`)

```golang
func (txs *Transactions) 
```

chris-4chain (changes requested):

pkg/wdk/table_output.go
ac4ch: i would say is needed otherwise i will need to keep checking this in the other places .. as for not all transactions we have `TxID`,  what is wrong with this method ?

pkg/storage/internal/actions/list_actions.go
ac4ch: if they dont we do the check using `GetTxID` unless i should also check here before putting SourceOutpoint ? 
i am not actually sure how to even proceed when u have this nil ? if empty string is not valid

pkg/internal/storage/repo/outputs.go
ac4ch: no this is correct, output can potentially belong to both categories (e.g., a change output that is already spent in the same tx). Using else if would prevent capturing both roles, tested this manually updating spentby in DB

pkg/internal/storage/repo/transactions.go
ac4ch: ya i saw this later but somehow linter did not caught this ...

pkg/storage/internal/actions/list_actions.go
nonus25: this was the reason for me to create GetTxID but i think i know why do not like it.. due to returning "" and in ur example we skiping this property if nil

pkg/internal/storage/repo/outputs.go
```diff
@@ -135,3 +135,62 @@ func (o *Outputs) ListAndCountOutputs(ctx context.Context, filter entity.ListOut
...
+	}
+
+	inputMap := make(map[uint][]*wdk.TableOutput)
+	outputMap := make(map[uint][]*wdk.TableOutput)
+
+	for _, out := range allOutputs {
+		tableOut := o.mapModelToTableOutput(out)
+		if out.GetSpentBy() != 0 {
+			inputMap[out.GetSpentBy()] = append(inputMap[out.GetSpentBy()], tableOut)
+		}
```
chris-4chain: This way we could skip the `GetSpentBy` method, which sill I don't like. 
```suggestion
		if out.SpentBy != nil {
			inputMap[*out.SpentBy] = append(inputMap[*out.SpentBy], tableOut)
		}
```

Why I don't like the `GetSpentBy` metod? Because:
- `0` can ba valid ID in database (depends on DB settings of autoincrement)

pkg/internal/storage/repo/outputs.go
```diff
@@ -135,3 +135,62 @@ func (o *Outputs) ListAndCountOutputs(ctx context.Context, filter entity.ListOut
 
 	return slices.Map(outputs, o.mapModelToTableOutput), total, nil
 }
+
+func (o *Outputs) FindInputsAndOutputsWithBuckets(ctx context.Context, txIDs []uint) (inputs map[uint][]*wdk.TableOutput, outputs map[uint][]*wdk.TableOutput, err error) {
```
chris-4chain, line 217: Normally, the documentation comment is not necessary, but here it would be good to describe what are the `inputs`/`outputs` returned maps - especially what `keys` mean.

pkg/internal/storage/repo/outputs.go
```diff
@@ -135,3 +135,62 @@ func (o *Outputs) ListAndCountOutputs(ctx context.Context, filter entity.ListOut
...
+			BasketConfiguration: wdk.BasketConfiguration{
+				Name:                    m.Name,
+				NumberOfDesiredUTXOs:    m.NumberOfDesiredUTXOs,
+				MinimumDesiredUTXOValue: m.MinimumDesiredUTXOValue,
+			},
+		})
+	}
+
+	return baskets, nil
+}
```
chris-4chain:
```suggestion
var dbBaskets []*models.OutputBasket
	err := o.db.WithContext(ctx).
		Model(&models.OutputBasket{}).
		Where("basket_id IN ?", basketIDs).
		Find(&dbBaskets).Error
	if err != nil {
		return nil, fmt.Errorf("failed to load output baskets: %w", err)
	}

	return slices.Map(dbBaskets, mapModelToTableOutputBasket), nil
}

func mapModelToTableOutputBasket(model *models.OutputBasket) *wdk.TableOutputBasket {
	return &wdk.TableOutputBasket{
		BasketID:  model.BasketID,
		UserID:    model.UserID,
		CreatedAt: model.CreatedAt,
		UpdatedAt: model.UpdatedAt,
		BasketConfiguration: wdk.BasketConfiguration{
			Name:                    model.Name,
			NumberOfDesiredUTXOs:    model.NumberOfDesiredUTXOs,
			MinimumDesiredUTXOValue: model.MinimumDesiredUTXOValue,
		},
	}
}
```

pkg/internal/storage/repo/outputs.go
```diff
@@ -135,3 +135,62 @@ func (o *Outputs) ListAndCountOutputs(ctx context.Context, filter entity.ListOut
...
+		if out.GetSpentBy() != 0 {
+			inputMap[out.GetSpentBy()] = append(inputMap[out.GetSpentBy()], tableOut)
+		}
+		outputMap[out.TransactionID] = append(outputMap[out.TransactionID], tableOut)
+	}
+
+	return inputMap, outputMap, nil
+}
+
+func (o *Outputs) FindBasketsByIDs(ctx context.Context, basketIDs []int) ([]*wdk.TableOutputBasket, error) {
```
chris-4chain: This should be rather moved to `baskets` repo

pkg/internal/storage/repo/proven_tx_req.go
```diff
@@ -278,3 +278,29 @@ func (p *ProvenTxReq) SetStatusForProvenTxAboveAttempts(ctx context.Context, att
 	}
 	return nil
 }
+
+func (p *ProvenTxReq) FindRawTxByTxIDs(ctx context.Context, txIDs []string) (map[string][]byte, error) {
```
chris-4chain: Above there is a very similar method `FindProvenTxRawTX` - consider moving this one right below it.

Also consider more consistent name, e.g `FindProvenTxRawTXs`

pkg/storage/internal/actions/list_actions.go
```diff
@@ -0,0 +1,241 @@
...
+	var inputMap, outputMap map[uint][]*wdk.TableOutput
+	if args.IncludeInputs || args.IncludeOutputs {
+		inputMap, outputMap, err = l.outputsRepo.FindInputsAndOutputsWithBuckets(ctx, txIDs)
+		if err != nil {
+			return nil, fmt.Errorf("failed to fetch inputs/outputs: %w", err)
+		}
+	}
+
+	basketIDSet := map[int]struct{}{}
+	for _, outputs := range outputMap {
```
chris-4chain: **Potential nil map dereference with outputMap**
   If neither `args.IncludeInputs` nor `args.IncludeOutputs` is true, `outputMap` will remain nil, causing a panic in the loop over `outputMap` later.

pkg/storage/internal/actions/list_actions.go
```diff
@@ -0,0 +1,241 @@
...
+		if args.IncludeLabels {
+			if labels, ok := labelMap[tx.TransactionID]; ok {
+				action.Labels = slices.Clone(labels)
+			} else {
+				action.Labels = []string{}
+			}
+		}
+
+		inputs := inputMap[tx.TransactionID]
+		outputs := outputMap[tx.TransactionID]
```
chris-4chain: Same problem with `nil` map dereference if neither `args.IncludeInputs` nor ``args.IncludeOutputs is true

pkg/internal/storage/database/models/tx_labels.go
```diff
@@ -0,0 +1,7 @@
+package models
+
+type TransactionLabels struct {
```
chris-4chain, line 3: What is this model? I can't see it in `AutoMigration` list.

pkg/internal/storage/repo/transactions.go
```diff
@@ -334,3 +334,80 @@ func (txs *Transactions) mapModelToTableTransaction(model *models.Transaction) *
...
+			query = query.Where("status IN ?", filter.Status)
+		}
+
+		if len(filter.Labels) > 0 {
+			labelSubQuery := tx.Model(&models.TransactionLabels{}).
+				Select("transaction_id").
+				Where("label_name IN ?", filter.Labels).
+				Where("label_user_id = ?", userID)
+
+			if filter.LabelQueryMode == "all" {
```
chris-4chain: Since it's our filter struct, I would prefer to change it to more straightforward and use a boolean value. 

Eg.
`filter.IncludeAllLabels` (or something desctriptive) - e.g. I don't understand what this "all" does :shrug:

pkg/wdk/storage_list_actions_args.go
```diff
@@ -0,0 +1,62 @@
...
+// ListActionsResult defines the result of listing actions
+type ListActionsResult struct {
+	TotalActions uint32         `json:"totalActions"`
+	Actions      []WalletAction `json:"actions"`
+}
+
+// WalletAction represents a transaction in the wallet
+type WalletAction struct {
+	TxID        string               `json:"txid"`
+	Satoshis    uint64               `json:"satoshis"`
```
chris-4chain: Shouldn't be `primitives.SatoshiValue`?

pkg/storage/internal/actions/list_actions.go
```diff
@@ -0,0 +1,241 @@
...
+	var txIDStrs []string
+
+	for i, tx := range txs {
+		txIDs[i] = tx.TransactionID
+		if tx.TxID != nil {
+			txIDStrs = append(txIDStrs, *tx.TxID)
+		}
+
+		actions[i] = wdk.WalletAction{
+			Satoshis:    must.ConvertToUInt64(to.NoLessThan(tx.Satoshis, 0)),
```
chris-4chain: Outgoing transactions have `negative` values - so **this logic cannot work**. 
It should be checked if `ts-version` returns negative value or "absolute value" - and adjusted here.

pkg/storage/internal/actions/list_actions.go
```diff
@@ -0,0 +1,241 @@
...
+		}
+		if tx.TxID != nil {
+			actions[i].TxID = *tx.TxID
+		}
+		if tx.Version != nil {
+			actions[i].Version = *tx.Version
+		}
+		if tx.LockTime != nil {
+			actions[i].LockTime = *tx.LockTime
+		}
```
chris-4chain:
```suggestion
			Description: tx.Description,
			TxID:        optional.OfPtr(tx.TxID).OrZeroValue(),
			Version:     optional.OfPtr(tx.Version).OrZeroValue(),
			LockTime:    optional.OfPtr(tx.LockTime).OrZeroValue(),
		}
```

pkg/storage/internal/actions/list_actions.go
```diff
@@ -0,0 +1,241 @@
...
+	"slices"
+
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/internal/logging"
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/internal/storage/entity"
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/wdk"
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/wdk/primitives"
+	"github.com/bsv-blockchain/go-sdk/transaction"
+	"github.com/go-softwarelab/common/pkg/must"
+	"github.com/go-softwarelab/common/pkg/seq"
+	comm_slices "github.com/go-softwarelab/common/pkg/slices"
```
chris-4chain: Eliminate snake case

pkg/storage/internal/actions/list_actions.go
```diff
@@ -0,0 +1,241 @@
...
+
+	basketIDSet := map[int]struct{}{}
+	for _, outputs := range outputMap {
+		for _, out := range outputs {
+			if out.BasketID != nil {
+				basketIDSet[*out.BasketID] = struct{}{}
+			}
+		}
+	}
+	basketIDs := seq.ToSlice(maps.Keys(basketIDSet), make([]int, 0, len(basketIDSet)))
```
chris-4chain: This could be moved to a separate helper method: `uniqueBasketIDs`

pkg/storage/internal/actions/list_actions.go
```diff
@@ -0,0 +1,241 @@
...
+			return nil, fmt.Errorf("failed to batch load raw transactions: %w", err)
+		}
+	}
+
+	for i, tx := range txs {
+		action := &actions[i]
+
+		if args.IncludeLabels {
+			if labels, ok := labelMap[tx.TransactionID]; ok {
+				action.Labels = slices.Clone(labels)
```
chris-4chain: Why is clonning necessary?

chris-4chain (changes requested):

pkg/storage/internal/actions/list_actions.go
ac4ch: linter suggestion as before i had this way
`Replace append with slices.Clone`
```go
if args.IncludeLabels {
	action.Labels = append([]string(nil), labelMap[tx.TransactionID]...)
}
```

pkg/storage/internal/actions/list_actions.go
ac4ch: ok i made same thing as u did here
```go
resultOutputs[i] = wdk.StorageCreateTransactionSdkOutput{
			Vout:             output.Vout,
			ProvidedBy:       output.ProvidedBy,
			Purpose:          output.Purpose,
			DerivationSuffix: output.DerivationSuffix,
			ValidCreateActionOutput: wdk.ValidCreateActionOutput{
				Satoshis:           primitives.SatoshiValue(must.ConvertToUInt64(output.Satoshis)),
				OutputDescription:  primitives.String5to2000Bytes(output.Description),
				CustomInstructions: output.CustomInstructions,
				LockingScript:      optional.OfPtr(output.LockingScript).OrZeroValue(),
			},
		}
```

`Satoshis:           primitives.SatoshiValue(must.ConvertToUInt64(output.Satoshis)),`

i can write logic to adjust negative values but i do not see example in our code how this was handled, could u suggest if above solution is wrong ?

ac4ch: ok we changing it to int64

pkg/storage/internal/actions/list_actions.go
```diff
@@ -0,0 +1,240 @@
...
+			} else {
+				action.Labels = []string{}
+			}
+		}
+
+		inputs := inputMap[tx.TransactionID]
+		outputs := outputMap[tx.TransactionID]
+
+		if args.IncludeOutputs {
+			action.Outputs = mapToWalletActionOutputs(outputs, basketMap)
```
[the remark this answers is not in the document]
ac4ch: this should be fixed by my latest refactoring

pkg/wdk/storage_list_actions_args.go
```diff
@@ -0,0 +1,62 @@
+package wdk
+
+import (
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/wdk/primitives"
+)
+
+// ListActionsArgs defines arguments for listing actions (transactions)
+type ListActionsArgs struct {
+	Labels                           []primitives.StringUnder300 `json:"labels,omitempty"`
+	Limit                            uint                        `json:"limit,omitempty"`
```
dorzepowski: why not type: `primitives.PositiveIntegerDefault10Max10000`?

pkg/wdk/storage_list_actions_args.go
```diff
@@ -0,0 +1,62 @@
...
+	Offset                           uint                        `json:"offset,omitempty"`
+	IncludeAllLabels                 bool                        `json:"includeAllLabels,omitempty"` // "all=true" or "any=false"
```
dorzepowski: you replaced labelQueryMode with this - which is not compatible between TS and Go, you should rather use enum for that

pkg/wdk/storage_list_actions_args.go
```diff
@@ -0,0 +1,62 @@
...
+	IncludeInputs                    bool                        `json:"includeInputs,omitempty"`
+	IncludeOutputs                   bool                        `json:"includeOutputs,omitempty"`
+	IncludeLabels                    bool                        `json:"includeLabels,omitempty"`
+	IncludeInputSourceLockingScripts bool                        `json:"includeInputSourceLockingScripts,omitempty"`
+	IncludeInputUnlockingScripts     bool                        `json:"includeInputUnlockingScripts,omitempty"`
+	IncludeOutputLockingScripts      bool                        `json:"includeOutputLockingScripts,omitempty"`
+	IncludeCustomInstructions        bool                        `json:"includeCustomInstructions,omitempty"`
```
dorzepowski: I don't see this in typescript version 
How ts is handling custom instructions without this flag?

pkg/wdk/storage_list_actions_args.go
```diff
@@ -0,0 +1,62 @@
...
+	Labels                           []primitives.StringUnder300 `json:"labels,omitempty"`
+	Limit                            uint                        `json:"limit,omitempty"`
+	Offset                           uint                        `json:"offset,omitempty"`
+	IncludeAllLabels                 bool                        `json:"includeAllLabels,omitempty"` // "all=true" or "any=false"
+	IncludeInputs                    bool                        `json:"includeInputs,omitempty"`
+	IncludeOutputs                   bool                        `json:"includeOutputs,omitempty"`
+	IncludeLabels                    bool                        `json:"includeLabels,omitempty"`
+	IncludeInputSourceLockingScripts bool                        `json:"includeInputSourceLockingScripts,omitempty"`
+	IncludeInputUnlockingScripts     bool                        `json:"includeInputUnlockingScripts,omitempty"`
+	IncludeOutputLockingScripts      bool                        `json:"includeOutputLockingScripts,omitempty"`
```
dorzepowski: the type of those fields should probably be primitives.BooleanDefaultFalse

pkg/wdk/storage_list_actions_args.go
```diff
@@ -0,0 +1,62 @@
+package wdk
+
+import (
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/wdk/primitives"
+)
+
+// ListActionsArgs defines arguments for listing actions (transactions)
+type ListActionsArgs struct {
```
dorzepowski, line 8: SeekPermission flag is missing - add task or `// TODO` to handle it later.

pkg/storage/internal/actions/list_actions_mapping.go
```diff
@@ -0,0 +1,249 @@
...
+		Labels:           labelNames,
+		Status:           statuses,
+		IncludeAllLabels: args.IncludeAllLabels,
+		Limit:            must.ConvertToIntFromUnsigned(args.Limit),
+		Offset:           must.ConvertToIntFromUnsigned(args.Offset),
+	}
+}
+
+func (l *listActions) mapTransactionsToActions(txs []*wdk.TableTransaction) ([]uint, []string, []wdk.WalletAction) {
+	txIDs := make([]uint, len(txs))
```
dorzepowski: ### minor
maybe `transactionIDs` and then txIDStrs - could be `txIDs`

pkg/storage/internal/actions/list_actions.go
```diff
@@ -0,0 +1,70 @@
...
+func (l *listActions) ListActions(ctx context.Context, auth wdk.AuthID, args *wdk.ListActionsArgs) (*wdk.ListActionsResult, error) {
+	userID := *auth.UserID
+	filter := l.toFilterParams(userID, args)
+
+	txs, total, err := l.transactionsRepo.ListAndCountActions(ctx, userID, filter)
+	if err != nil {
+		return nil, fmt.Errorf("failed to list transactions: %w", err)
+	}
+
+	txIDs, txIDStrs, actions := l.mapTransactionsToActions(txs)
```
dorzepowski: ### minor - suggestion
Same renaming as for variables in mapTransactionsToActions

chris-4chain (approved):

pkg/storage/provider_list_actions_test.go
```diff
@@ -0,0 +1,144 @@
...
+	require.NoError(t, err)
+
+	// When:
+	args := wdk.ListActionsArgs{
+		Limit:            10,
+		Offset:           0,
+		IncludeAllLabels: false, // "any=false" or "all=true"
+		IncludeLabels:    true,
+		IncludeOutputs:   true,
+		IncludeInputs:    true,
```
dorzepowski: we should have separated tests for each of available flags and also for limit and offset.

But you can add it in another PR, if you want to merge this one.

pkg/wdk/storage_list_actions_args.go
ac4ch: i dont know why .. it should be `primitives.PositiveIntegerDefault10Max10000` ?

pkg/wdk/storage_list_actions_args.go
chris-4chain: I suggested this but only for "our internal" filter/entity.
I didn't mean to change public interface...

chris-4chain: https://github.com/bsv-blockchain/go-wallet-toolbox/pull/257#discussion_r2120432749

pkg/wdk/storage_list_actions_args.go
dorzepowski: in TS it is this type, and actually it has the validation for max build in (it also could have method for default value)

pkg/wdk/storage_list_actions_args.go
dorzepowski: For internal usage, I agree, but for the public interface, we must (unfortunately) stick to API provided by TS, so we cannot remove fields or change their type because it won't work in integration TS <-> GO both directions.

pkg/internal/validate/list_actions_args.go
```diff
@@ -0,0 +1,49 @@
...
+)
+
+func ListActionsArgs(args *wdk.ListActionsArgs) error {
+	if args == nil {
+		return fmt.Errorf("args cannot be nil")
+	}
+
+	switch strings.ToLower(args.LabelQueryMode) {
+	case "", "any":
+		args.LabelQueryMode = "any"
```
dorzepowski: this should be done, before validation, otherwise, the validation has some side effect, that makes it harder to spot the bug.

dorzepowski (approved):

PR #265, opened 2025-06-05
feat: init sync/backup mechanism 

chris-4chain:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter

### DIFF
--- a/pkg/internal/storage/database/scopes/scopes.go
+++ b/pkg/internal/storage/database/scopes/scopes.go
@@ -42,9 +42,10 @@ func Since(since *queryopts.Since) func(*gorm.DB) *gorm.DB {
 	return func(db *gorm.DB) *gorm.DB {
 		since.ApplyDefaults()
 		return db.Where(
-			"?? >= ?",
-			clause.Column{Name: since.Field},
-			since.Time,
+			&clause.Gte{
+				Column: clause.Column{Name: since.Field},
+				Value:  since.Time,
+			},
 		)
 	}
 }
--- a/pkg/internal/storage/database/scopes/scopes.go
+++ b/pkg/internal/storage/database/scopes/scopes.go
@@ -50,8 +50,8 @@ func Since(since *queryopts.Since) func(*gorm.DB) *gorm.DB {
 	}
 }
 
-func FromQueryOpts(opts ...queryopts.Options) []func(*gorm.DB) *gorm.DB {
-	options := queryopts.MergeOptions(opts...)
+func FromQueryOpts(opts []queryopts.Options) []func(*gorm.DB) *gorm.DB {
+	options := queryopts.MergeOptions(opts)
 
 	var sc []func(*gorm.DB) *gorm.DB
 	if options.Page != nil {
--- a/pkg/internal/storage/database/scopes/scopes.go
+++ b/pkg/internal/storage/database/scopes/scopes.go
@@ -50,7 +50,7 @@ func Since(since *queryopts.Since) func(*gorm.DB) *gorm.DB {
 	}
 }
 
-func FromQueryOpts(opts ...queryopts.QueryOptsUnion) []func(*gorm.DB) *gorm.DB {
+func FromQueryOpts(opts ...queryopts.Options) []func(*gorm.DB) *gorm.DB {
 	options := queryopts.MergeOptions(opts...)
 
 	var sc []func(*gorm.DB) *gorm.DB
--- a/pkg/internal/storage/database/scopes/scopes.go
+++ b/pkg/internal/storage/database/scopes/scopes.go
@@ -1,15 +1,20 @@
 package scopes
 
 import (
-	"github.com/4chain-ag/go-wallet-toolbox/pkg/internal/storage/paging"
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/internal/storage/queryopts"
 	"gorm.io/gorm"
+	"gorm.io/gorm/clause"
 )
 
-// Paginate is a Scope function that handle pagination.
-func Paginate(page *paging.Page) func(db *gorm.DB) *gorm.DB {
+// Paginate is a Scope function that handles pagination.
+func Paginate(page *queryopts.Paging) func(db *gorm.DB) *gorm.DB {
 	page.ApplyDefaults()
 	return func(db *gorm.DB) *gorm.DB {
-		return db.Order(page.SortBy + " " + page.Sort).Offset(page.Offset).Limit(page.Limit)
+		orderClause := clause.OrderByColumn{
+			Column: clause.Column{Name: page.SortBy},
+			Desc:   page.IsDesc(),
+		}
+		return db.Order(orderClause).Offset(page.Offset).Limit(page.Limit)
 	}
 }
 
@@ -31,4 +36,29 @@ func BasketID(basketID int) func(*gorm.DB) *gorm.DB {
 	return func(db *gorm.DB) *gorm.DB {
 		return db.Where("basket_id = ?", basketID)
 	}
+}
+
+func Since(since *queryopts.Since) func(*gorm.DB) *gorm.DB {
+	return func(db *gorm.DB) *gorm.DB {
+		since.ApplyDefaults()
+		return db.Where(
+			"?? >= ?",
+			clause.Column{Name: since.Field},
+			since.Time,
+		)
+	}
+}
+
+func FromQueryOpts(opts ...queryopts.QueryOptsUnion) []func(*gorm.DB) *gorm.DB {
+	options := queryopts.MergeOptions(opts...)
+
+	var sc []func(*gorm.DB) *gorm.DB
+	if options.Page != nil {
+		sc = append(sc, Paginate(options.Page))
+	}
+	if options.Since != nil {
+		sc = append(sc, Since(options.Since))
+	}
+
+	return sc
 }
--- a/pkg/internal/storage/queryopts/union.go
+++ b/pkg/internal/storage/queryopts/union.go
@@ -1,0 +1,36 @@
+package queryopts
+
+type QueryOptsUnion struct {
+	Page  *Paging
+	Since *Since
+}
+
+func WithPage(page Paging) QueryOptsUnion {
+	return QueryOptsUnion{
+		Page:  &page,
+	}
+}
+
+func WithSince(since Since) QueryOptsUnion {
+	return QueryOptsUnion{
+		Since: &since,
+	}
+}
+
+func MergeOptions(opts ...QueryOptsUnion) QueryOptsUnion {
+	if len(opts) == 0 {
+		return QueryOptsUnion{}
+	}
+
+	result := QueryOptsUnion{}
+	for _, opt := range opts {
+		if opt.Page != nil {
+			result.Page = opt.Page
+		}
+		if opt.Since != nil {
+			result.Since = opt.Since
+		}
+	}
+
+	return result
+}
--- a/pkg/internal/storage/queryopts/union.go
+++ b/pkg/internal/storage/queryopts/union.go
@@ -1,28 +1,28 @@
 package queryopts
 
-type QueryOptsUnion struct {
+type Options struct {
 	Page  *Paging
 	Since *Since
 }
 
-func WithPage(page Paging) QueryOptsUnion {
-	return QueryOptsUnion{
+func WithPage(page Paging) Options {
+	return Options{
 		Page:  &page,
 	}
 }
 
-func WithSince(since Since) QueryOptsUnion {
-	return QueryOptsUnion{
+func WithSince(since Since) Options {
+	return Options{
 		Since: &since,
 	}
 }
 
-func MergeOptions(opts ...QueryOptsUnion) QueryOptsUnion {
+func MergeOptions(opts ...Options) Options {
 	if len(opts) == 0 {
-		return QueryOptsUnion{}
+		return Options{}
 	}
 
-	result := QueryOptsUnion{}
+	result := Options{}
 	for _, opt := range opts {
 		if opt.Page != nil {
 			result.Page = opt.Page
--- a/pkg/internal/storage/queryopts/union.go
+++ b/pkg/internal/storage/queryopts/union.go
@@ -17,7 +17,7 @@ func WithSince(since Since) Options {
 	}
 }
 
-func MergeOptions(opts ...Options) Options {
+func MergeOptions(opts []Options) Options {
 	if len(opts) == 0 {
 		return Options{}
 	}
--- a/pkg/storage/internal/sync/chunker.go
+++ b/pkg/storage/internal/sync/chunker.go
@@ -9,7 +9,7 @@ import (
 
 type Chunker interface {
 	Name() string
-	MaxDivider() int
+	MaxDivider() uint64
 	IsApplicable(requestedEntities OffsetsLookup) bool
-	Process(ctx context.Context, userID, limit, relativeOffset int, offsetsLookup OffsetsLookup, since *time.Time, result *wdk.SyncChunk) (num int, err error)
+	Process(ctx context.Context, userID int, limit, relativeOffset uint64, offsetsLookup OffsetsLookup, since *time.Time, result *wdk.SyncChunk) (num uint64, err error)
 }
--- a/pkg/storage/internal/sync/chunker.go
+++ b/pkg/storage/internal/sync/chunker.go
@@ -1,0 +1,14 @@
+package sync
+
+import (
+	"context"
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/wdk"
+	"time"
+)
+
+type Chunker interface {
+	Name() string
+	MaxDivider() int
+	IsApplicable(requestedEntities OffsetsLookup) bool
+	Process(ctx context.Context, userID, limit, relativeOffset int, offsetsLookup OffsetsLookup, since *time.Time, result *wdk.SyncChunk) (num int, err error)
+}
--- a/pkg/storage/internal/sync/chunker.go
+++ b/pkg/storage/internal/sync/chunker.go
@@ -2,8 +2,9 @@ package sync
 
 import (
 	"context"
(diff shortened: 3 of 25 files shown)


pkg/internal/storage/queryopts/union.go
```diff
@@ -0,0 +1,36 @@
...
+
+	result := Options{}
+	for _, opt := range opts {
+		if opt.Page != nil {
+			result.Page = opt.Page
+		}
+		if opt.Since != nil {
+			result.Since = opt.Since
+		}
+	}
```
mgosek-4chain, line 33: Are you sure this loop is correct? It looks more like overriding instead of merging. Or maybe I missed something 😸

pkg/storage/internal/sync/chunker_baskets.go
```diff
@@ -0,0 +1,58 @@
...
+			Limit:  must.ConvertToIntFromUnsigned(limit),
+			Offset: must.ConvertToIntFromUnsigned(offset),
+		}),
+	}
+
+	if since != nil {
+		opts = append(opts, queryopts.WithSince(queryopts.Since{Time: *since}))
+	}
+
+	rows, err := c.repo.FindBasketsByUserID(ctx, userID, opts...)
```
mgosek-4chain, line 50: It looks like you don't have to create the options slice, instead you can just pass the `WithPage` as argument.

mgosek-4chain (changes requested):

pkg/internal/storage/queryopts/union.go
chris-4chain: I'm sure.
By design, each `With` function produces the `Options object` but with one assumption: only one field is `not-nil`.
Here, those objects (that contain only one field `not-nil`) are `"merged"` to single struct. 

This is slightly different approach than "well-known" `functional options`. Note, that in that pattern there is also a problem when a programmer provides some option twice.
- I would say, that here (in this pattern) is even simpler to detect the issue. 

Why I've introduced this pattern?
- Because I thought, it's easier to debug the options array, because they are normal structs - comparing to `function variables` - where you cannot see how it'll modify the options, until it's applied.

mgosek-4chain (approved):

pkg/storage/internal/sync/sync_chunk_action.go
```diff
@@ -0,0 +1,130 @@
...
+
+			relativeOffset += num
+		}
+	}
+
+	return nil
+}
+
+func (s *syncChunkAction) makeOffsetsLookup(args *wdk.RequestSyncChunkArgs) OffsetsLookup {
+	offsetsLookup := make(OffsetsLookup)
```
dorzepowski: ### minor thing
wouldn't it be more efficient?
```suggestion
	offsetsLookup := make(OffsetsLookup, len(args.Offsets))
```

pkg/storage/internal/sync/sync_chunk_action.go
```diff
@@ -0,0 +1,130 @@
...
+	offsetsLookup := s.makeOffsetsLookup(args)
+
+	for _, chunker := range s.chunkers {
+		if !chunker.IsApplicable(offsetsLookup) {
+			continue
+		}
+
+		var relativeOffset uint64
+
+		for range s.safeLoopIterator(maxIterations) {
```
dorzepowski: I like that you used the common library, but it seems that it would be sufficient to do it this way:
```suggestion
		for range maxIterations {
```

pkg/storage/internal/sync/chunker.go
```diff
@@ -0,0 +1,15 @@
...
+import (
+	"context"
+	"time"
+
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/wdk"
+)
+
+type Chunker interface {
+	Name() string
+	MaxDivider() uint64
```
dorzepowski: I don't understand the idea of MaxDivider, that's why I prepared a suggestion to replace it with MaxPageSize,
https://github.com/bsv-blockchain/go-wallet-toolbox/pull/269

 but maybe this is something smarter that I don't understand 🤔

pkg/storage/internal/sync/sync_chunk_action.go
```diff
@@ -0,0 +1,130 @@
...
+			if err = ctx.Err(); err != nil {
+				return fmt.Errorf("context canceled, aborting: %w", err)
+			}
+
+			freeSlots := args.MaxItems - itemsCounter
+			limit, err := to.UInt64(min(freeSlots, max(minPageSize, args.MaxItems/chunker.MaxDivider())))
+			if err != nil {
+				return fmt.Errorf("failed to calculate limit: %w", err)
+			}
```
dorzepowski: I belive that this could be done that way with clearer intention:
```suggestion
			limit := to.ValueBetween(args.MaxItems/chunker.MaxDivider(), minPageSize, freeSlots)
```

pkg/storage/internal/sync/sync_chunk_action.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+	return offsetsLookup
+}
+
+func (s *syncChunkAction) safeLoopIterator(maxIterations int) iter.Seq[int] {
+	return seq.RangeWithStep(0, maxIterations, 1)
+}
+
+func (s *syncChunkAction) approxJSONSize(chunk *wdk.SyncChunk) uint64 {
+	// TODO it could be less precise and use a more efficient way to estimate size
```
dorzepowski: I don't understand this comment 🙈

pkg/storage/internal/sync/sync_chunk_action.go
```diff
@@ -0,0 +1,130 @@
...
+
+	for _, chunker := range s.chunkers {
+		if !chunker.IsApplicable(offsetsLookup) {
+			continue
+		}
+
+		var relativeOffset uint64
+
+		for range s.safeLoopIterator(maxIterations) {
+			if err = ctx.Err(); err != nil {
```
dorzepowski: hmm...
```suggstion
			if ctx.Err() != nil {
```

pkg/internal/storage/database/scopes/scopes.go
```diff
@@ -32,3 +37,29 @@ func BasketID(basketID int) func(*gorm.DB) *gorm.DB {
...
+		return db.Where(
+			&clause.Gte{
+				Column: clause.Column{Name: since.Field},
+				Value:  since.Time,
+			},
+		)
+	}
+}
+
+func FromQueryOpts(opts ...queryopts.Options) []func(*gorm.DB) *gorm.DB {
```
dorzepowski: maybe it doesn't need to be vararg?

pkg/internal/storage/database/scopes/scopes.go
```diff
@@ -32,3 +37,29 @@ func BasketID(basketID int) func(*gorm.DB) *gorm.DB {
...
+	options := queryopts.MergeOptions(opts...)
```
dorzepowski: and same here, maybe it doesn't need to be vararg

pkg/internal/storage/queryopts/union.go
```diff
@@ -0,0 +1,36 @@
...
+}
+
+func WithSince(since Since) Options {
+	return Options{
+		Since: &since,
+	}
+}
+
+func MergeOptions(opts ...Options) Options {
```
dorzepowski: mabye it doesn't need to be public

pkg/internal/storage/queryopts/union.go
```diff
@@ -0,0 +1,36 @@
...
+	if len(opts) == 0 {
+		return Options{}
+	}
+
+	result := Options{}
+	for _, opt := range opts {
+		if opt.Page != nil {
```
dorzepowski, line 27: ### just saying
I'm not sure about this pattern, but maybe it's not worth to burn a time on discussing this as it's still quite small part of code.

pkg/storage/internal/sync/sync_chunk_action.go
```diff
@@ -0,0 +1,130 @@
...
+			if err != nil {
+				return fmt.Errorf("chunker %s failed: %w", chunker.Name(), err)
+			}
+
+			if num == 0 {
+				break
+			}
+
+			itemsCounter += num
+			reachedMax := itemsCounter >= args.MaxItems || s.approxJSONSize(result) >= args.MaxRoughSize
```
dorzepowski: ### just saying
I know that this is the issue also in TS, but it can be that the result json will exceed th MaxRoughSize - maybe we should make a task for solving this, or not 😀

pkg/storage/internal/sync/sync_chunk_action.go
dorzepowski: It would be awesome to extract some methods for processing each chunker and for processing each page
But it would probably require to create additional struct (for example chunksProcessor which would hold the state of the iterations and would know when to stop.

pkg/storage/internal/sync/sync_chunk_action.go
```diff
@@ -0,0 +1,130 @@
...
+			if reachedMax {
+				// NOTE: This breaks also the "chunkers" loop because we don't have any more free slots.
+				return nil
+			}
+
+			relativeOffset += num
```
dorzepowski: I feel that this lines makes the safety guard, at the for loop statement, redundant (and could lead to errors: too eager returns), maybe those could be made a check in a for loop or something 🤔

pkg/storage/provider_get_sync_chunk_test.go
```diff
@@ -0,0 +1,216 @@
...
+		IdentityKey:            testusers.Alice.IdentityKey(t),
+		MaxItems:               10,
+		MaxRoughSize:           100_000,
+
+		Offsets: []wdk.SyncOffsets{
+			{
+				Name:   wdk.OutputBasketEntityName,
+				Offset: 0,
+			},
+			// TODO: Add more offsets for other entities when implemented
```
dorzepowski, line 38: maybe we should have a separated test for each entity. 
And then one or few tests with several entities just to ensure that limits are working properly and that it's collectin more then one entity.

pkg/storage/internal/sync/chunker_baskets.go
```diff
@@ -0,0 +1,58 @@
...
+func (c *chunkerBaskets) MaxDivider() uint64 {
+	return 1
+}
+
+func (c *chunkerBaskets) IsApplicable(requestedEntities OffsetsLookup) bool {
+	_, ok := requestedEntities[wdk.OutputBasketEntityName]
+	return ok
+}
+
+func (c *chunkerBaskets) Process(ctx context.Context, userID int, limit, relativeOffset uint64, offsetsLookup OffsetsLookup, since *time.Time, result *wdk.SyncChunk) (num uint64, err error) {
```
dorzepowski: I thought it would be easier to use Paging directly,
that's why I prepared a suggestion for that https://github.com/bsv-blockchain/go-wallet-toolbox/pull/268

pkg/storage/internal/sync/chunker.go
chris-4chain: This divides the maximum total items requested by a divisor specific to each data type. Different data types have different maxDivider values (ranging from 1 to 100 as seen in the chunkers array) to control their relative contribution to the overall result set.

NOTE that `args.MaxItems` is one for the whole request - so the `dividers` defined for each entity help to calculate **dynamic** "MaxPageSize".
- In contrast, you're proposing **static/constant** MaxPageSize, that doesn't depend on requested `args.MaxItems`.

To be honest, I like this `MaxDivider` approach - but I do understand that is not obvious - so my proposition is to describe this param in a doc-comment

pkg/storage/internal/sync/sync_chunk_action.go
chris-4chain: I merged your suggestion, so you already addressed that 🙈

pkg/storage/internal/sync/sync_chunk_action.go
chris-4chain: I merged your suggestion, so you already addressed that 🙈

pkg/storage/internal/sync/sync_chunk_action.go
chris-4chain: Marshalling the result object to JSON every iteration, only for approximate size calculation impacts to performance. 
- In this comment I wanted to note that concern to revisit it in the future (for example, during review).

pkg/storage/internal/sync/sync_chunk_action.go
chris-4chain: hmm...
```golang
if err := ctx.Err(); err != nil {
	return fmt.Errorf("context canceled, aborting: %w", err)
}
```

pkg/internal/storage/database/scopes/scopes.go
chris-4chain: Sure, for current use cases this is true 👍

pkg/internal/storage/queryopts/union.go
chris-4chain: `MergeOptions` is in `queryopts` and the usage is in `scopes`, so it do need to be public.

pkg/internal/storage/queryopts/union.go
chris-4chain: Agree, we'll see how it goes and maybe in the future we'll revisit this topic.

pkg/storage/internal/sync/sync_chunk_action.go
chris-4chain: I would treat this `MaxRoughSize` as a suggestion that:
👉 "it would be good to not exceed the maxsize, but if you already fetched the data just include it".

I think, this approach improves a little a overall performance of the whole syncing process because it prevents loading some rows twice.

pkg/storage/provider_get_sync_chunk_test.go
chris-4chain: Maybe,
Note that for this PR we have only single entity, so your suggestion is kina not-applicable. 
- But I'm going to extend and improve the tests when I add subsequent entities. 
- And then I can consider your suggestion.

pkg/storage/internal/sync/chunker_baskets.go
chris-4chain: Sure, I've merged your suggestion 👍

pkg/storage/internal/sync/sync_chunk_action.go
dorzepowski: Hmm... I would suggest to remove the comment - as we already have plenty of TODOs in code, and this should rather be optimised much much later when we can check the wallet efficiency.

pkg/storage/internal/sync/state.go
```diff
@@ -0,0 +1,72 @@
...
+			page.Next()
+			if !yield(page) {
+				return
+			}
+		}
+	}
+}
+
+func (s *chunkingState) chunkProcessed() bool {
+	return s.prevItemsCounter == s.itemsCounter || s.reachedMax()
```
dorzepowski, line 46: ### minor
I know that it's implementation detail, but this comparison of prevItemsCounter and itemsCounter is  a little confusing - you need to load much more code into context to understand the idea. I would rather suggest to create field `chunkerFinished` 
and set it to false in `doWhileChunkProcessed`
and set it to true update if `num == 0`

pkg/storage/internal/sync/chunker.go
dorzepowski: 1. > so the dividers defined for each entity help to calculate dynamic "MaxPageSize".

It's not true, they are used to calculate current page size, not the max page size - there is no exact limit in a page size

2. I see some potential issues in this approach:

- Let's say that the user has a lot of transactions - and have already 1bln of transactions in db
- let's say that single transaction item in json would add 0,5kB to the result json
- Let's say that he want to synchronise any number of transactions but want to only limit the rough size to 124kB
- So he is passing MaxUint64 as a MaxItems

and now, let's make a simple math in both approaches:
a. Current approach:
i. transaction divider is 100
ii. MaxUint64 / 100 = 184 467 440 737 095 516
iii. because 10 < 184 467 440 737 095 516 < MaxUint64 => limit = 184 467 440 737 095 516
iv. result = 1bln tx * 0,5KB = 1mln kB - it is much higher than 124kB

b. proposed approach
i. transaction limit is 1000
ii. because MaxUint64 > 1000 -> limit = 1000
iii. result = 1000 * 0,5kB = 1000kB -> still more then 124kB but not that much


3. Additionally from my perspective it's easier to make a MaxPageSize adjusted to the data that we have in db, 
for example in the future it could accept userId or maybe even a MaxRawSize or something else and with query  check the median size of BEEFs and according to this calculate more accurate max size - especially taking into account potential memory issues when loading a lot of data.


I adjusted the PR with that proposition to current changes https://github.com/bsv-blockchain/go-wallet-toolbox/pull/269, so please take a look.
BTW. I also notice (and solve) potential panic if MaxUint64 is passed as MaxItems and the divider is 1.

pkg/storage/provider_get_sync_chunk_test.go
```diff
@@ -0,0 +1,216 @@
...
+
+	// given:
+	givenProvider := given.Provider()
+	activeStorage := givenProvider.GORM()
+
+	args := wdk.RequestSyncChunkArgs{
+		FromStorageIdentityKey: "from_storage",
+		ToStorageIdentityKey:   "to_storage",
+		IdentityKey:            testusers.Alice.IdentityKey(t),
+		MaxItems:               10,
```
dorzepowski, line 30: It would be good to have a test for MaxUint64 as MaxItems - because currently it is panicing in that case

pkg/storage/internal/sync/state.go
chris-4chain: The idea is clear to me and I'd like to stick to it. 
- I don't like setting the flag in `doWhileChunkProcessed` - because it makes this iter-method "not-constant" method that changes the state (and makes this even more complex)

dorzepowski (approved):

PR #273, opened 2025-06-11
[FEATURE] WalletServices: add MerklePath method to WoC service

### LINKED ISSUE #230: [FEATURE] WalletServices: add MerklePath method to WoC service

ac4ch:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter

### DIFF
--- a/pkg/internal/testabilities/testservices/fixture_woc.go
+++ b/pkg/internal/testabilities/testservices/fixture_woc.go
@@ -2,18 +2,30 @@ package testservices
 
 import (
 	"fmt"
+	"net"
 	"net/http"
+	"regexp"
 	"testing"
 
 	"github.com/4chain-ag/go-wallet-toolbox/pkg/defs"
+	"github.com/go-resty/resty/v2"
 	"github.com/go-softwarelab/common/pkg/to"
 	"github.com/jarcoal/httpmock"
 )
 
 type WhatsOnChainFixture interface {
 	WillRespondWithRates(status int, content string, err error)
-
 	WillRespondWithRawTx(status int, txID, rawTx string, err error)
+	WillRespondWithMerklePath(status int, txID string, responseBody string)
+	WillRespondWithBlockHeader(status int, blockHash string, responseBody string)
+
+	WhenQueryingMerklePath(txID string) WhatsOnChainMerklePathQueryFixture
+	WhenQueryingBlockHeader(blockHash string) WhatsOnChainBlockHeaderQueryFixture
+
+	Transport() *httpmock.MockTransport
+	HttpClient() *resty.Client
+
+	WillBeUnreachable() error
 }
 
 type wocFixture struct {
@@ -64,4 +76,53 @@ func (f *wocFixture) WillRespondWithRawTx(status int, txID, rawTx string, err er
 
 	url := fmt.Sprintf("https://api.whatsonchain.com/v1/bsv/%s/tx/%s/hex", f.network, txID)
 	f.transport.RegisterResponder("GET", url, responder(status, rawTx))
+}
+
+func (f *wocFixture) WillRespondWithMerklePath(status int, txID, responseBody string) {
+	responder := func(*http.Request) (*http.Response, error) {
+		resp := httpmock.NewStringResponse(status, responseBody)
+		resp.Header.Set("Content-Type", "application/json")
+		return resp, nil
+	}
+	url := fmt.Sprintf("https://api.whatsonchain.com/v1/bsv/%s/tx/%s/proof/tsc", f.network, txID)
+	f.transport.RegisterResponder("GET", url, responder)
+}
+
+func (f *wocFixture) WillRespondWithBlockHeader(status int, blockHash, responseBody string) {
+	responder := func(*http.Request) (*http.Response, error) {
+		resp := httpmock.NewStringResponse(status, responseBody)
+		resp.Header.Set("Content-Type", "application/json")
+		return resp, nil
+	}
+	url := fmt.Sprintf("https://api.whatsonchain.com/v1/bsv/%s/block/%s/header", f.network, blockHash)
+	f.transport.RegisterResponder("GET", url, responder)
+}
+
+func (f *wocFixture) WhenQueryingMerklePath(txID string) WhatsOnChainMerklePathQueryFixture {
+	return &wocMerklePathQueryFixture{fixture: f, txID: txID}
+}
+
+func (f *wocFixture) WhenQueryingBlockHeader(blockHash string) WhatsOnChainBlockHeaderQueryFixture {
+	return &wocBlockHeaderQueryFixture{fixture: f, blockHash: blockHash}
+}
+
+func (f *wocFixture) Transport() *httpmock.MockTransport {
+	return f.transport
+}
+
+func (f *wocFixture) HttpClient() *resty.Client {
+	client := resty.New()
+	client.GetClient().Transport = f.transport
+	return client
+}
+
+func (f *wocFixture) WillBeUnreachable() error {
+	err := net.UnknownNetworkError("tests defined this endpoint as unreachable")
+	f.TB.Helper()
+	f.transport.RegisterRegexpResponder(
+		http.MethodGet,
+		regexp.MustCompile(`^https://api\.whatsonchain\.com/.*`),
+		httpmock.NewErrorResponder(err),
+	)
+	return err
 }
--- a/pkg/services/internal/whatsonchain/merkle_path.go
+++ b/pkg/services/internal/whatsonchain/merkle_path.go
@@ -1,0 +1,216 @@
+package whatsonchain
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/wdk"
+	"github.com/bsv-blockchain/go-sdk/chainhash"
+	"github.com/bsv-blockchain/go-sdk/transaction"
+	"github.com/go-resty/resty/v2"
+	"github.com/go-softwarelab/common/pkg/must"
+	"github.com/go-softwarelab/common/pkg/to"
+)
+
+// tscProof represents the response from the WoC /tx/{txid}/proof/tsc endpoint
+type tscProof struct {
+	Index  int      `json:"index"`
+	Nodes  []string `json:"nodes"`
+	Target string   `json:"target"` // block hash
+	TxOrID string   `json:"txOrId"` // txid
+}
+
+// blockHeaderResponse represents the response from WoC /block/{hash}/header
+type blockHeaderResponse struct {
+	Height     int    `json:"height"`
+	MerkleRoot string `json:"merkleRoot"`
+}
+
+// MerklePath retrieves the merkle path for a transaction using WoC TSC proof.
+func (woc *WhatsOnChain) MerklePath(ctx context.Context, txID string) (*wdk.MerklePathResult, error) {
+	proof, err := woc.getTscProof(ctx, txID)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get TSC proof: %w", err)
+	}
+	if proof == nil {
+		// Proof not found
+		return &wdk.MerklePathResult{Name: ServiceName}, nil
+	}
+
+	header, err := woc.hashToHeader(ctx, proof.Target)
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch block header: %w", err)
+	}
+
+	merklePath, err := convertTscProofToMerklePath(txID, proof.Index, proof.Nodes, header.Height)
+	if err != nil {
+		return nil, fmt.Errorf("failed to convert proof to merkle path: %w", err)
+	}
+
+	merkleRoot, err := merklePath.ComputeRootHex(&txID)
+	if err != nil {
+		return nil, fmt.Errorf("failed to compute merkle root: %w", err)
+	}
+	if merkleRoot != header.MerkleRoot {
+		return nil, fmt.Errorf("computed merkle root %q does not match block header %q", merkleRoot, header.MerkleRoot)
+	}
+
+	return &wdk.MerklePathResult{
+		Name:       ServiceName,
+		MerklePath: merklePath,
+		Header:     header,
+	}, nil
+}
+
+func (woc *WhatsOnChain) hashToHeader(ctx context.Context, blockHash string) (*wdk.BlockHeader, error) {
+	var hdrResp blockHeaderResponse
+	res, err := woc.httpClient.R().
+		SetContext(ctx).
+		SetResult(&hdrResp).
+		AddRetryCondition(retryOnTooManyRequestsStatus).
+		Get(fmt.Sprintf("%s/block/%s/header", woc.url, blockHash))
+
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch block header: %w", err)
+	}
+	if res.StatusCode() != http.StatusOK {
+		return nil, fmt.Errorf("unexpected status code %d fetching block header", res.StatusCode())
+	}
+
+	height, err := to.UInt32(hdrResp.Height)
+	if err != nil {
+		return nil, fmt.Errorf("invalid block height %d: %w", hdrResp.Height, err)
+	}
+
+	return &wdk.BlockHeader{
+		Height:     height,
+		Hash:       blockHash,
+		MerkleRoot: hdrResp.MerkleRoot,
+	}, nil
+}
+
+// getTscProof retrieves the TSC proof from WoC.
+func (woc *WhatsOnChain) getTscProof(ctx context.Context, txID string) (*tscProof, error) {
+	var proof tscProof
+	req := woc.httpClient.R().
+		SetContext(ctx).
+		SetResult(&proof).
+		AddRetryCondition(retryOnTooManyRequestsStatus)
+
+	res, err := req.Get(fmt.Sprintf("%s/tx/%s/proof/tsc", woc.url, txID))
+	if err != nil {
+		return nil, fmt.Errorf("failed to query TSC proof: %w", err)
+	}
+	if res.StatusCode() == http.StatusNotFound {
+		return nil, nil // No proof found
+	}
+	if res.StatusCode() != http.StatusOK {
+		return nil, fmt.Errorf("unexpected status code %d fetching TSC proof", res.StatusCode())
+	}
+
+	return &proof, nil
+}
+
+func convertTscProofToMerklePath(txid string, index int, nodes []string, blockHeight uint32) (*transaction.MerklePath, error) {
+	if len(nodes) == 0 {
+		return nil, fmt.Errorf("no nodes provided in TSC proof for txid %s", txid)
+	}
+
+	txidHash, err := chainhash.NewHashFromHex(txid)
+	if err != nil {
+		return nil, fmt.Errorf("invalid txid: %w", err)
+	}
+
+	level0, nextIndex, err := buildLevel0PathElement(txid, txidHash, nodes[0], index)
+	if err != nil {
+		return nil, fmt.Errorf("failed to build level 0 path element: %w", err)
+	}
+
+	upperLevels, err := buildUpperLevels(nodes, 1, nextIndex)
+	if err != nil {
+		return nil, fmt.Errorf("failed to build upper levels: %w", err)
+	}
+
+	treeHeight := len(nodes)
+	path := make([][]*transaction.PathElement, treeHeight)
+	path[0] = level0
+	for i := 1; i < treeHeight; i++ {
+		path[i] = upperLevels[i]
+	}
+
+	return transaction.NewMerklePath(blockHeight, path), nil
+}
+
+func buildLevel0PathElement(txid string, txidHash *chainhash.Hash, node string, index int) ([]*transaction.PathElement, int, error) {
+	isOdd := index%2 == 1
+	siblingIndex := index ^ 1
+
+	sibling, err := createPathElement(node, siblingIndex, true, txid)
+	if err != nil {
+		return nil, 0, fmt.Errorf("invalid node hash at level 0: %w", err)
+	}
+
+	txidLeaf := &transaction.PathElement{
+		Offset: must.ConvertToUInt64(index),
+		Hash:   txidHash,
+		Txid:   to.Ptr(true),
+	}
+
+	var level0 []*transaction.PathElement
+	if isOdd {
+		level0 = []*transaction.PathElement{sibling, txidLeaf}
+	} else {
+		level0 = []*transaction.PathElement{txidLeaf, sibling}
+	}
+
+	nextIndex := index >> 1
+	return level0, nextIndex, nil
+}
+
+func buildUpperLevels(nodes []string, startLevel int, startIndex int) ([][]*transaction.PathElement, error) {
+	treeHeight := len(nodes)
+	path := make([][]*transaction.PathElement, treeHeight)
+
+	currentIndex := startIndex
+
+	for level := startLevel; level < treeHeight; level++ {
+		siblingIndex := currentIndex ^ 1
+
+		sibling, err := createPathElement(nodes[level], siblingIndex, false, "")
+		if err != nil {
+			return nil, fmt.Errorf("invalid node hash at level %d: %w", level, err)
+		}
+
+		path[level] = []*transaction.PathElement{sibling}
+		currentIndex >>= 1
+	}
+
+	return path, nil
+}
+
+// createPathElement builds a PathElement given node string and sibling index.
+func createPathElement(node string, siblingIndex int, isLevel0 bool, txid string) (*transaction.PathElement, error) {
+	const duplicateNodeMarker = "*"
+
+	element := &transaction.PathElement{
+		Offset: must.ConvertToUInt64(siblingIndex),
+	}
+
+	if node == duplicateNodeMarker || (isLevel0 && node == txid) {
+		element.Duplicate = to.Ptr(true)
+	} else {
+		nodeHash, err := chainhash.NewHashFromHex(node)
+		if err != nil {
+			return nil, fmt.Errorf("invalid node hash %q: %w", node, err)
+		}
+		element.Hash = nodeHash
+	}
+
+	return element, nil
+}
+
+// retryOnTooManyRequestsStatus defines retry condition for Too Many Requests response.
+func retryOnTooManyRequestsStatus(res *resty.Response, err error) bool {
+	return res.StatusCode() == http.StatusTooManyRequests
+}
--- a/pkg/services/internal/whatsonchain/service.go
+++ b/pkg/services/internal/whatsonchain/service.go
@@ -62,7 +62,7 @@ func New(httpClient *resty.Client, logger *slog.Logger, network defs.BSVNetwork,
 		url:               fmt.Sprintf("https://api.whatsonchain.com/v1/bsv/%s", network),
 		logger:            logger,
 		bsvExchangeRate:   config.BSVExchangeRate,
-		bsvUpdateInterval: to.IfThen(config.BSVUpdateInterval != nil, *config.BSVUpdateInterval).ElseThen(defs.DefaultBSVExchangeUpdateInterval),
+		bsvUpdateInterval: to.If(config.BSVUpdateInterval != nil, func() time.Duration { return *config.BSVUpdateInterval }).ElseThen(defs.DefaultBSVExchangeUpdateInterval),
 	}
 }
 
@@ -70,9 +70,7 @@ func (woc *WhatsOnChain) RawTx(ctx context.Context, txID string) (*wdk.RawTxResu
 	req := woc.httpClient.
 		R().
 		SetContext(ctx).
-		AddRetryCondition(func(res *resty.Response, err error) bool {
-			return res.StatusCode() == http.StatusTooManyRequests
-		})
+		AddRetryCondition(retryOnTooManyRequestsStatus)
 	req.SetHeader("Cache-Control", "no-cache")
 
 	res, err := req.Get(fmt.Sprintf("%s/tx/%s/hex", woc.url, txID))
@@ -114,9 +112,7 @@ func (woc *WhatsOnChain) UpdateBsvExchangeRate() (defs.BSVExchangeRate, error) {
 	var exchangeRateResponse bsvExchangeRateResponse
 	req := woc.httpClient.
 		R().
-		AddRetryCondition(func(res *resty.Response, err error) bool {
-			return res.StatusCode() == http.StatusTooManyRequests
-		})
+		AddRetryCondition(retryOnTooManyRequestsStatus)
 
 	res, err := req.
 		SetResult(&exchangeRateResponse).
--- a/pkg/services/internal/whatsonchain/testabilities/constants.go
+++ b/pkg/services/internal/whatsonchain/testabilities/constants.go
@@ -1,0 +1,8 @@
+package testabilities
+
+const (
+	BlockHeight     = 54321
+	TestTxID        = "11223344556677889900aabbccddeeff00112233445566778899aabbccddeeff"
+	TestTargetHash  = "deadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeef"
+	TestSiblingHash = "29eab26928d7721042114c47f15735877d290cd00773e8be44bcc3dc7b541c02"
+)
--- a/pkg/services/internal/whatsonchain/testabilities/fixtures.go
+++ b/pkg/services/internal/whatsonchain/testabilities/fixtures.go
@@ -1,0 +1,42 @@
+package testabilities
+
+import (
+	"testing"
+
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/defs"
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/internal/logging"
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/internal/testabilities/testservices"
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/services/internal/whatsonchain"
+	"github.com/go-softwarelab/common/pkg/to"
+)
+
+type WoCServiceFixture interface {
+	testservices.ServicesFixture
+	NewWoCService() *whatsonchain.WhatsOnChain
+}
+
+type wocServiceFixture struct {
+	testservices.ServicesFixture
+	t testing.TB
+}
+
+func Given(t testing.TB) WoCServiceFixture {
+	return &wocServiceFixture{
+		ServicesFixture: testservices.GivenServices(t),
+		t:               t,
+	}
+}
+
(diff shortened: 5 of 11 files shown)


pkg/services/internal/whatsonchain/service.go
```diff
@@ -62,7 +62,7 @@ func New(httpClient *resty.Client, logger *slog.Logger, network defs.BSVNetwork,
 		url:               fmt.Sprintf("https://api.whatsonchain.com/v1/bsv/%s", network),
 		logger:            logger,
 		bsvExchangeRate:   config.BSVExchangeRate,
-		bsvUpdateInterval: to.IfThen(config.BSVUpdateInterval != nil, *config.BSVUpdateInterval).ElseThen(defs.DefaultBSVExchangeUpdateInterval),
+		bsvUpdateInterval: to.If(config.BSVUpdateInterval != nil, func() time.Duration { return *config.BSVUpdateInterval }).ElseThen(defs.DefaultBSVExchangeUpdateInterval),
```
dorzepowski, line 65: check `to.ValueOr(config.BSVUpdateInterval, defs.DefaultBSVExchangeUpdateInterval)` as an alternative

pkg/services/services.go
```diff
@@ -67,6 +67,7 @@ func New(logger *slog.Logger, config defs.WalletServices, opts ...func(*options.
 			logger,
 			"MerklePath",
 			servicequeue.NewService1(whatsonchain.ServiceName, arcService.MerklePath),
```
dorzepowski: It was probably my mistake, but I just spot it, can you please 🥺  change it:
```suggestion
			servicequeue.NewService1(arc.ServiceName, arcService.MerklePath),
```

pkg/services/internal/whatsonchain/service.go
```diff
@@ -141,3 +141,7 @@ func (woc *WhatsOnChain) UpdateBsvExchangeRate() (defs.BSVExchangeRate, error) {
 
 	return woc.bsvExchangeRate, nil
 }
+
+func (w *WhatsOnChain) SetBaseURL(url string) {
```
chris-4chain: Please use `woc` as a receiver to keep it consistent.

pkg/services/internal/whatsonchain/woc_get_merkle_path.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+	if merkleRoot != hdrResp.MerkleRoot {
+		return nil, fmt.Errorf("computed merkle root %q does not match block header %q", merkleRoot, hdrResp.MerkleRoot)
+	}
+
+	return &wdk.MerklePathResult{
+		Name:       ServiceName,
+		MerklePath: merklePath,
+		Header: &wdk.BlockHeader{
+			Height:     must.ConvertToUInt32(hdrResp.Height),
```
chris-4chain: Since `hdrResp.Height` is from a response from 3rd party service I would suggest to use `height, err := to.ConvertUint32()` instead of `must` that panics on error.

pkg/services/internal/whatsonchain/testabilities/constants.go
```diff
@@ -0,0 +1,8 @@
+package testabilities
+
+const (
+	TestTxID       = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
```
chris-4chain: Please use some more "standard-looking" hash.

pkg/services/internal/whatsonchain/service_test.go
```diff
@@ -0,0 +1,204 @@
...
+
+func TestMerklePath_Success(t *testing.T) {
+	// Given
+	given := tst.Given(t)
+	txID := tst.TestTxID
+	targetHash := tst.TestTargetHash
+	proof := map[string]any{
+		"height": 54321,
+		"index":  0,
+		"nodes":  []string{"aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"},
```
chris-4chain: Please use some `contant` that would help to understand what this "hash" represents. You could use one of already defined constants.

pkg/services/internal/whatsonchain/service.go
```diff
@@ -62,7 +62,7 @@ func New(httpClient *resty.Client, logger *slog.Logger, network defs.BSVNetwork,
 		url:               fmt.Sprintf("https://api.whatsonchain.com/v1/bsv/%s", network),
```
chris-4chain, line 62: I think that you've made some redundant logic (I mean `SetBaseURL` method), because the url is already set here.

pkg/services/internal/whatsonchain/woc_get_merkle_path.go
```diff
@@ -0,0 +1,161 @@
...
+	Height     int    `json:"height"`
+	MerkleRoot string `json:"merkleroot"`
+}
+
+// MerklePath retrieves the merkle path for a transaction using WoC TSC proof.
+func (woc *WhatsOnChain) MerklePath(ctx context.Context, txID string) (*wdk.MerklePathResult, error) {
+	req := woc.httpClient.R().
+		SetContext(ctx).
+		SetHeader("Accept", "application/json").
+		AddRetryCondition(retryOn429)
```
chris-4chain:
```suggestion
	var proof tscProof
	req := woc.httpClient.R().
		SetContext(ctx).
		SetResult(&proof).
		SetHeader("Accept", "application/json").
		AddRetryCondition(retryOn429)
```

This way you don't have to unmarshall the response object manually.

pkg/services/internal/whatsonchain/woc_get_merkle_path.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+	if hdrRes.StatusCode() != http.StatusOK {
+		return nil, fmt.Errorf("unexpected status code %d fetching block header", hdrRes.StatusCode())
+	}
+	if len(hdrRes.Body()) == 0 {
+		return nil, errors.New("empty response body fetching block header")
+	}
+
+	var hdrResp blockHeaderResponse
+	if err := json.Unmarshal(hdrRes.Body(), &hdrResp); err != nil {
```
chris-4chain: Same comment about `SetResult` that automatically unmarshalls the response object.

pkg/services/internal/whatsonchain/woc_get_merkle_path.go
```diff
@@ -0,0 +1,161 @@
...
+		Header: &wdk.BlockHeader{
+			Height:     must.ConvertToUInt32(hdrResp.Height),
+			Hash:       proof.Target,
+			MerkleRoot: merkleRoot,
+		},
+	}, nil
+}
+
+// convertTscProofToMerklePath converts WoC TSC proof format into internal MerklePath structure.
+func convertTscProofToMerklePath(txid string, index int, nodes []string, blockHeight uint32) (*transaction.MerklePath, error) {
```
chris-4chain: wow - do we need such complex algorithm to do the job? :thinking: I'm surprised.

pkg/services/internal/whatsonchain/service_test.go
```diff
@@ -0,0 +1,204 @@
+package whatsonchain_test
```
chris-4chain: Service tests should be conducted from the Public Facade to ensure that they accurately reflect the user's perspective. 
- This approach is consistent with the principle of not testing private functions directly.
- See where are tests of other "(sub)services" already. 

However, there is one "challenge" that needs to be resolved - because of the services are in `ServiceQueues` we need to figure out how to disable other subservices and make that only the tested one responses. 

image

pkg/services/internal/whatsonchain/testabilities/fixtures.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+	fixture.SetupHandler(http.NotFoundHandler())
+	return fixture
+}
+
+func (f *wocServiceFixture) SetupHandler(h http.Handler) {
+	if f.server != nil {
+		f.server.Close()
+	}
+	f.server = httptest.NewServer(h)
```
dorzepowski: We're using httpmock and mocktransport
see how it's done in [arc fixture](https://github.com/bsv-blockchain/go-wallet-toolbox/blob/a13db1efb01897c92d2768d8f4c9c1589265ddb2/pkg/internal/testabilities/testservices/fixture_arc.go#L71-L70)
This way we don't need to override the baseURL in a service.

pkg/services/internal/whatsonchain/service.go
dorzepowski: Please remove this method. It's just for tests, and when httpmock instead httptest server will be used, this method won't be needed at all.

pkg/services/internal/whatsonchain/service_test.go
```diff
@@ -0,0 +1,204 @@
...
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/internal/txutils"
+	tst "github.com/4chain-ag/go-wallet-toolbox/pkg/services/internal/whatsonchain/testabilities"
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/wdk"
+	"github.com/bsv-blockchain/go-sdk/transaction"
+	txtestabilities "github.com/bsv-blockchain/universal-test-vectors/pkg/testabilities"
+	"github.com/go-softwarelab/common/pkg/to"
+	"github.com/stretchr/testify/require"
+)
+
+func TestRawTx(t *testing.T) {
```
dorzepowski: there are tests for most of those methods on Services level that use already some testability, maybe you could make it the same way here.

See:
- [services_rawtx_test.go](https://github.com/bsv-blockchain/go-wallet-toolbox/blob/6522144f810ee4166ae9c3bd17455f70fc805d87/pkg/services/services_rawtx_test.go)
- [services_bsv_exchange_rate_test.go](https://github.com/bsv-blockchain/go-wallet-toolbox/blob/6522144f810ee4166ae9c3bd17455f70fc805d87/pkg/services/services_bsv_exchange_rate_test.go)

pkg/services/internal/whatsonchain/woc_get_merkle_path.go
```diff
@@ -0,0 +1,161 @@
...
+type blockHeaderResponse struct {
+	Height     int    `json:"height"`
+	MerkleRoot string `json:"merkleroot"`
+}
+
+// MerklePath retrieves the merkle path for a transaction using WoC TSC proof.
+func (woc *WhatsOnChain) MerklePath(ctx context.Context, txID string) (*wdk.MerklePathResult, error) {
+	req := woc.httpClient.R().
+		SetContext(ctx).
+		SetHeader("Accept", "application/json").
```
dorzepowski: this header is already set on woc.httpClient in whatsonchain.New() function

pkg/services/internal/whatsonchain/woc_get_merkle_path.go
dorzepowski: NICE, 
can you use retryOn429 in RawTx and UpdateBsvExchangeRate.

pkg/services/internal/whatsonchain/woc_get_merkle_path.go
```diff
@@ -0,0 +1,161 @@
...
+		}
+
+		index >>= 1
+	}
+
+	return transaction.NewMerklePath(blockHeight, path), nil
+}
+
+// retryOn429 defines retry condition for Too Many Requests response.
+func retryOn429(res *resty.Response, err error) bool {
```
dorzepowski: maybe `retryOnTooManyRequestsStatus`

pkg/services/internal/whatsonchain/woc_get_merkle_path.go
```diff
@@ -0,0 +1,161 @@
...
+		return &wdk.MerklePathResult{Name: ServiceName}, nil
+	}
+	if res.StatusCode() != http.StatusOK {
+		return nil, fmt.Errorf("unexpected status code %d fetching TSC proof", res.StatusCode())
+	}
+	if len(res.Body()) == 0 {
+		return nil, errors.New("empty response body fetching TSC proof")
+	}
+
+	var proof tscProof
```
dorzepowski: as suggested by Krzysiek, use resty to unmarshal the body for you 
see: `SetResult(&proof)`

pkg/services/internal/whatsonchain/woc_get_merkle_path.go
```diff
@@ -0,0 +1,161 @@
...
+	if err := json.Unmarshal(res.Body(), &proof); err != nil {
+		return nil, fmt.Errorf("failed to parse TSC proof: %w", err)
+	}
+
+	hdrRes, err := woc.httpClient.R().
+		SetContext(ctx).
+		SetHeader("Accept", "application/json").
```
dorzepowski: already set when creating a service

pkg/services/internal/whatsonchain/woc_get_merkle_path.go
```diff
@@ -0,0 +1,161 @@
...
+	if len(res.Body()) == 0 {
+		return nil, errors.New("empty response body fetching TSC proof")
+	}
+
+	var proof tscProof
+	if err := json.Unmarshal(res.Body(), &proof); err != nil {
+		return nil, fmt.Errorf("failed to parse TSC proof: %w", err)
+	}
+
+	hdrRes, err := woc.httpClient.R().
```
dorzepowski: It would be good to extract 2 methods for 
1. requesting for proof  
2. requesting for header
So the logic here in MerklePath would be less bloated with the request handling

pkg/services/internal/whatsonchain/woc_get_merkle_path.go
```diff
@@ -0,0 +1,161 @@
...
+	if len(hdrRes.Body()) == 0 {
+		return nil, errors.New("empty response body fetching block header")
+	}
+
+	var hdrResp blockHeaderResponse
+	if err := json.Unmarshal(hdrRes.Body(), &hdrResp); err != nil {
+		return nil, fmt.Errorf("failed to parse block header response: %w", err)
+	}
+
+	merklePath, err := convertTscProofToMerklePath(txID, proof.Index, proof.Nodes, must.ConvertToUInt32(hdrResp.Height))
```
dorzepowski: use `to.UInt32` and handle error if Height is not valid UInt32, 
`must.ConvertToUInt32` would panic, which isn't what we want here - we should use that function only if we're sure that the panic shouldn't ever happen because we're the provider of that value or we already made a validation on the value.

pkg/services/internal/whatsonchain/woc_get_merkle_path.go
```diff
@@ -0,0 +1,161 @@
...
+	if merkleRoot != hdrResp.MerkleRoot {
+		return nil, fmt.Errorf("computed merkle root %q does not match block header %q", merkleRoot, hdrResp.MerkleRoot)
+	}
+
+	return &wdk.MerklePathResult{
+		Name:       ServiceName,
+		MerklePath: merklePath,
+		Header: &wdk.BlockHeader{
+			Height:     must.ConvertToUInt32(hdrResp.Height),
+			Hash:       proof.Target,
```
dorzepowski: maybe we should either compare it to header hash (so we should take hash also when getting the header) or just put a heade hash here.

pkg/services/internal/whatsonchain/woc_get_merkle_path.go
dorzepowski: yes, because merkle proof from WoC is in TSC format, not in BUMP format 🙁

pkg/services/internal/whatsonchain/service.go
ac4ch: this was giving nil pointer `bsvUpdateInterval: to.IfThen(config.BSVUpdateInterval != nil, *config.BSVUpdateInterval).ElseThen(defs.DefaultBSVExchangeUpdateInterval),` current change i applied using same lib works as expected

pkg/services/internal/whatsonchain/service_test.go
ac4ch: look under other components like arc httpx .. so i followed same approach as in other subpackages

pkg/services/internal/whatsonchain/service.go
ac4ch: well dont think so .. i added this set url as here u have api.whatsonchain.com hard-coded .. and in test i was hard to mock it ... but maybe with httpmock instead httptest will be handled as it should

pkg/services/internal/whatsonchain/service_test.go
ac4ch: ok i can put this under const .. but just to let u know this is valid hash ...

pkg/services/internal/whatsonchain/service_test.go
ac4ch: so lets have chat about, if other where also wrong and u wanna start the "correct pattern with me" :)

pkg/services/internal/whatsonchain/woc_get_merkle_path.go
ac4ch: that what code 429 stand for ... but sure i can change the name 
`HTTP/1.1 429 Too Many Requests`

pkg/services/internal/whatsonchain/woc_get_merkle_path.go
nonus25: not sure if iam getting what u mean exactly here will need more explanation :)

pkg/services/service_woc_merkle_path_test.go
```diff
@@ -0,0 +1,204 @@
...
+
+func TestMerklePath_Success(t *testing.T) {
+	// Given
+	given := tst.Given(t)
+	txID := tst.TestTxID
+	targetHash := tst.TestTargetHash
+	proof := map[string]any{
+		"height": 54321,
+		"index":  0,
+		"nodes":  []string{"aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"},
```
chris-4chain: Please use the constant here ☝️

pkg/services/internal/whatsonchain/woc_get_merkle_path.go
```diff
@@ -0,0 +1,167 @@
...
+	if res.StatusCode() != http.StatusOK {
+		return nil, fmt.Errorf("unexpected status code %d fetching TSC proof", res.StatusCode())
+	}
+	if len(res.Body()) == 0 {
+		return nil, errors.New("empty response body fetching TSC proof")
+	}
+
+	if err := json.Unmarshal(res.Body(), &proof); err != nil {
+		return nil, fmt.Errorf("failed to parse TSC proof: %w", err)
+	}
```
chris-4chain: You applied our suggestion but didn't remove manual json parsing
```suggestion
	var proof tscProof
	req := woc.httpClient.R().
		SetContext(ctx).
		SetResult(&proof).
		AddRetryCondition(retryOnTooManyRequestsStatus)

	res, err := req.Get(fmt.Sprintf("%s/tx/%s/proof/tsc", woc.url, txID))
	if err != nil {
		return nil, fmt.Errorf("failed to query TSC proof: %w", err)
	}
	if res.StatusCode() == http.StatusNotFound {
		return &wdk.MerklePathResult{Name: ServiceName}, nil
	}
	if res.StatusCode() != http.StatusOK {
		return nil, fmt.Errorf("unexpected status code %d fetching TSC proof", res.StatusCode())
	}
```

pkg/services/internal/whatsonchain/woc_get_merkle_path.go
chris-4chain: The same here ☝️. The `SetResult` option makes that the `hdrResp` will be automatically filled after the request - so, no need to `json.Unmarshal` the body manually.

pkg/services/internal/whatsonchain/woc_get_merkle_path.go
```diff
@@ -0,0 +1,167 @@
...
+			MerkleRoot: merkleRoot,
+		},
+	}, nil
+}
+
+// convertTscProofToMerklePath converts WoC TSC proof format into internal MerklePath structure.
+func convertTscProofToMerklePath(txid string, index int, nodes []string, blockHeight uint32) (*transaction.MerklePath, error) {
+	const duplicateNodeMarker = "*"
+	if len(nodes) == 0 {
+		return nil, errors.New("no merkle proof nodes provided")
```
chris-4chain:
```suggestion
		return nil, fmt.Errorf("no merkle proof nodes provided")
```

pkg/services/internal/whatsonchain/woc_get_merkle_path.go
```diff
@@ -0,0 +1,167 @@
...
+
+	treeHeight := len(nodes)
+	path := make([][]*transaction.PathElement, treeHeight)
+
+	txidHash, err := chainhash.NewHashFromHex(txid)
+	if err != nil {
+		return nil, fmt.Errorf("invalid txid: %w", err)
+	}
+
+	for level := 0; level < treeHeight; level++ {
```
chris-4chain:
```suggestion
	for level := range treeHeight {
```

pkg/services/internal/whatsonchain/woc_get_merkle_path.go
```diff
@@ -0,0 +1,167 @@
...
+		siblingOffset := must.ConvertToUInt64(siblingIndex)
+
+		leaf := &transaction.PathElement{
+			Offset: siblingOffset,
+		}
+
+		node := nodes[level]
+		if node == duplicateNodeMarker {
+			dup := true
+			leaf.Duplicate = &dup
```
chris-4chain:
```suggestion
			leaf.Duplicate = to.Ptr(true)
```

pkg/services/internal/whatsonchain/woc_get_merkle_path.go
```diff
@@ -0,0 +1,167 @@
...
+		isOdd := index%2 == 1
+		siblingIndex := index ^ 1
+		siblingOffset := must.ConvertToUInt64(siblingIndex)
+
+		leaf := &transaction.PathElement{
+			Offset: siblingOffset,
+		}
+
+		node := nodes[level]
+		if node == duplicateNodeMarker {
```
chris-4chain: Don't you forget the "second" component in `if`?
In TS this looks like this:
```js
if (node === '*' || (level === 0 && node === txid)) {
      leaf.duplicate = true
    } else {
      leaf.hash = node
    }
```

pkg/services/internal/whatsonchain/woc_get_merkle_path.go
```diff
@@ -0,0 +1,167 @@
...
+
+		path[level] = []*transaction.PathElement{leaf}
+
+		if level == 0 {
+			txidFlag := true
+			txidLeaf := &transaction.PathElement{
+				Offset: must.ConvertToUInt64(index),
+				Hash:   txidHash,
+				Txid:   &txidFlag,
+			}
```
chris-4chain:
```suggestion
			txidLeaf := &transaction.PathElement{
				Offset: must.ConvertToUInt64(index),
				Hash:   txidHash,
				Txid:   to.Ptr(true),
			}
```

pkg/services/internal/whatsonchain/woc_get_merkle_path.go
```diff
@@ -0,0 +1,167 @@
...
+			if isOdd {
```
chris-4chain: Because of the `isOdd` is used only here - I would suggest to move the var definition, here, to the `if level == 0` scope.

pkg/services/internal/whatsonchain/service.go
chris-4chain: Use  httpmock instead and this solves your problem. 
- And please remove the `SetBaseURL` method.

pkg/services/internal/whatsonchain/service_test.go
chris-4chain: Maybe you were right. Sorry for that - If for ARC was such approach, you did right. 
But you said we need to chat about this, however you already moved the tests in a way that this has less sense. 

What I meant was that, in file `services/services_get_merkle_path_test.go` to do similar test as it is for ARC:
https://github.com/bsv-blockchain/go-wallet-toolbox/blob/a13db1efb01897c92d2768d8f4c9c1589265ddb2/pkg/services/services_get_merkle_path_test.go#L64-L111
This way we make sure that this actually works from the "public facade" perspective. 

So, What I'm suggesting now: 
- Move the tests you made, back where they were. 
- Add additional test into `services_get_merkle_path_test.go` "get merkle path from WoC" - similar to the menioned one ☝️

chris-4chain (changes requested):

pkg/services/internal/whatsonchain/service_test.go
ac4ch: now i agree

pkg/services/internal/whatsonchain/woc_get_merkle_path.go
ac4ch: thx i even did not notes that i used errors new here ...

pkg/services/internal/whatsonchain/woc_get_merkle_path.go
ac4ch: yes new golang way ... as before u had to do the for loop in old fashion

pkg/services/internal/whatsonchain/merkle_path.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+	if res.StatusCode() == http.StatusNotFound {
+		return &wdk.MerklePathResult{Name: ServiceName}, nil
+	}
+	if res.StatusCode() != http.StatusOK {
+		return nil, fmt.Errorf("unexpected status code %d fetching TSC proof", res.StatusCode())
+	}
+	if len(res.Body()) == 0 {
+		return nil, errors.New("empty response body fetching TSC proof")
+	}
```
chris-4chain: I'm not sure if the body-len-check is necessary. But, maybe it's good 🤷

pkg/services/internal/whatsonchain/merkle_path.go
```diff
@@ -0,0 +1,158 @@
...
+		path[level] = []*transaction.PathElement{leaf}
+
+		if level == 0 {
+			txidFlag := true
+			txidLeaf := &transaction.PathElement{
+				Offset: must.ConvertToUInt64(index),
+				Hash:   txidHash,
+				Txid:   &txidFlag,
+			}
+			if isOdd {
```
chris-4chain: What I meant by moving the `isOdd` value is to calculate this variable only if necessary - and place is closer to actual usage.
image

pkg/services/internal/whatsonchain/woc_get_merkle_path.go
ac4ch: i refactored this to separate helper functions .. should be more readable

pkg/services/internal/whatsonchain/woc_get_merkle_path.go
ac4ch: ok i see what u saying here ... generally we passing unsafe hash without of verification i will try to address this, thx

ac4ch: added logic like in TS `hashToHeader`

pkg/services/services_get_merkle_path_test.go
```diff
@@ -110,4 +112,83 @@ func TestGetMerklePath(t *testing.T) {
...
+			MerklePath: &merklePath,
+			Header: &wdk.BlockHeader{
+				Height:     tst.BlockHeight,
+				Hash:       tst.TestTargetHash,
+				MerkleRoot: merkleRoot,
+			},
+		}, *response)
+	})
+
+	t.Run("return result without Merkle Path when transaction is not mined yet WoC", func(t *testing.T) {
```
dorzepowski: there is already test for that, but it hasn't setup for WoC.
IMHO single test with all services set up would be enough.

pkg/services/internal/whatsonchain/merkle_path copy.gox
dorzepowski, line 1: whats that? (looks like something committed by accident)

pkg/services/internal/whatsonchain/merkle_path.go
```diff
@@ -0,0 +1,216 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("invalid txid: %w", err)
+	}
+
+	level0, nextIndex, err := buildLevel0PathElement(txid, txidHash, nodes[0], index)
+	if err != nil {
+		return nil, fmt.Errorf("failed to build level 0 path element: %w", err)
+	}
+
+	upperLevels, err := buildUpperLevels(nodes, 1, nextIndex)
```
dorzepowski, line 130: No it looks much better.

pkg/services/internal/whatsonchain/merkle_path.go
```diff
@@ -0,0 +1,216 @@
...
+	for i := 1; i < treeHeight; i++ {
+		path[i] = upperLevels[i]
+	}
+
+	return transaction.NewMerklePath(blockHeight, path), nil
+}
+
+func buildLevel0PathElement(txid string, txidHash *chainhash.Hash, node string, index int) ([]*transaction.PathElement, int, error) {
+	isOdd := index%2 == 1
+	siblingIndex := index ^ 1
```
dorzepowski, line 147: I forgot to say previously:
Awesome work with those bitwise operators 👏

pkg/services/internal/whatsonchain/testabilities/constants.go
```diff
@@ -0,0 +1,9 @@
+package testabilities
+
+const (
+	BlockHeight     = 54321
+	TestTxID        = "11223344556677889900aabbccddeeff00112233445566778899aabbccddeeff"
+	TestTargetHash  = "deadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeef"
+	TestSiblingHash = "29eab26928d7721042114c47f15735877d290cd00773e8be44bcc3dc7b541c02"
+	TestRawTxHex    = "0100000001abcdef"
```
dorzepowski: this looks like not used.

pkg/internal/testabilities/testservices/fixture_woc.go
```diff
@@ -22,6 +41,16 @@ type wocFixture struct {
 	network   defs.BSVNetwork
 }
 
+type wocMerklePathQueryFixture struct {
```
dorzepowski: it looks like a time to move those additional fixtures to their own files.

pkg/internal/testabilities/testservices/fixture_woc.go
```diff
@@ -36,32 +65,91 @@ func NewWoCFixture(t testing.TB, opts ...Option) WhatsOnChainFixture {
...
+			resp.Header.Set("Content-Type", "text/plain")
+			return resp, nil
 		}
 	}
-
 	url := fmt.Sprintf("https://api.whatsonchain.com/v1/bsv/%s/tx/%s/hex", f.network, txID)
 	f.transport.RegisterResponder("GET", url, responder(status, rawTx))
 }
+
+func (f *wocFixture) WillRespondWithMerklePath(status int, txID, responseBody string, err error) {
```
dorzepowski: the in args is not needed:
1. because you don't pass it in any tests
2. because it's making this method hard to understand, how to use it, and what to pass
3. Michał G. make a PR that is adding WillBeUnreachable that actually is better simulating the case when httpclient would get an error: see [WillBeUnreachable in PR 280](https://github.com/bsv-blockchain/go-wallet-toolbox/pull/280/files#diff-0a498ce2b803514ad0826deb0f2d90e0d7664b8166232bf949179dda7f93cac9R87)

pkg/internal/testabilities/testservices/fixture_woc.go
```diff
@@ -36,32 +65,91 @@ func NewWoCFixture(t testing.TB, opts ...Option) WhatsOnChainFixture {
...
+		}
+		resp := httpmock.NewStringResponse(status, responseBody)
+		resp.Header.Set("Content-Type", "application/json")
+		return resp, nil
+	}
+	url := fmt.Sprintf("https://api.whatsonchain.com/v1/bsv/%s/tx/%s/proof/tsc", f.network, txID)
+	f.transport.RegisterResponder("GET", url, responder)
+}
+
+func (f *wocFixture) WillRespondWithBlockHeader(status int, blockHash, responseBody string, err error) {
```
dorzepowski: same as in WillRespondWithMerklePath - no need for err argument.

pkg/internal/testabilities/testservices/fixture_woc.go
```diff
@@ -36,32 +65,91 @@ func NewWoCFixture(t testing.TB, opts ...Option) WhatsOnChainFixture {
...
+func (f *wocFixture) WhenQueryingMerklePath(txID string) WhatsOnChainMerklePathQueryFixture {
+	return &wocMerklePathQueryFixture{fixture: f, txID: txID}
+}
+
+func (f *wocFixture) WhenQueryingBlockHeader(blockHash string) WhatsOnChainBlockHeaderQueryFixture {
+	return &wocBlockHeaderQueryFixture{fixture: f, blockHash: blockHash}
+}
+
+func (q *wocMerklePathQueryFixture) WillReturnMerklePathResponse(status int, responseBody string, err error) {
+	q.fixture.WillRespondWithMerklePath(status, q.txID, responseBody, err)
```
dorzepowski: IMHO the body of WillRespondWithMerklePath should be here instead of calling parent fixture

pkg/internal/testabilities/testservices/fixture_woc.go
```diff
@@ -36,32 +65,91 @@ func NewWoCFixture(t testing.TB, opts ...Option) WhatsOnChainFixture {
...
+}
+
+func (q *wocMerklePathQueryFixture) WillReturnTSCProof(status int, tscProofJSON string) {
+	q.fixture.WillRespondWithMerklePath(status, q.txID, tscProofJSON, nil)
+}
+
+func (q *wocBlockHeaderQueryFixture) WillReturnBlockHeaderResponse(status int, responseBody string, err error) {
+	q.fixture.WillRespondWithBlockHeader(status, q.blockHash, responseBody, err)
```
dorzepowski: IMHO the body of WillRespondWithBlockHeader should be here instead of calling parent fixture

pkg/services/internal/whatsonchain/merkle_path.go
ac4ch: u mean `Now` correct ?

pkg/services/internal/whatsonchain/merkle_path copy.gox
ac4ch: yes i did  copy of this file while was working on refactoring it

pkg/services/services_get_merkle_path_test.go
ac4ch: ok ya got it .. and removed it

dorzepowski (approved):

chris-4chain (approved):

PR #275, opened 2025-06-11
feat(#274): create BRC29 script template

dorzepowski:


pkg/internal/sdk/template/brc29/brc29_address.go
```diff
@@ -0,0 +1,52 @@
...
+		opt(options)
+	}
+
+	if err := keyID.Validate(); err != nil {
+		return nil, fmt.Errorf("invalid key ID: %w", err)
+	}
+
+	senderKeyDeriver, err := toKeyDeriver(sender)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create sender key deriver: %w", err)
```
ac4ch:
```suggestion
		return nil, fmt.Errorf("failed to create sender key deriver (input type %T): %w", sender, err)
```
if this can fail due to invalid type would be helpful to log the type WDYT ?

pkg/internal/sdk/template/brc29/brc29_address.go
```diff
@@ -0,0 +1,52 @@
...
+
+	recipientIdentityKey, err := toIdentityKey(recipient)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create recipient identity key: %w", err)
+	}
+
+	key, err := senderKeyDeriver.DerivePublicKey(Protocol, keyID.String(), sdk.Counterparty{
+		Type:         sdk.CounterpartyTypeOther,
+		Counterparty: recipientIdentityKey,
+	}, false)
```
ac4ch: consider adding a named constant if `false` is always passed ex. `isHardened := false`

pkg/internal/sdk/template/brc29/brc29_address_test.go
```diff
@@ -0,0 +1,152 @@
...
+
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/internal/sdk"
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/internal/sdk/template/brc29"
+	ec "github.com/bsv-blockchain/go-sdk/primitives/ec"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestAddress(t *testing.T) {
+	errorTestCases := map[string]struct {
```
ac4ch, line 94:
```suggestion
        // Given
	errorTestCases := map[string]struct {
```

pkg/internal/sdk/template/brc29/brc29_address_test.go
```diff
@@ -0,0 +1,152 @@
...
+		},
+		"return error when recipient key parsing fails": {
+			sender:    senderPrivateKeyHex,
+			keyID:     keyID,
+			recipient: invalidKeyHex,
+		},
+	}
+	for name, test := range errorTestCases {
+		t.Run(name, func(t *testing.T) {
+			address, err := brc29.Address(test.sender, test.keyID, test.recipient)
```
ac4ch, line 128:
```suggestion
                        // When:
			address, err := brc29.Address(test.sender, test.keyID, test.recipient)
```

pkg/internal/sdk/template/brc29/brc29_address_test.go
```diff
@@ -0,0 +1,152 @@
...
+
+			require.Nil(t, address)
```
ac4ch, line 131:
```suggestion
                       // Then
			require.Nil(t, address)
```

pkg/internal/sdk/template/brc29/brc29_address_test.go
```diff
@@ -0,0 +1,152 @@
...
+			require.Error(t, err)
+		})
+	}
+
+	t.Run("return error when nil is passed as sender private key deriver", func(t *testing.T) {
+		var keyDeriver *sdk.KeyDeriver
```
ac4ch, line 138:
```suggestion
                // Given:
		var keyDeriver *sdk.KeyDeriver
```

pkg/internal/sdk/template/brc29/brc29_address_test.go
```diff
@@ -0,0 +1,152 @@
...
+
+		address, err := brc29.Address(keyDeriver, keyID, recipientPublicKeyHex)
```
ac4ch, line 141:
```suggestion
                // When:
		address, err := brc29.Address(keyDeriver, keyID, recipientPublicKeyHex)
```

pkg/internal/sdk/template/brc29/brc29_address_test.go
```diff
@@ -0,0 +1,152 @@
...
+
+		assert.Error(t, err)
```
ac4ch, line 144:
```suggestion
                // Then
		assert.Error(t, err)
```

ac4ch: Please do the same to other cases :) if u agree with my comments

pkg/internal/sdk/template/brc29/brc29_address_test.go
```diff
@@ -0,0 +1,152 @@
...
+		require.Nil(t, address)
+	})
+
+	t.Run("return error when nil is passed as sender private key", func(t *testing.T) {
+		var priv *ec.PrivateKey
+
+		address, err := brc29.Address(priv, keyID, recipientPublicKeyHex)
+
+		assert.Error(t, err)
```
ac4ch, line 156:
```suggestion
		require.Error(t, err)
```
this is minor thing here, but slightly inconsistent with the test fail-fast .. would u agree ?

pkg/internal/sdk/template/brc29/brc29_address_test.go
```diff
@@ -0,0 +1,152 @@
...
+		require.Equal(t, expectedAddress, address.AddressString)
+	})
+
+	t.Run("return testnet address created with brc29", func(t *testing.T) {
+		pub, err := ec.PublicKeyFromString(recipientPublicKeyHex)
+		require.NoError(t, err)
+
+		address, err := brc29.Address(senderPrivateKeyHex, keyID, pub, brc29.WithTestNet())
+
+		assert.NoError(t, err)
```
ac4ch:
```suggestion
		require.NoError(t, err)
```
Don't think there is need to check next condition if the assertion on error fails here ... therefor i would kindly suggest to use `require`

pkg/internal/sdk/template/brc29/brc29_opts.go
```diff
@@ -0,0 +1,32 @@
+package brc29
+
+import sighash "github.com/bsv-blockchain/go-sdk/transaction/sighash"
+
+type lockOptions struct {
+	mainNet bool
+}
```
ac4ch, line 7:
```suggestion
// Network represents the blockchain network.
type Network int

const (
	MainNet Network = iota
	TestNet
)

func (n Network) String() string {
	switch n {
	case MainNet:
		return "MainNet"
	case TestNet:
		return "TestNet"
	default:
		return "Unknown"
	}
}

type lockOptions struct {
	network Network
}
```

WDYT ? for me looks nicer :) and more descriptive

pkg/internal/sdk/template/brc29/brc29_address_test.go
dorzepowski: I'm trying to use assert and require such a way, that if the test fail, I have full picture of what was not right,
therefore, even if there is an error, I would like to see if the result was nil.
This way, when sticked to this approach One can spot a weird behaviour, that even if the error was returned, the result wasn't nil and for example even pass all the rest of requirements.

pkg/internal/sdk/template/brc29/brc29_opts.go
dorzepowski: I started with that,
But then I will need to convert it back to boolean. 
This is implementation detail that is driven by the needs of script.Address function. No one from outside should bother about it.

pkg/internal/sdk/template/brc29/brc29_address_test.go
dorzepowski: thanks for pointing this out, I was biased by tests in go-sdk, but I prefer this approach.
Will update all the tests to have given:when:then

pkg/internal/sdk/template/brc29/brc29_template_test.go
```diff
@@ -0,0 +1,189 @@
...
+	t.Run("return error when nil is passed as recipient public key", func(t *testing.T) {
+		var pub *ec.PublicKey
+
+		lockingScript, err := brc29.Lock(senderPrivateKeyHex, keyID, pub)
+
+		assert.Error(t, err)
+		require.Nil(t, lockingScript)
+	})
+
+	errorTestCases := map[string]struct {
```
ac4ch, line 80:
```suggestion
        // Given:
	errorTestCases := map[string]struct {
```

pkg/internal/sdk/template/brc29/brc29_template_test.go
```diff
@@ -0,0 +1,189 @@
...
+		},
+		"return error when recipient key parsing fails": {
+			sender:    senderPrivateKeyHex,
+			keyID:     keyID,
+			recipient: invalidKeyHex,
+		},
+	}
+	for name, test := range errorTestCases {
+		t.Run(name, func(t *testing.T) {
+			lockingScript, err := brc29.Lock(test.sender, test.keyID, test.recipient)
```
ac4ch, line 114:
```suggestion
                        // When:
			lockingScript, err := brc29.Lock(test.sender, test.keyID, test.recipient)
```

pkg/internal/sdk/template/brc29/brc29_template_test.go
```diff
@@ -0,0 +1,189 @@
...
+
+			require.Nil(t, lockingScript)
```
ac4ch, line 117:
```suggestion
                        // Then:
			require.Nil(t, lockingScript)
```

pkg/internal/sdk/template/brc29/brc29_template_test.go
```diff
@@ -0,0 +1,189 @@
...
+	ec "github.com/bsv-blockchain/go-sdk/primitives/ec"
+	"github.com/bsv-blockchain/go-sdk/script/interpreter"
+	"github.com/bsv-blockchain/go-sdk/transaction"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestBRC29TemplateLock(t *testing.T) {
+	t.Run("should lock with P2PKH and BRC29 calculated address", func(t *testing.T) {
+		lockingScript, err := brc29.Lock(senderPrivateKeyHex, keyID, recipientPublicKeyHex)
```
ac4ch, line 19:
```suggestion
                // Given
		lockingScript, err := brc29.Lock(senderPrivateKeyHex, keyID, recipientPublicKeyHex)
```

and same suggestions as i pointed out in above test file

pkg/internal/sdk/template/brc29/brc29_address_test.go
dorzepowski: same thing as here:
https://github.com/bsv-blockchain/go-wallet-toolbox/pull/275#discussion_r2139542316

pkg/internal/sdk/template/brc29/brc29_types.go
```diff
@@ -0,0 +1,80 @@
...
+//   - string: a private key in DER HEX format
+//   - WIF: a WIF string
+//   - *sdk.KeyDeriver: a key deriver that can be used to derive the private key
+//   - *ec.PrivateKey: a private key object
+type CounterpartyPrivateKey interface {
+	string |
+		WIF |
+		*sdk.KeyDeriver |
+		*ec.PrivateKey
+}
```
ac4ch: i did say in `CounterpartyPublicKey` looks better and would suggest to do same here 
```suggestion
type CounterpartyPrivateKey interface {
	string | WIF | *sdk.KeyDeriver | *ec.PrivateKey
}
```

pkg/internal/sdk/template/brc29/brc29_types.go
```diff
@@ -0,0 +1,80 @@
...
+// Validate validates the key ID.
+//
+// The key ID must have a derivation prefix and derivation suffix.
+func (k *KeyID) Validate() error {
+	if k.DerivationPrefix == "" {
+		return fmt.Errorf("derivation prefix is required")
+	}
+	if k.DerivationSuffix == "" {
+		return fmt.Errorf("derivation suffix is required")
+	}
```
ac4ch:
```suggestion
	if k.DerivationPrefix == "" {
	return fmt.Errorf("key ID validation failed: derivation prefix is required")
}
if k.DerivationSuffix == "" {
	return fmt.Errorf("key ID validation failed: derivation suffix is required")
}
```
this way errors are more informative i think ... WDYT ?

mgosek-4chain (approved):

pkg/internal/sdk/template/brc29/brc29_types.go
```diff
@@ -0,0 +1,80 @@
...
+	return nil
+}
+
+// String returns the string that can be used for derivation.
+func (k *KeyID) String() string {
+	return k.DerivationPrefix + " " + k.DerivationSuffix
+}
```
ac4ch, line 77:
```suggestion
import "strings"

func (k *KeyID) String() string {
	prefix := strings.TrimSpace(k.DerivationPrefix)
	suffix := strings.TrimSpace(k.DerivationSuffix)
	return prefix + " " + suffix
}
```
do u think we should sanitize or normalize this (trim spaces) a bit more ? however as long as upstream correctly forms KeyID, this is safe too as it was :)

pkg/internal/sdk/template/brc29/brc29_utils.go
ac4ch, line 1: since both functions share similar/identical structure, u could DRY some of this logic via reusable internal helpers functions ? WDYT ?

ac4ch (approved):

pkg/internal/sdk/template/brc29/brc29_types.go
dorzepowski: there is no such a logic in typescript, and changing it would probably cause different output from Address. Therefore I'm keeping it as is.

pkg/internal/sdk/template/brc29/brc29_utils.go
dorzepowski: they looks similar but they differs that way, that I have no idea how to DRY (😉 ) it more

PR #277, opened 2025-06-11
[FEATURE] create WoC broadcaster

### LINKED ISSUE #163: [FEATURE] create WoC broadcaster

ac4ch:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter

### DIFF
--- a/pkg/internal/testabilities/testservices/fixture_woc.go
+++ b/pkg/internal/testabilities/testservices/fixture_woc.go
@@ -1,6 +1,8 @@
 package testservices
 
 import (
+	"encoding/hex"
+	"encoding/json"
 	"fmt"
 	"net"
 	"net/http"
@@ -8,6 +10,7 @@ import (
 	"testing"
 
 	"github.com/4chain-ag/go-wallet-toolbox/pkg/defs"
+	"github.com/bsv-blockchain/go-sdk/transaction"
 	"github.com/go-resty/resty/v2"
 	"github.com/go-softwarelab/common/pkg/to"
 	"github.com/jarcoal/httpmock"
@@ -22,10 +25,10 @@ type WhatsOnChainFixture interface {
 	WillRespondWithInternalFailure()
 	WillRespondWithMerklePath(status int, txID string, responseBody string)
 	WillRespondWithBlockHeader(status int, blockHash string, responseBody string)
-
 	WhenQueryingMerklePath(txID string) WhatsOnChainMerklePathQueryFixture
 	WhenQueryingBlockHeader(blockHash string) WhatsOnChainBlockHeaderQueryFixture
-
+	WillRespondWithBroadcast(status int, responseBody string, err error)
+	WillAlwaysReturnPostBEEFSuccess(txids ...string)
 	Transport() *httpmock.MockTransport
 	HttpClient() *resty.Client
 }
@@ -232,4 +235,61 @@ func (f *wocFixture) HttpClient() *resty.Client {
 	client := resty.New()
 	client.GetClient().Transport = f.transport
 	return client
+}
+
+func (f *wocFixture) WillRespondWithBroadcast(status int, responseBody string, err error) {
+	responder := func(req *http.Request) (*http.Response, error) {
+		if err != nil {
+			return nil, err
+		}
+		res := httpmock.NewStringResponse(status, responseBody)
+		res.Header.Set("Content-Type", "application/json")
+		return res, nil
+	}
+
+	url := mockBroadcastURL(f.network)
+	f.transport.RegisterResponder("POST", url, responder)
+}
+
+func (f *wocFixture) WillAlwaysReturnPostBEEFSuccess(txids ...string) {
+	f.Transport().RegisterResponder("POST", mockBroadcastURL(f.network), func(req *http.Request) (*http.Response, error) {
+		var body struct {
+			TxHex string `json:"txhex"`
+		}
+		err := json.NewDecoder(req.Body).Decode(&body)
+		if err != nil {
+			return httpmock.NewStringResponse(http.StatusBadRequest, "bad request"), nil
+		}
+
+		rawTx, err := hex.DecodeString(body.TxHex)
+		if err != nil {
+			return httpmock.NewStringResponse(http.StatusBadRequest, "invalid hex"), nil
+		}
+
+		computedTxid := computeTxID(rawTx)
+
+		for _, txid := range txids {
+			if txid == computedTxid {
+				respBody := fmt.Sprintf(`{"txid":"%s"}`, txid)
+				resp := httpmock.NewStringResponse(http.StatusOK, respBody)
+				resp.Header.Set("Content-Type", "application/json")
+				return resp, nil
+			}
+		}
+
+		return httpmock.NewStringResponse(http.StatusBadRequest, "txid not found"), nil
+	})
+}
+
+func mockBroadcastURL(network defs.BSVNetwork) string {
+	return fmt.Sprintf("https://api.whatsonchain.com/v1/bsv/%s/tx/raw", network)
+}
+
+// computeTxID takes raw transaction bytes and returns the transaction ID (txid) as string.
+func computeTxID(rawTx []byte) string {
+	tx, err := transaction.NewTransactionFromBytes(rawTx)
+	if err != nil {
+		return ""
+	}
+	return tx.TxID().String()
 }
--- a/pkg/services/internal/whatsonchain/broadcast.go
+++ b/pkg/services/internal/whatsonchain/broadcast.go
@@ -1,0 +1,85 @@
+package whatsonchain
+
+import (
+	"context"
+	"encoding/hex"
+	"fmt"
+	"net/http"
+	"strings"
+	"time"
+
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/internal/txutils"
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/wdk"
+)
+
+type BroadcastStatus int
+
+const (
+	StatusError BroadcastStatus = iota
+	StatusSuccess
+	StatusAlreadyBroadcasted
+	StatusDoubleSpend
+	StatusMissingInputs
+)
+
+type broadcastRequest struct {
+	TxHex string `json:"txhex"`
+}
+
+// broadcast submits a raw transaction to WhatsOnChain.
+func (woc *WhatsOnChain) broadcast(ctx context.Context, rawTx []byte) (BroadcastStatus, string, error) {
+	rawTxHex := hex.EncodeToString(rawTx)
+	txid := txutils.TransactionIDFromRawTx(rawTx)
+
+	type txResponse struct {
+		Txid string `json:"txid"`
+	}
+	var txResp txResponse
+
+	req := woc.httpClient.
+		R().
+		SetContext(ctx).
+		SetResult(&txResp).
+		SetBody(broadcastRequest{TxHex: rawTxHex}).
+		AddRetryCondition(retryOnTooManyRequestsStatus)
+
+	url := fmt.Sprintf("%s/tx/raw", woc.url)
+	res, err := req.Post(url)
+	if err != nil {
+		return StatusError, "", fmt.Errorf("failed to send request to WoC: %w", err)
+	}
+
+	responseText := res.String()
+
+	if res.StatusCode() != http.StatusOK {
+		switch {
+		case strings.Contains(responseText, "already in mempool"):
+			return StatusAlreadyBroadcasted, txid, nil
+		case strings.Contains(responseText, "txn-mempool-conflict"):
+			return StatusDoubleSpend, txid, nil
+		case strings.Contains(responseText, "missing inputs"):
+			return StatusMissingInputs, txid, nil
+		default:
+			return StatusError, "", fmt.Errorf("woc returned error %d: %s", res.StatusCode(), responseText)
+		}
+	}
+
+	if txResp.Txid != txid {
+		return StatusError, txResp.Txid, fmt.Errorf("woc returned txid %s does not match local calculated txid %s", txResp.Txid, txid)
+	}
+
+	return StatusSuccess, txid, nil
+}
+
+func convertNotes(notes []string) wdk.Notes {
+	converted := make(wdk.Notes, len(notes))
+	for i, note := range notes {
+		now := time.Now()
+		converted[i] = wdk.ReqHistoryNote{
+			When: &now,
+			What: note,
+			Args: nil,
+		}
+	}
+	return converted
+}
--- a/pkg/services/internal/whatsonchain/constants.go
+++ b/pkg/services/internal/whatsonchain/constants.go
@@ -1,6 +1,8 @@
 package whatsonchain
 
-import "time"
+import (
+	"time"
+)
 
 const (
 	// Retries is the number of retries the client should make when trying to query for the chain's resource
--- a/pkg/services/internal/whatsonchain/service.go
+++ b/pkg/services/internal/whatsonchain/service.go
@@ -14,6 +14,7 @@ import (
 	"github.com/4chain-ag/go-wallet-toolbox/pkg/services/internal/httpx"
 	"github.com/4chain-ag/go-wallet-toolbox/pkg/services/internal/whatsonchain/internal/dto"
 	"github.com/4chain-ag/go-wallet-toolbox/pkg/wdk"
+	"github.com/bsv-blockchain/go-sdk/transaction"
 	"github.com/go-resty/resty/v2"
 	"github.com/go-softwarelab/common/pkg/to"
 )
@@ -164,4 +165,88 @@ func (woc *WhatsOnChain) FindChainTipHeader(ctx context.Context) (*wdk.ChainBloc
 	}
 
 	return header, nil
+}
+
+// PostBEEF attempts to post beef with given txIDs
+func (woc *WhatsOnChain) PostBEEF(ctx context.Context, beef *transaction.Beef, txIDs []string) (*wdk.PostedBEEF, error) {
+	if len(txIDs) == 0 {
+		return nil, fmt.Errorf("no txids provided")
+	}
+
+	if beef == nil {
+		return nil, fmt.Errorf("beef is required to post transactions")
+	}
+
+	rawTxs := make([][]byte, 0, len(txIDs))
+	for _, txid := range txIDs {
+		tx := beef.FindTransaction(txid)
+		if tx == nil {
+			return nil, fmt.Errorf("cannot find transaction %s in BEEF to broadcast", txid)
+		}
+
+		rawTxBytes := tx.Bytes()
+		if len(rawTxBytes) == 0 {
+			return nil, fmt.Errorf("empty raw transaction bytes for txid: %s", txid)
+		}
+		rawTxs = append(rawTxs, rawTxBytes)
+	}
+
+	txResults := make([]wdk.PostedTxID, 0, len(txIDs))
+	delayDuration := 3 * time.Second
+	var delay bool
+
+	for i, txid := range txIDs {
+		if delay {
+			select {
+			case <-ctx.Done():
+				return nil, fmt.Errorf("context canceled while waiting to broadcast transaction %s: %w", txid, ctx.Err())
+			case <-time.After(delayDuration):
+			}
+		}
+		delay = true
+
+		status, returnedTxid, err := woc.broadcast(ctx, rawTxs[i])
+		notes := []string{}
+		var result wdk.PostedTxIDResultStatus
+
+		if err != nil {
+			result = wdk.PostedTxIDResultError
+			notes = append(notes, fmt.Sprintf("broadcast error: %v", err))
+			returnedTxid = txid
+		} else {
+			switch status {
+			case StatusSuccess:
+				result = wdk.PostedTxIDResultSuccess
+			case StatusAlreadyBroadcasted:
+				result = wdk.PostedTxIDResultAlreadyKnown
+				notes = append(notes, "Transaction already in mempool")
+			case StatusDoubleSpend:
+				result = wdk.PostedTxIDResultDoubleSpend
+				notes = append(notes, "Possible double spend (txn-mempool-conflict)")
+			case StatusMissingInputs:
+				result = wdk.PostedTxIDResultMissingInputs
+				notes = append(notes, "Missing inputs (possible double spend)")
+			case StatusError:
+				result = wdk.PostedTxIDResultError
+				notes = append(notes, "Error broadcasting transaction")
+			default:
+				result = wdk.PostedTxIDResultError
+				notes = append(notes, "Unknown error broadcasting transaction")
+			}
+		}
+
+		convertedNotes := convertNotes(notes)
+
+		txResults = append(txResults, wdk.PostedTxID{
+			Result:       result,
+			TxID:         returnedTxid,
+			DoubleSpend:  status == StatusDoubleSpend || status == StatusMissingInputs,
+			AlreadyKnown: status == StatusAlreadyBroadcasted,
+			Notes:        convertedNotes,
+		})
+	}
+
+	return &wdk.PostedBEEF{
+		TxIDResults: txResults,
+	}, nil
 }
--- a/pkg/services/internal/whatsonchain/testabilities/fixtures.go
+++ b/pkg/services/internal/whatsonchain/testabilities/fixtures.go
@@ -12,12 +12,8 @@ import (
 
 type WoCServiceFixture interface {
 	testservices.ServicesFixture
-	NewWoCService() *whatsonchain.WhatsOnChain
-}
 
-type wocServiceFixture struct {
-	testservices.ServicesFixture
-	t testing.TB
+	NewWoCService(opts ...func(*whatsonchain.WhatsOnChain)) *whatsonchain.WhatsOnChain
 }
 
 func Given(t testing.TB) WoCServiceFixture {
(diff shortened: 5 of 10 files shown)


pkg/services/internal/whatsonchain/broadcast.go
```diff
@@ -0,0 +1,94 @@
...
+		return nil, fmt.Errorf("woc returned error %d: %s", res.StatusCode(), responseText)
+	}
+}
+
+func convertNotes(notes []string) wdk.Notes {
+	now := time.Now()
+	converted := make(wdk.Notes, len(notes))
+	for i, note := range notes {
+		converted[i] = wdk.ReqHistoryNote{
+			When: &now,
```
chris-4chain, line 79: a pointer to the same instance of time.Time (now) is used for every `wdk.ReqHistoryNote` in the returned slice.

This means all notes will point to the same underlying `time.Time` object, so if any one is mutated, the "When" field in all notes will reflect the change, which is almost certainly unintended.

pkg/services/internal/whatsonchain/broadcast.go
```diff
@@ -0,0 +1,94 @@
...
+		return &wdk.BroadcastResult{
+			Name: ServiceName,
+			TxID: txid,
+		}, nil
+	}
+
+	responseText := res.String()
+
+	switch responseText {
+	case ErrAlreadyInMempool:
```
chris-4chain: It assumes the response from `WhatsOnChain` will match exactly, down to whitespace and code structure. If WoC adds extra data, formatting, or returns a JSON object instead of a plain string, this check could easily break, leading to misclassification of error types.

**Suggestion**: consider using `strings.Contains`, or better, properly unmarshal and handle structured error responses if WoC supports it. 

For example:
```golang
if strings.Contains(responseText, "already in the mempool") { ... }
```

pkg/services/internal/whatsonchain/broadcast.go
```diff
@@ -0,0 +1,94 @@
...
+		})
+
+	url := fmt.Sprintf("%s/tx/raw", woc.url)
+
+	res, err := req.Post(url)
+	if err != nil {
+		return nil, fmt.Errorf("failed to send request to WoC: %w", err)
+	}
+
+	if res.StatusCode() == http.StatusOK {
```
chris-4chain: It seems to me more natural when error things are before final "successful" return.

Suggestion:
```golang
   return handleErrorResponse() // the switch case stuff
}

respTxID := res.String()
if respTxID != txid {
    return &wdk.BroadcastResult{
        Name:  ServiceName,
        TxID:  respTxID,
        Notes: []string{"Returned TxID does not match local calculation"},
    }, nil
}
return &wdk.BroadcastResult{
    Name: ServiceName,
    TxID: txid,
}, nil

```

pkg/services/internal/whatsonchain/broadcast.go
```diff
@@ -0,0 +1,94 @@
...
+			Notes:       []string{"Possible double spend (txn-mempool-conflict)"},
+		}, nil
+
+	case ErrMissingInputs:
+		return &wdk.BroadcastResult{
+			Name:        ServiceName,
+			TxID:        txid,
+			DoubleSpend: true,
+			Notes:       []string{"Missing inputs (possible double spend)"},
+		}, nil
```
chris-4chain: Hmmm... The `BroadcastResult` struct has a field: `Error` - maybe we should set it in "error-cases"?

pkg/services/internal/whatsonchain/constants.go
```diff
@@ -8,3 +8,12 @@ const (
 	// RetriesWaitTime is the duration to wait to retry a failed call to outside service
 	RetriesWaitTime = 2 * time.Second
 )
+
+const (
+	ErrAlreadyInMempool   = "unexpected response code 500: Transaction already in the mempool"
+	ErrTxnMempoolConflict = "unexpected response code 500: 258: txn-mempool-conflict"
+	ErrMissingInputs      = "unexpected response code 500: Missing inputs"
+	contentTypeJSON       = "application/json"
+	acceptTextPlain       = "text/plain"
```
chris-4chain: `contentTypeJSON` & `acceptTextPlain` seem to be unused.

pkg/services/internal/whatsonchain/service.go
```diff
@@ -137,3 +138,44 @@ func (woc *WhatsOnChain) UpdateBsvExchangeRate() (defs.BSVExchangeRate, error) {
...
+		if beef != nil {
+			tx := beef.FindTransaction(txid)
+			if tx != nil {
+				rawTxBytes = tx.Bytes()
+			}
+		}
+
+		if rawTxBytes == nil {
+			return nil, fmt.Errorf("cannot find transaction for txid: %s", txid)
+		}
```
chris-4chain: This could be refactored. 

1. Beef can be checked against `nil` only once. 
2. tx.Bytes() never returns `nil`, but eventually could return empty bytes

```suggestion

	if beef == nil {
		return nil, fmt.Errorf("beef is required to post transactions")
	}

	txResults := make([]wdk.PostedTxID, 0, len(txIDs))

	for _, txid := range txIDs {
		tx := beef.FindTransaction(txid)
		if tx == nil {
			return nil, fmt.Errorf("cannot find transaction for txid: %s", txid)
		}

		rawTxBytes := tx.Bytes()
		if len(rawTxBytes) == 0 {
			return nil, fmt.Errorf("empty raw transaction bytes for txid: %s", txid)
		}
```

pkg/wdk/services_broadcast.go
```diff
@@ -0,0 +1,25 @@
+package wdk
+
+// BroadcastResult is result of submitting single raw transaction to the service
+type BroadcastResult struct {
```
chris-4chain: I'm not sure if this struct should be added to `wdk`. 
If you need it, add it to internal `service/woc` but the "public" interface is through `wdk.PostedBEEF`.

pkg/services/internal/whatsonchain/broadcast_test.go
```diff
@@ -0,0 +1,120 @@
...
+	fixture := testabilities.Given(t)
+	woc := fixture.NewWoCService()
+
+	client := fixture.WhatsOnChain().HttpClient()
+	httpmock.ActivateNonDefault(client.GetClient())
+	defer httpmock.DeactivateAndReset()
+
+	rawTxHex := "0100000001abcdef"
+	rawTxBytes, _ := hex.DecodeString(rawTxHex)
+	calculatedTxID := txutils.TransactionIDFromRawTx(rawTxBytes)
```
chris-4chain: Please use `testvectors` 🙏

pkg/services/internal/whatsonchain/service.go
chris-4chain: I can't see any changes regarding my suggestion.

ac4ch: ok now i changed fully to what u "suggested"

pkg/services/services.go
```diff
@@ -92,6 +95,24 @@ func (s *WalletServices) FindChainTipHeader(ctx context.Context) (*wdk.ChainBloc
 	return result, nil
 }
 
+// EnableOnlyArc allows tests to enable only ARC for PostBEEF
+func (s *WalletServices) EnableOnlyArc() {
```
dorzepowski: I totally don't want those methods, also IMHO we need to look at the test cases differently not from the point of having or not having some single service enabled. But rather how it is working as a whole.

pkg/services/services.go
```diff
@@ -51,6 +52,7 @@ func New(logger *slog.Logger, config defs.WalletServices, opts ...func(*options.
 		config:       &config,
 		logger:       logger,
 		whatsonchain: woc,
+		arcService:   arcService,
```
dorzepowski: we don't need this assignment.

pkg/internal/testabilities/testservices/fixture_woc.go
```diff
@@ -233,3 +233,29 @@ func (f *wocFixture) HttpClient() *resty.Client {
...
+			return nil, err
+		}
+		res := httpmock.NewStringResponse(status, responseBody)
+		res.Header.Set("Content-Type", "application/json")
+		return res, nil
+	}
+
+	url := mockBroadcastURL(f.network)
+	f.transport.RegisterResponder("POST", url, responder)
+}
```
mgosek-4chain, line 252: @dorzepowski  @chris-4chain  I'd prefer to implement two distinct methods: one for handling success and one for failure. WDYT? 

Additionally, please use `http.MethodPost` instead of string literals.

pkg/services/internal/whatsonchain/broadcast.go
```diff
@@ -0,0 +1,117 @@
...
+
+	// Data (optional) raw response data for debugging
+	Data string
+
+	// Error (optional) error returned by service, if any
+	Error error
+}
+
+// Broadcast submits a raw transaction to WhatsOnChain.
+func (woc *WhatsOnChain) Broadcast(ctx context.Context, rawTx []byte) (*BroadcastResult, error) {
```
dorzepowski: do we need this method public?

until we really need it public, I would recommend to keep it private.

pkg/services/internal/whatsonchain/broadcast.go
```diff
@@ -0,0 +1,117 @@
...
+	rawTxHex := hex.EncodeToString(rawTx)
+	txid := txutils.TransactionIDFromRawTx(rawTx)
+
+	req := woc.httpClient.
+		R().
+		SetContext(ctx).
+		SetBody(map[string]string{txHexField: rawTxHex}).
```
dorzepowski: I would prefer a dedicated struct for that, as it is easier to extend/change in the future.

pkg/services/internal/whatsonchain/broadcast.go
```diff
@@ -0,0 +1,117 @@
...
+		AddRetryCondition(func(res *resty.Response, err error) bool {
```
dorzepowski: you have function for that 🙂

pkg/services/internal/whatsonchain/broadcast.go
```diff
@@ -0,0 +1,117 @@
...
+			return &BroadcastResult{
+				Name:        ServiceName,
+				TxID:        txid,
+				DoubleSpend: true,
+				Notes:       []string{"Missing inputs (possible double spend)"},
+				Error:       ErrMissingInputs,
+			}, fmt.Errorf("woc returned error %d response message: %s - %w", res.StatusCode(), responseText, ErrMissingInputs)
+		}
+		return nil, fmt.Errorf("woc returned error %d: %s", res.StatusCode(), responseText)
+	}
```
mgosek-4chain: Why do we need to return the broadcast result along with the error? 🤔

pkg/services/internal/whatsonchain/broadcast.go
```diff
@@ -0,0 +1,117 @@
...
+		return nil, fmt.Errorf("failed to send request to WoC: %w", err)
+	}
+
+	if res.StatusCode() != http.StatusOK {
+		responseText := res.String()
+		if strings.Contains(responseText, "already in mempool") {
+			return &BroadcastResult{
+				Name:  ServiceName,
+				TxID:  txid,
+				Notes: []string{"Transaction already in mempool"},
```
dorzepowski: I would skip all those notes for now, because we're not handling them at all, and the structure of the notes is something different then simple a list of strings. For example, when should be set already here to date now, but as I said, let's skip it for now, until storages synchronization will be finished, then we will have a task to support notes in different places.

pkg/services/internal/whatsonchain/broadcast_test.go
```diff
@@ -0,0 +1,121 @@
...
+			result, err := woc.Broadcast(t.Context(), rawTxBytes)
+
+			// Then:
+			if tc.expectError {
+				require.Error(t, err)
+				if tc.expectResult != nil && tc.expectResult.Error != nil {
+					require.ErrorIs(t, err, tc.expectResult.Error)
+				}
+				return
+			}
```
mgosek-4chain: I think it makes more sense to have two separate unit tests. One for positive cases, and one for errors.

pkg/services/internal/whatsonchain/service.go
```diff
@@ -165,3 +166,46 @@ func (woc *WhatsOnChain) FindChainTipHeader(ctx context.Context) (*wdk.ChainBloc
...
+
+		notes := convertNotes(broadcastResult.Notes)
+
+		txResults = append(txResults, wdk.PostedTxID{
+			Result:      wdk.PostedTxIDResultSuccess,
+			TxID:        broadcastResult.TxID,
+			DoubleSpend: broadcastResult.DoubleSpend,
+			Notes:       notes,
+		})
+	}
```
mgosek-4chain: To follow the SOLID principles, consider moving the broadcasting functionality into a separate service. Also note that this is a synchronous implementation and will fail if the initial broadcast request fails.

pkg/services/services_post_beef_test.go
```diff
@@ -84,4 +86,74 @@ func TestPostBEEF(t *testing.T) {
...
+		slices.ForEach(response, func(item *wdk.PostBEEFServiceResult) {
+			assert.NotEmpty(t, item.Name)
+			assert.Equal(t, "WhatsOnChain", item.Name)
+			assert.NoError(t, item.Error)
+			if assert.NotNil(t, item.PostedBEEFResult) {
+				result := item.PostedBEEFResult
+				assert.Lenf(t, result.TxIDResults, len(txids), "service %s returned unexpected number of results", item.Name)
+			}
+		})
+	})
```
mgosek-4chain: Why don't you create the expected slice and compare it with the `response` using standard assertions? 🤔

mgosek-4chain (changes requested):

pkg/services/internal/whatsonchain/broadcast.go
```diff
@@ -0,0 +1,117 @@
...
+				Error: ErrAlreadyInMempool,
+			}, fmt.Errorf("woc returned error %d response message: %s - %w", res.StatusCode(), responseText, ErrAlreadyInMempool)
```
dorzepowski: why you're returning both, result and error?

This is something misleading and also is not working as expected:
For example in case of double spending, it will return an error instead of the result with double spending.
Other example - in case of Already in mempool the result should be success, while the code will treat it as an failure.


IMHO this method, should actually return a simple status (it can be iota) xor error if this is something that we really treat as an error.

In other words the possible results should be like:
1. `success, nil`
2. `alreadyBroadcasted, nil`
3. `doubleSpending, nil`
4. `missingInputs, nil`
5. `0, fmt.Errorf(...)`  -  zero value in iota can be named as error status

And the in the PostBEEF it should be mapped to proper wdk.PostedTxID result

pkg/services/internal/whatsonchain/broadcast_test.go
ac4ch: for me this is readable why i would need to change test ?

pkg/services/internal/whatsonchain/service.go
ac4ch: what separate service ? i am not getting what u mean 

current implementation uses a dedicated `WhatsOnChain` service struct with `Broadcast` method...this method handles the full request/response cycle synchronously and returns detailed results/errors ...
from my perspective, this encapsulates the broadcasting logic as a service already, following the single responsibility principle at high level... 

Could u please elaborate a bit on what u mean by *“a separate service”*? Are u suggesting a separate component specifically to handle retries, queueing, or asynchronous broadcast workflows outside the current method? or fully independent service ?

pkg/services/internal/whatsonchain/broadcast.go
ac4ch: are u saying this is done different in TS ?

pkg/services/internal/whatsonchain/service.go
```diff
@@ -165,3 +166,46 @@ func (woc *WhatsOnChain) FindChainTipHeader(ctx context.Context) (*wdk.ChainBloc
...
+		tx := beef.FindTransaction(txid)
+		if tx == nil {
+			return nil, fmt.Errorf("cannot find transaction for txid: %s", txid)
+		}
+
+		rawTxBytes := tx.Bytes()
+		if len(rawTxBytes) == 0 {
+			return nil, fmt.Errorf("empty raw transaction bytes for txid: %s", txid)
+		}
+
```
dorzepowski, line 193: in typescript there is a delay between each call to give WoC a time to propagate the transaction.

pkg/services/services_post_beef_test.go
ac4ch: because other did same why .. so why sudden i should change this ? 
```go
slices.ForEach(response, func(item *wdk.PostBEEFServiceResult) {
			assert.NotEmpty(t, item.Name)
			assert.NoError(t, item.Error)
			if assert.NotNil(t, item.PostedBEEFResult) {
				result := item.PostedBEEFResult
				assert.Lenf(t, result.TxIDResults, len(txids), "service %s returned unexpected number of results", item.Name)
			}
		})
```

pkg/services/internal/whatsonchain/service.go
```diff
@@ -165,3 +166,46 @@ func (woc *WhatsOnChain) FindChainTipHeader(ctx context.Context) (*wdk.ChainBloc
...
+	txResults := make([]wdk.PostedTxID, 0, len(txIDs))
+
+	for _, txid := range txIDs {
+		tx := beef.FindTransaction(txid)
+		if tx == nil {
+			return nil, fmt.Errorf("cannot find transaction for txid: %s", txid)
+		}
+
+		rawTxBytes := tx.Bytes()
+		if len(rawTxBytes) == 0 {
```
dorzepowski, line 188: IMHO we should make the validation (or collect all the rawTx's) before broadcasting loop, 
Because if there is something missing, I'm would rather expect to fail everything before broadcasting anything - so we won't end up with unknow status.

pkg/services/services_post_beef_test.go
ac4ch: unless u saying whole test need redesign .. and follow now your suggestions ?

pkg/services/internal/whatsonchain/service.go
```diff
@@ -165,3 +166,46 @@ func (woc *WhatsOnChain) FindChainTipHeader(ctx context.Context) (*wdk.ChainBloc
...
+			return nil, fmt.Errorf("empty raw transaction bytes for txid: %s", txid)
+		}
+
+		broadcastResult, err := woc.Broadcast(ctx, rawTxBytes)
+		if err != nil {
+			return nil, fmt.Errorf("failed to broadcast transaction %s: %w", txid, err)
```
dorzepowski: we shouldn't return error here.
The expected result of PostBEEF is a result for each transaction even if it was an error.

The only case when we should return an error, is when the provided arguments are invalid.

So in case when woc.Broadcast returns an error, we should append txResults with PostedTxID error result.

pkg/services/internal/whatsonchain/broadcast.go
mgosek-4chain: No, I'm not. I'm just suggesting this is not typical. @chris-4chain  @dorzepowski Please have a look and let us know if it's all right :)

pkg/services/internal/whatsonchain/broadcast_test.go
```diff
@@ -0,0 +1,121 @@
...
+		WithP2PKHOutput(90)
+
+	rawTxBytes := txSpec.TX().Bytes()
+	calculatedTxID := txSpec.ID()
+
+	fixture := testabilities.Given(t)
+	woc := fixture.NewWoCService()
+
+	client := fixture.WhatsOnChain().HttpClient()
+	httpmock.ActivateNonDefault(client.GetClient())
```
dorzepowski: there is no need for that, becuase we're instantiating the client with mocked transport so this is redundant without any benefits.

pkg/services/internal/whatsonchain/broadcast_test.go
```diff
@@ -0,0 +1,121 @@
...
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			// Given:
+			fixture.WhatsOnChain().WillRespondWithBroadcast(tc.httpStatus, tc.httpResponse, nil)
+
+			// When:
+			result, err := woc.Broadcast(t.Context(), rawTxBytes)
```
dorzepowski: you should have tests for PostBEEF not for Broadcast
Because PostBEEF is something that we really want to work according to some "contract".

pkg/services/internal/whatsonchain/testabilities/fixtures.go
```diff
@@ -27,16 +22,27 @@ func Given(t testing.TB) WoCServiceFixture {
 	}
 }
 
-func (f *wocServiceFixture) NewWoCService() *whatsonchain.WhatsOnChain {
+type wocServiceFixture struct {
+	testservices.ServicesFixture
+	t testing.TB
+}
+
+func (f *wocServiceFixture) NewWoCService(opts ...func(*whatsonchain.WhatsOnChain)) *whatsonchain.WhatsOnChain {
```
dorzepowski, line 31: why did you add those opts? you don't even use them?

pkg/services/services_post_beef_test.go
```diff
@@ -14,7 +14,7 @@ import (
 )
 
 func TestPostBEEF(t *testing.T) {
-	t.Run("successfully post BEEF with single tx IDs", func(t *testing.T) {
+	t.Run("successfully post BEEF with single tx IDs (ARC)", func(t *testing.T) {
```
dorzepowski: instead of limiting this test to ARC only, we should setup all the subservices to sucessfully handle broadcasting the transaction.

pkg/services/services_post_beef_test.go
```diff
@@ -47,7 +48,7 @@ func TestPostBEEF(t *testing.T) {
 		})
 	})
 
-	t.Run("successfully post BEEF with multiple tx IDs", func(t *testing.T) {
+	t.Run("successfully post BEEF with multiple tx IDs (ARC)", func(t *testing.T) {
```
dorzepowski: sam here - all services should handle broadcasting correctly.

pkg/services/services_post_beef_test.go
```diff
@@ -84,4 +86,74 @@ func TestPostBEEF(t *testing.T) {
 		})
 	})
 
+	t.Run("successfully post BEEF with single tx IDs (WoC)", func(t *testing.T) {
```
dorzepowski: not needed test case

But now it would be good to have also a tests, where:
1. WoC returns error, all the rest returns success
2. Arc returns error, all the rest returns success
3. All returns error

Maybe those test cases could be somehow represented as table tests.

pkg/services/services_post_beef_test.go
```diff
@@ -84,4 +86,74 @@ func TestPostBEEF(t *testing.T) {
...
+			assert.NotEmpty(t, item.Name)
+			assert.NoError(t, item.Error)
+			if assert.NotNil(t, item.PostedBEEFResult) {
+				result := item.PostedBEEFResult
+				assert.Lenf(t, result.TxIDResults, len(txids), "service %s returned unexpected number of results", item.Name)
+			}
+		})
+	})
+
+	t.Run("successfully post BEEF with multiple tx IDs (WoC)", func(t *testing.T) {
```
dorzepowski: same as above not needed test case, should be covered by All Services Succeed test case

pkg/services/internal/whatsonchain/broadcast.go
ac4ch: i made it public as i used _test pkg to test this method .. yes we can do private .. but then do i need to make test for this ?

pkg/services/internal/whatsonchain/broadcast.go
ac4ch: yes it was left waiting on merge from merkle path ..

pkg/services/internal/whatsonchain/testabilities/fixtures.go
ac4ch:
```go
type ARCFServiceFixture interface {
	testservices.ServicesFixture

	NewArcService(opts ...func(*arc.Config)) *arc.Service
}
```
and u do not use them too 
image

ac4ch: so i will ask same question to u as i following ur fixtures implementations .. why u have opts when u do not use them ? :)

pkg/services/internal/whatsonchain/service.go
ac4ch: yes 3000 .. i was thinking if we need this here too or not .. will add it

pkg/services/internal/whatsonchain/service.go
mgosek-4chain: > current implementation uses a dedicated WhatsOnChain service struct with Broadcast method...this method handles the full request/response cycle synchronously and returns detailed results/errors ...
from my perspective, this encapsulates the broadcasting logic as a service already, following the single responsibility principle at high level...

Yes, and that's ok. We can switch to an asynchronous approach later. 

> Could u please elaborate a bit on what u mean by “a separate service”? Are u suggesting a separate component specifically to handle retries, queueing, or asynchronous broadcast workflows outside the current method? or fully independent service ?

I was thinking about a dedicated structure that encapsulates all the broadcasting logic, including retry policies etc., and then simply extending `WOC` using composition.  


```
type WOCBrodcaster struct {
	// ... 
}

func (w *WOCBrodcaster) Brodcast() error {
	// ... 
	return nil 
} 


type WOC struct {
	brodcaster *WOCBrodcaster
}

func (w *WOC) Brodcast() error {
	return w.brodcaster.Brodcast()
}
```

pkg/services/internal/whatsonchain/service.go
```diff
@@ -165,3 +166,87 @@ func (woc *WhatsOnChain) FindChainTipHeader(ctx context.Context) (*wdk.ChainBloc
...
+
+	if beef == nil {
+		return nil, fmt.Errorf("beef is required to post transactions")
+	}
+
+	rawTxs := make([][]byte, 0, len(txIDs))
+	for _, txid := range txIDs {
+		tx := beef.FindTransaction(txid)
+		if tx == nil {
+			return nil, fmt.Errorf("cannot find transaction for txid: %s", txid)
```
dorzepowski: minor
```suggestion
			return nil, fmt.Errorf("cannot find transaction %s in BEEF to broadcast", txid)
```

dorzepowski:
Overall looks fine now - so please merge when @mgosek-4chain and @chris-4chain will approve ✅

mgosek-4chain (approved):

PR #280, opened 2025-06-12
feat(#272): add find chain tip header support.

### LINKED ISSUE #272: [FEATURE] WalletServices: add findChainTipHeader method to WhatsOnChain service
## Desired Solution

## Suggested Implementation

## Alternatives Considered

## Additional Context

## Categorize Feature

mgosek-4chain:
## Description of Changes

- Extended what's on chain service with find block header tip function
- Update documentation & project types structure 

## Linked Issues / Tickets

"Closes #272".

## Testing Procedure

- [X] All tests pass locally

## Checklist:

- [X] I have run the linter


pkg/wdk/services_merkle_path.go
```diff
@@ -12,15 +12,15 @@ type MerklePathResult struct {
 	// MerklePath is the MerklePath of the transaction
 	MerklePath *transaction.MerklePath
 
-	// Header is the header of the block containing the transaction
```
dorzepowski, line 15: IMHO name of the field could stay the same, or changed only to `BlockHeader`

pkg/wdk/services_merkle_path.go
```diff
@@ -12,15 +12,15 @@ type MerklePathResult struct {
...
+	TransactionBlockHeader *TransactionBlockHeader
 
 	// Notes are the service debug notes for processing the request
 	Notes Notes
 }
 
-// BlockHeader is the header of a block
-type BlockHeader struct {
+// TransactionBlockHeader is the header of a block
+type TransactionBlockHeader struct {
```
dorzepowski: if you really want to make this name clear, then it should be rather MerklePathBlockHeader, because it's related to MerklePathResult.

pkg/internal/testabilities/testservices/fixture_woc.go
```diff
@@ -35,7 +37,32 @@ func NewWoCFixture(t testing.TB, opts ...Option) WhatsOnChainFixture {
...
 
+func (f *wocFixture) WillRespondWithTipBlockHeader(status int, err error, expectedResponse any) {
+	f.TB.Helper()
+	responder := func(status int, expectedResponse any) func(req *http.Request) (*http.Response, error) {
+		return func(req *http.Request) (*http.Response, error) {
+			if err != nil {
+				return nil, err
+			}
+
+			bb, err := json.Marshal(expectedResponse)
```
chris-4chain: Please rename it to sth more descriptive, like `expectedResponseJSON`

pkg/internal/testabilities/testservices/fixture_woc.go
```diff
@@ -35,7 +37,32 @@ func NewWoCFixture(t testing.TB, opts ...Option) WhatsOnChainFixture {
 	}
 }
 
+func (f *wocFixture) WillRespondWithTipBlockHeader(status int, err error, expectedResponse any) {
+	f.TB.Helper()
+	responder := func(status int, expectedResponse any) func(req *http.Request) (*http.Response, error) {
```
dorzepowski: there is already httpmock.NewJsonResponder that should do the same - so this custome is not needed.

pkg/services/internal/whatsonchain/dto/block_headers.go
```diff
@@ -0,0 +1,121 @@
+package dto
```
chris-4chain: The concept of an additional DTO package can be beneficial, but we need to keep in mind that we have to adjust other services and the things that are already in the PR to this DTO concept.

pkg/services/internal/whatsonchain/dto/block_headers.go
```diff
@@ -0,0 +1,121 @@
...
+type BlockHeaders []BlockHeader
+
+// First returns the first BlockHeader in the slice.
+//
+// This method is useful when only the earliest (or latest, depending on sort order) block
+// is of interest, such as when querying for the most recent block.
+//
+// It panics if the slice is nil or has zero elements. Use IsEmpty to check before calling
+// if the slice may be empty.
+func (bb BlockHeaders) First() BlockHeader {
```
chris-4chain: If some method can panic, we add a `Must` prefix to the name, like:
```aiignore
func (bb BlockHeaders) MustFirst() BlockHeader {
```

pkg/services/services.go
```diff
@@ -68,9 +69,32 @@ func New(logger *slog.Logger, config defs.WalletServices, opts ...func(*options.
 			"MerklePath",
 			servicequeue.NewService1(whatsonchain.ServiceName, arcService.MerklePath),
 		),
+
+		chainHeaderServices: servicequeue.NewQueue(
+			logger,
+			"What's on Chain",
```
chris-4chain: The method name should be rather `FindChainTipHeader`.

pkg/internal/testabilities/testservices/fixture_woc.go
```diff
@@ -35,7 +37,32 @@ func NewWoCFixture(t testing.TB, opts ...Option) WhatsOnChainFixture {
 	}
 }
 
+func (f *wocFixture) WillRespondWithTipBlockHeader(status int, err error, expectedResponse any) {
```
dorzepowski: I would split it into 2 separate methods,
- OnTipBlockHeaderWillRespondWith(status, body)
- OnTipBlockHeaderWillBeUnreachable() - actually this one could be rather WillBeUnreachable() and register responder to everything on `https://api.whatsonchain.com`

pkg/services/services_find_chain_tip_header_test.go
```diff
@@ -0,0 +1,100 @@
...
+		// then:
+		require.NoError(t, err)
+		require.Equal(t, expectedBlock, actualBlock)
+	})
+
+	t.Run("return an error when all block header services are unreachable", func(t *testing.T) {
+		// given:
+		err := fmt.Errorf("header service internal server error")
+		given := testservices.GivenServices(t)
+		given.WhatsOnChain().WillRespondWithTipBlockHeader(http.StatusInternalServerError, err, dto.BlockHeaders{})
```
dorzepowski: I believe, the way you setup it, it is working a different way than it would work in real scenario.

Because when error is passed your implementation will return an error from responder instead of response with status and error.

AFAIK - the error instead of json should be returned when there is some network issue or something related to communication that's why in ARC fixture I'm returning 
`net.UnknownNetworkError("tests defined this endpoint for tx as unreachable")` when setting some  endpoint to be unreachable. 
Or just not setting up the ARC mock if I want whole ARC to be unreachable.


AND if you want to have a result with status 500 you should simply still use a responder to respond with status 500 and some json

pkg/services/internal/whatsonchain/dto/block_headers.go
```diff
@@ -0,0 +1,121 @@
...
+// and can be used to avoid panics when accessing elements of the slice.
+func (bb BlockHeaders) IsEmpty() bool {
+	return len(bb) == 0
+}
+
+// BlockHeader represents a Bitcoin block as returned by a block explorer or full node API.
+//
+// It includes core header fields, metadata about the block’s position and content,
+// and references to adjacent blocks.
+type BlockHeader struct {
```
dorzepowski: maybe it doesn't need to have all those fields, as we don't use them.

pkg/services/services_find_chain_tip_header_test.go
```diff
@@ -0,0 +1,100 @@
...
+
+func TestFindChainTipHeader(t *testing.T) {
+	t.Run("return a single block header after call to the whats on chain service", func(t *testing.T) {
+		// given:
+		blockHeaders := newBlockHeaders(t)
+		expectedBlock := newChainBlockHeader(t, blockHeaders.First())
+
+		// and:
+		given := testservices.GivenServices(t)
+		given.WhatsOnChain().WillRespondWithTipBlockHeader(http.StatusOK, nil, blockHeaders)
```
dorzepowski: IMHO instead of this custom block building in this particular test, you could make method 
`given.WhatsOnChain().WillRespondWithTipBlockHeader()` to create a block header by itself - and return them so you can make assertion on it.

pkg/services/internal/whatsonchain/service.go
```diff
@@ -141,3 +135,37 @@ func (woc *WhatsOnChain) UpdateBsvExchangeRate() (defs.BSVExchangeRate, error) {
...
+
+	if res.StatusCode() != http.StatusOK {
+		return nil, fmt.Errorf("unexpected response from WhatsOnChain (URL: %s): status code %d", url, res.StatusCode())
+	}
+
+	if results.IsEmpty() {
+		return nil, fmt.Errorf("no block headers returned from WhatsOnChain (URL: %s); at least one expected", url)
+	}
+
+	first := results.First()
```
dorzepowski: IMHO it's not really needed to have this First method, because you already check if it is empty.

Eventually you could change the method First, to use `optional` package
```go
func (bb BlockHeaders) First() optional.Value[BlockHeader] {
	if bb.IsEmpty() {
		return optional.Empty[BlockHeader]()
	}
	return optional.OfValue(bb[0])
}
```

and then this part could be changed to:
```suggestion
	first, err := results.First().OrError(fmt.Errorf("no block headers returned from WhatsOnChain (URL: %s); at least one expected", url))
	if err != nil {
	  return nil, err
	}
```

Aha, this approach would need a small adjustement of .golangci-lint.yml by adding:
```yaml
linters-settings:
  wrapcheck:
    extra-ignore-sigs:
      - .OrError(err error)
```

pkg/services/internal/whatsonchain/dto/bsv_exchange_rate.go
```diff
@@ -0,0 +1,17 @@
+package dto
+
+// BSVExchangeRateResponse represents the exchange rate information for BSV
+// as returned by the WhatsOnChain API.
+//
+// It includes the timestamp of the rate, the exchange rate value itself,
+// and the fiat currency that the rate is quoted in.
+type BSVExchangeRateResponse struct {
```
dorzepowski: this way, you make it public in entire codebase (at least not published to the whole world)

I would prefer either of this (the order is not accidental):
1. keep those dto's private in a whatsonchain package
2. move dto to `internal/dto` which makes them internal to whatsonchain.

pkg/services/internal/whatsonchain/dto/block_headers.go
```diff
@@ -0,0 +1,121 @@
...
+
+	// NumTx is the number of transactions in the block.
+	// It may be equal to or override NTx depending on the API source.
+	NumTx int `json:"num_tx"`
+}
+
+// ConvertToChainBlockHeader converts a BlockHeader into a *wdk.ChainBlockHeader used in chain processing.
+//
+// It parses the `Bits` field from a hexadecimal string to a uint64 integer.
+// Returns an error if the conversion fails.
```
dorzepowski: this is not needed - and could become obsolete if something change in the future (for example ChainBlockHeader would also use string)

pkg/services/internal/whatsonchain/dto/block_headers.go
dorzepowski: IMHO we should declare here only fields that are needed.

pkg/services/internal/whatsonchain/service.go
mgosek-4chain: As discussed during chat, we'll use the straightforward approach (plain Go).

pkg/dto/block_headers_dto.go
```diff
@@ -0,0 +1,61 @@
...
+
+	// MerkleRoot is the root hash of the Merkle tree formed by all transactions in the block.
+	// This value appears in the block header and proves transaction inclusion.
+	MerkleRoot string `json:"merkleroot"`
+
+	// Time is the block’s creation timestamp in Unix epoch seconds, set by the miner.
+	Time uint64 `json:"time"`
+
+	// Nonce is a 32-bit number that miners iterate to find a hash meeting the difficulty target.
+	Nonce uint64 `json:"nonce"`
```
chris-4chain: As in your comment: `Nonce is a 32-bit`
- Nonce `uint64`

FYI: Bitcoin Nonce is 32-bit integer (uint32): https://en.bitcoin.it/wiki/Nonce

Similar comment regarding Version

pkg/services/internal/whatsonchain/dto/bsv_exchange_rate.go
dorzepowski: I think that you misunderstand me because now you make dto's "published" so it can be used not only by this code base, but by anyone who add this lib as dependency

Going back to previous options to choose:
1. move it to `pkg/services/internal/whatsonchain` and make the structs private
2. move it to `pkg/services/internal/whatsonchain/internal/dto` which make it internal for whatsonchain package

pkg/internal/testabilities/testservices/fixture_woc.go
```diff
@@ -35,7 +41,62 @@ func NewWoCFixture(t testing.TB, opts ...Option) WhatsOnChainFixture {
...
+		}),
+	)
+}
+
+func (f *wocFixture) OnTipBlockHeaderWillRespondWithEmptyList() []dto.BlockHeader {
+	f.TB.Helper()
+	f.transport.RegisterResponder(
+		http.MethodGet,
+		fmt.Sprintf("https://api.whatsonchain.com/v1/bsv/%s/block/headers?limit=1", f.network),
+		httpmock.NewJsonResponderOrPanic(http.StatusOK, []dto.BlockHeader{}),
```
dorzepowski: there is no need to use dto.BlockHeader structure, and you shouldn't actually use it, to ensure that this structure have right json setup to handle json from WoC.

Use string format, or map instead.

pkg/services/services.go
```diff
@@ -68,9 +69,32 @@ func New(logger *slog.Logger, config defs.WalletServices, opts ...func(*options.
...
+		),
 	}
 }
 
+// FindChainTipHeader queries multiple chain header services in sequence
+// and returns the most recent block header (chain tip) available.
+//
+// It attempts each service one by one until a successful result is returned.
+// If no service returns a result, it wraps and returns servicequeue.ErrEmptyResult.
+// Any other errors encountered during the process are returned with context.
```
dorzepowski:
```suggestion
// FindChainTipHeader attempts to get the most recent block header (chain tip) available.
```

chris-4chain (approved):

pkg/internal/testabilities/testservices/fixture_woc.go
```diff
@@ -62,6 +133,56 @@ func (f *wocFixture) WillRespondWithRawTx(status int, txID, rawTx string, err er
...
+
+const (
+	TestBlockVersionHex        = "30000000"
+	TestBlockMerkleRoot        = "c7a78f2edd611b0fe7aad6829a243e4a9e351e5ab203b7beb875ba1e6a80249e"
+	TestBlockBits              = "18110ef8"
+	TestBlockChainWork         = "000000000000000000000000000000000000000001669c7b159861f30c53271e"
+	TestBlockPreviousBlockHash = "000000000000000001885e0c6c302cbbacf927e1b5cf7884588973e72f8b704e"
+)
+
+type blockDTO struct {
```
dorzepowski: ### minor
```suggestion
type wocBlockResponseItem struct {
```

dorzepowski (approved):

PR #281, opened 2025-06-12
feat: manager sync to writer

chris-4chain:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter


pkg/internal/storage/repo/output_baskets.go
```diff
@@ -37,32 +36,59 @@ func (o *OutputBaskets) FindBasketByName(ctx context.Context, userID int, name s
...
+	err = o.db.WithContext(ctx).Transaction(func(tx *gorm.DB) error {
+		err := tx.Model(&models.OutputBasket{}).
+			Scopes(scopes.UserID(userID)).
+			Where("name = ?", basket.Name).
+			Count(&initialCount).Error
+		if err != nil {
+			return fmt.Errorf("failed to count existing output baskets: %w", err)
+		}
 
+		if initialCount > 0 {
```
ac4ch: are u using count to check existence ?  if so use `First()` or `OnConflict()` instead of `Count()`
since 
```sql
sqlite> PRAGMA index_list('bsv_output_baskets');
0|idx_bsv_output_baskets_deleted_at|0|c|0
1|idx_name_user_id|1|c|0
sqlite> PRAGMA index_info('idx_name_user_id');
0|4|name
1|7|user_id
sqlite> 
```
u are safe to use `First`

pkg/internal/storage/repo/sync_state.go
```diff
@@ -0,0 +1,123 @@
...
+		RefNum:             syncState.Reference,
+		SyncMap:            syncMapJSON,
+		When:               syncState.When,
+	}
+
+	if syncState.Satoshis != nil {
+		model.Satoshis = to.Ptr(syncState.Satoshis.Int64())
+	}
+
+	err = s.db.WithContext(ctx).Create(&model).Error
```
ac4ch, line 64:
```suggestion
	err = s.db.WithContext(ctx).Clauses(clause.OnConflict{
        Columns:   []clause.Column{{Name: "user_id"}, {Name: "storage_identity_key"}},
        DoUpdates: clause.AssignmentColumns([]string{"status", "sync_map", "when", "ref_num", "storage_name", "satoshis"}),
    }).Create(&model).Error
```
would not be better ? this way u would  not need `FindSyncState` ?

pkg/internal/storage/repo/sync_state.go
```diff
@@ -0,0 +1,123 @@
...
+
+	toUpdate := models.SyncState{
+		Status:  syncState.Status,
+		SyncMap: syncMapJSON,
+		When:    syncState.When,
+	}
+
+	if syncState.Satoshis != nil {
+		toUpdate.Satoshis = to.Ptr(syncState.Satoshis.Int64())
+	}
```
ac4ch:
```suggestion
	toUpdate := map[string]interface{}{
    "status":  syncState.Status,
    "sync_map": syncMapJSON,
    "when":    syncState.When,
}

if syncState.Satoshis != nil {
    toUpdate["satoshis"] = syncState.Satoshis.Int64()
}
```

would not be better to use map here instead of struct ? since i think GORM works better with maps as only updates what u provided ... no risk of zero overwrite ? 
https://gorm.io/docs/update.html#Update-Changed-Fields

pkg/internal/storage/repo/users.go
```diff
@@ -57,11 +54,30 @@ func (u *Users) CreateUser(ctx context.Context, identityKey, activeStorage strin
...
-	return &wdk.TableUser{
-		UserID:        user.UserID,
+	return mapUserModelToEntity(&user), nil
+}
+
+func (u *Users) UpdateUser(ctx context.Context, userID int, activeStorage string, updatedAt time.Time) error {
+	err := u.db.WithContext(ctx).
+		Model(&models.User{}).
+		Scopes(scopes.UserID(userID)).
+		Updates(models.User{
```
ac4ch: same thing as above ?

pkg/internal/testabilities/dbfixtures/db.go
```diff
@@ -48,3 +55,18 @@ func TestDatabase(t testing.TB) (db *database.Database, cleanup func()) {
 	require.NoError(t, err)
 	return db, func() {}
 }
+
+const sqliteFileNamePattern = `^file:(.+)\.sqlite(.*)$`
```
ac4ch, line 59: would not be better to compile it once globally ? instead of doing this inside below func each time func is called ? 
```suggestion
var sqliteFileNameRegex = regexp.MustCompile(`^file:(.+)\.sqlite(.*)$`)
```

pkg/internal/testabilities/dbfixtures/db.go
```diff
@@ -48,3 +55,18 @@ func TestDatabase(t testing.TB) (db *database.Database, cleanup func()) {
...
+
+func WithSQLiteFileName(fileName string) DBConfigModifier {
+	return func(config *defs.Database) {
+		if config.Engine != defs.DBTypeSQLite {
+			panic("WithSQLiteFileName modifier can only be used with SQLite engine (config.Engine = 'sqlite')")
```
ac4ch, line 64: is it not to harsh to call panic ? why Error would not be enough ?

pkg/storage/internal/sync/chunk_processor.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	currentDBHasNewerVersion := chunkUserInCurrentDB.UpdatedAt.After(p.chunk.User.UpdatedAt)
+	if currentDBHasNewerVersion {
+		return nil // No update needed
+	}
+
+	err = p.parent.repo.UpdateUser(p.ctx, p.chunk.User.UserID, p.chunk.User.ActiveStorage, p.chunk.User.UpdatedAt)
+	if err != nil {
+		return fmt.Errorf("failed to update user: %w", err)
```
ac4ch:
```suggestion
		fmt.Errorf("failed to update user %d: %w", p.chunk.User.UserID, err)
```
i would include here user info here ...

pkg/storage/internal/sync/sync_to_writer.go
```diff
@@ -0,0 +1,150 @@
...
+
+	userAuthOnWriterSide := wdk.AuthID{
+		IdentityKey: userIdentityKey,
+		UserID:      to.Ptr(userOnWriterSide.User.UserID),
+	}
+
+	var state syncingState
+
+	for range state.doWhileChangesMade() {
+		writerSyncState, err := writer.FindOrInsertSyncStateAuth(ctx, userAuthOnWriterSide, readerSettings.StorageIdentityKey, readerSettings.StorageName)
```
ac4ch, line 56: could we move this call before the loop ? dont think u expecting sync state to change outside loop ? am i right or not ?

pkg/internal/storage/repo/sync_state.go
chris-4chain: The `Find` is needed anyway because this functions should:
- return existing row if exists
- create if doesn't exist

"Upsert" (the `INSERT` with `ON CONFLICT`) doesn't do the same job.

pkg/internal/storage/repo/output_baskets.go
chris-4chain: `First` gets the whole object - which is unnecessary and get's CPU cycles - that's why I went for `count` strategy. 
👉 NOTE: That the function needs to "know" if the raw has been "created" or just "updated".

pkg/internal/storage/repo/sync_state.go
chris-4chain: Sure, why not. 
But there is one drawback of this method, (just to note) - that we need to use "db column names" - so the golang compiler can't check if we use proper column names 🤷

pkg/internal/testabilities/dbfixtures/db.go
chris-4chain: Doesn't matter - this is just for tests.
(`testabilities`)

pkg/internal/testabilities/dbfixtures/db.go
chris-4chain: 1. `Testabilities`
2. This shouldn't happen by design
3. Panic signals programmer that something weird happened - but only running tests.

pkg/internal/storage/repo/output_baskets.go
ac4ch: sure u have unique index per `user_id` and `name` so multiple rows are impossible.. 
`Count()` forces full scan (`aggregate`) which can be more expensive ... where `First()` stops at first matching row .. generally cheaper for existence check :) 

even this i think First is better solution here .. since u do not need to scan whole table .. but sure up to u .. since u trying to mimic `Exist` which is not exposed by GORM, and Count always aggregate entire result set ...

pkg/storage/internal/sync/sync_to_writer.go
chris-4chain: Sadly: no, because sync state changes every iteration - and moreover, the `writer` can be a "remote" one (e.g. JS version).
So, my answer would be:
- No, because of how the public api and sync mechanism are designed.

pkg/internal/storage/database/models/sync_state.go
```diff
@@ -0,0 +1,22 @@
...
+	gorm.Model
+
+	UserID             int    `gorm:"uniqueIndex:idx_user_storage_key"`
+	StorageIdentityKey string `gorm:"type:varchar(130);not null;uniqueIndex:idx_user_storage_key"`
+	StorageName        string `gorm:"type:varchar(128);not null"`
+	Status             wdk.SyncStatus
+	RefNum             string `gorm:"not null;uniqueIndex"`
+	SyncMap            json.RawMessage
+	When               *time.Time
+	Satoshis           *int64
```
mgosek-4chain, line 21: Open question: Should we add columns description? It seems to me that database schema could break if someone changes the field name. @dorzepowski  @chris-4chain

pkg/internal/storage/repo/output_baskets.go
dorzepowski: just saying 🙈

Insead of Count or First, you could:
1. try to update
2. if rows affected is 0 then insert

pkg/storage/storage_manager.go
```diff
@@ -73,6 +75,25 @@ func (m *WalletStorageManager) GetAuth(ctx context.Context) (wdk.AuthID, error)
 	}, nil
 }
 
+// SyncToWriter synchronizes wallet data from the active storage to the provided writer storage provider.
+// NOTE: reader(source) => writer(backup)
+func (m *WalletStorageManager) SyncToWriter(ctx context.Context, auth wdk.AuthID, writer wdk.WalletStorageProvider) (inserts, updates int, err error) {
+	// TODO: add locking mechanism to ensure that the active storage is not being modified while syncing
+
+	if m.identityKey != auth.IdentityKey {
```
dorzepowski: are you sure that auth argument is needed here?

pkg/storage/storage_manager.go
```diff
@@ -73,6 +75,25 @@ func (m *WalletStorageManager) GetAuth(ctx context.Context) (wdk.AuthID, error)
 	}, nil
 }
 
+// SyncToWriter synchronizes wallet data from the active storage to the provided writer storage provider.
+// NOTE: reader(source) => writer(backup)
+func (m *WalletStorageManager) SyncToWriter(ctx context.Context, auth wdk.AuthID, writer wdk.WalletStorageProvider) (inserts, updates int, err error) {
```
dorzepowski: shouldn't you check if writer is not nil?

pkg/storage/internal/sync/sync_to_writer.go
dorzepowski: OMG 😭😭😭

pkg/storage/internal/sync/sync_to_writer.go
```diff
@@ -0,0 +1,150 @@
...
+			MaxItems:     maxSyncItems,
+			Offsets:      s.buildOffsets(syncMap),
+		}
+
+		chunk, err := reader.GetSyncChunk(ctx, getSyncChunkArgs)
+		if err != nil {
+			return 0, 0, fmt.Errorf("failed to get sync chunk from reader storage: %w", err)
+		}
+
+		processChunkResult, err := writer.ProcessSyncChunk(ctx, getSyncChunkArgs, chunk)
```
dorzepowski, line 84: Maybe we should check if chunk has any data before calling ProcessSyncChunk?

pkg/storage/internal/sync/chunk_processor.go
```diff
@@ -0,0 +1,149 @@
...
+	err = p.parent.repo.UpdateSyncState(p.ctx, p.syncState)
+	if err != nil {
+		return fmt.Errorf("failed to update sync state: %w", err)
+	}
+
+	return nil
+}
+
+func (p *chunkProcessor) mergeUser() error {
+	chunkUserInCurrentDB, err := p.parent.repo.FindUser(p.ctx, p.chunk.User.IdentityKey)
```
dorzepowski: why we are retrieving this user again, we just did that when setting the p.User

pkg/storage/internal/sync/chunk_processor.go
```diff
@@ -0,0 +1,149 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to find chunk user: %w", err)
+	}
+
+	if chunkUserInCurrentDB == nil {
+		return fmt.Errorf("chunk user not found for userID %d", p.chunk.User.UserID)
+	}
+
+	if chunkUserInCurrentDB.ID != p.user.ID {
```
dorzepowski: how this is possible to ever happen?

pkg/storage/internal/sync/chunk_processor.go
```diff
@@ -0,0 +1,149 @@
...
+		return fmt.Errorf("chunk user ID %d does not match current DB user ID %d", p.chunk.User.UserID, chunkUserInCurrentDB.ID)
+	}
+
+	currentDBHasOlderVersion := chunkUserInCurrentDB.UpdatedAt.Before(p.chunk.User.UpdatedAt)
+	if !currentDBHasOlderVersion {
+		return nil // No update needed
+	}
+
+	err = p.parent.repo.UpdateUser(p.ctx, p.chunk.User.UserID, p.chunk.User.ActiveStorage, p.chunk.User.UpdatedAt)
```
dorzepowski: shouldn't you use `p.user.ID`?

pkg/internal/testabilities/dbfixtures/db.go
ac4ch: ya but still i would use this once globally :)

pkg/internal/testabilities/dbfixtures/db.go
```diff
@@ -48,3 +55,18 @@ func TestDatabase(t testing.TB) (db *database.Database, cleanup func()) {
...
+		}
+		re := regexp.MustCompile(sqliteFileNamePattern)
```
dorzepowski, line 66: ### just saying - alternative approach
Instead, you could check the testmode and if it is SqliteFile then set the sqlite file connection,
if it is postgres, then panic, that it's not supported yet,
if it is default then set the in memory string.

pkg/storage/storage_manager.go
chris-4chain: It is in "public" api in TS 🤷

pkg/storage/internal/sync/sync_to_writer.go
chris-4chain: Nope, that could break the logic (especially the TS one) because writer storage may "want" to receive an empty chunk to be informed that, it's over.

pkg/storage/internal/sync/chunk_processor.go
chris-4chain: Hmm, good point - I'll rework this section.

pkg/storage/internal/sync/chunk_processor.go
chris-4chain: 1. https://github.com/4chain-ag/wallet-toolbox/blob/7450651c043580426eb6ce0097746ccea896daa0/src/storage/schema/entities/EntityUser.ts#L83
2. After above comment the `if` looks differently, but still is necessary:
```golang
if p.chunk.User.IdentityKey != p.user.IdentityKey {
	return fmt.Errorf("chunk user identity key %s does not match current user identity key %s", p.chunk.User.IdentityKey, p.user.IdentityKey)
}
```

Keep in mind, that the `chunk` origins from "remote" source storage and we need to check if provided data are consistent.

pkg/storage/internal/sync/chunk_processor.go
chris-4chain: Yes, good catch 👍

pkg/internal/testabilities/dbfixtures/db.go
chris-4chain: "If it works, don't touch it" 😅

pkg/internal/storage/database/models/sync_state.go
chris-4chain: DB schema evolves at this stage of development so I think this would make more problems than benefits - but - in the future - why not 👍

pkg/internal/storage/repo/output_baskets.go
chris-4chain: @dorzepowski - That solves our discussion 👍 - Applied ✅

mgosek-4chain (approved):

pkg/storage/storage_manager.go
dorzepowski: But it seems pointless as manager already can create authId and it's doing that in other places. And we're actually using only identity key from the AuthID so it's even more confusing, why it can be passed.

I would remove it, and add if it's really needed.

Although it's kind of public interface, it isn't something that adopters (especially playing around with wallet toolbox) would call often - because there are better methods, that are using this one underneath.

pkg/storage/internal/sync/chunk_processor.go
dorzepowski: now it makes sense, because you're comparing chunk indentity key with contextual identity key.

But previously you were comparing identity key of the user loaded from database in different lines in the same code, in other words you were comparing the same row from database but from diffrent point of time (but the difference was about few miliseconds).

pkg/storage/storage_manager.go
chris-4chain: ok - adjusted. If the `auth` will be necessary, we will add it.

dorzepowski (approved):

PR #288, opened 2025-06-17
Suggestion/add internalize tx example

dorzepowski:


examples/setup/setup.go
```diff
@@ -0,0 +1,38 @@
...
+		// cleanup function
+	}
+}
+
+type Environment struct {
+	BSVNetwork defs.BSVNetwork `mapstructure:"bsv_network"`
+}
+
+func CreateAlice() *Setup {
+	// 1. Load the Environment config with use of viper,
```
chris-4chain: Would be good to use our config loader - but we would need to make it "exported".

examples/setup/setup.go
```diff
@@ -0,0 +1,38 @@
...
+	return nil, func() {
+		// cleanup function
+	}
+}
+
+type Environment struct {
+	BSVNetwork defs.BSVNetwork `mapstructure:"bsv_network"`
+}
+
+func CreateAlice() *Setup {
```
chris-4chain: Most probably that would need to return error as a second argument.

chris-4chain: Suggestion how to cover other users also:

```golang
func CreateBob() *Setup {
	return createUser(&Something{})
}

func createUser(config *Something) *Setup {
	return nil
}
```

examples/internalize_tx/internalize_tx.go
```diff
@@ -1,55 +1,40 @@
...
+var txID = pflag.String("txid", "15f47f2db5f26469c081e8d80d91a4b0f06e4a97abcc022b0b5163ac5f6cc0c8", "pass the chosen txid or simply change the default value when running the example")
+
+func main() {
+
+	pflag.Parse()
+
+	alice := setup.CreateAlice()
+
+	// Maybe instead we should have alice.GetWallet() which will have lazy initialization of the wallet 🤷
+	aliceWallet, cleanup := alice.CreateWallet()
```
chris-4chain: I would go for simpler implementation - explicit `CreateWallet` - without the "lazy/cache" magic

examples/src/method/internalize_tx.go
```diff
@@ -1,17 +1,33 @@
...
+
+	iar, err := userWallet.InternalizeAction(ctx, internalizeArgs, "originator")
+	if err != nil {
+		return sdk.InternalizeActionResult{}, err
+	}
+
+	return *iar, nil
+}
+
+func InternalizeFromFaucet(txID string, wallet *sdk.Wallet) error {
```
chris-4chain: I would advocate for `wallet` to be the first argument - for all "methods".
Like `self` in python :)
```golang
InternalizeFromFaucet(wallet *sdk.Wallet, txID string) error
```


Except, if ctx is required - in this case: `ctx, wallet, restOfArguments...`

examples/internalize_tx/internalize_tx.go
```diff
@@ -1,55 +1,40 @@
...
+	defer cleanup()
+
+	// Maybe allow for beef from const also
+	// if beef == "" {
+	//    beef = methods.GetBEEFHex(txID)
+	// }
+
+	err := methods.InternalizeFromFaucet(*txID, aliceWallet)
+	if err != nil {
+		fmt.Println("Failed to internalize tx")
```
chris-4chain: For `examples` we could consistently use `log` package.
Additionally, for cases, when "app" needs to be shut down - instead of returning an error, we could do `log.Fatal`/`log.Fatalf`

examples/setup/setup.go
dorzepowski: maybe it could simply panic?

chris-4chain: Or `log.Fatalf` - of course :)

dorzepowski: yes, I was thinking about something like this - but I think that createUser interface would be defined when we would prepare CreateBob function

examples/internalize_tx/internalize_tx.go
dorzepowski: I'm not sure about using log - the output from log methods is structured with date time, maybe it's better to have custom `show` package to show to handle outputs consistently?

chris-4chain: Or `print` package. That can contain more "specialized" methods too. 

```golang
package print

import (
	"fmt"
	"os"
)

// Println prints the arguments followed by a newline.
func Println(a ...any) {
	fmt.Println(a...)
}

// Printf formats according to a format specifier and writes to standard output.
func Printf(format string, a ...any) {
	fmt.Printf(format, a...)
}

// Fatal prints the arguments, then exits the program with status code 1.
func Fatal(a ...any) {
	fmt.Println(a...)
	os.Exit(1)
}

// Fatalf formats according to a format specifier, prints, then exits with status code 1.
func Fatalf(format string, a ...any) {
	fmt.Printf(format, a...)
	os.Exit(1)
}
```

PR #291, opened 2025-06-18
Feat/add faucet internalize example

cain4chain:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter


examples/utils/keyDerivation.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+}
+
+// KeyID creates a key ID from derivation prefix and suffix
+func KeyID(derivationPrefix, derivationSuffix string) string {
+	return fmt.Sprintf("%s %s", derivationPrefix, derivationSuffix)
+}
+
+// DerivationBytes creates derivation bytes from prefix and suffix
+func DerivationBytes(prefix, suffix string, opts *DerivationBytesOpts) DerivationBytesResult {
```
dorzepowski: Can be private, and no need for docs

examples/utils/keyDerivation.go
```diff
@@ -0,0 +1,101 @@
...
+
+	return DerivationPartsResult{
+		KeyID:             KeyID(derivationPrefix, derivationSuffix),
+		IdentityKey:       identityKey,
+		PaymentRemittance: paymentRemittance,
+	}
+}
+
+// KeyID creates a key ID from derivation prefix and suffix
+func KeyID(derivationPrefix, derivationSuffix string) string {
```
dorzepowski: can be private and no need for docs

examples/utils/keyDerivation.go
```diff
@@ -0,0 +1,101 @@
...
+
+	var identityKey string
+	if len(senderIdentityKey) > 0 && senderIdentityKey[0] != "" {
+		identityKey = senderIdentityKey[0]
+	} else {
+		privateKey, err := primitives.NewPrivateKey()
+		if err != nil {
+			panic(err)
+		}
+		identityKey = privateKey.PubKey().ToDERHex()
```
dorzepowski: I would go with Anyone key to have deterministic output from this function
```suggestion
		_, publicKey := sdk.AnyoneKey()
		identityKey = publicKey.ToDERHex()
```

examples/utils/keyDerivation.go
```diff
@@ -0,0 +1,101 @@
...
+	case "base64":
+		derivationPrefix, _ = base64.StdEncoding.DecodeString(prefix)
+	default:
+		derivationPrefix = []byte(prefix)
+	}
+
+	encoding = "utf8"
+	if opts != nil && opts.Encoding != "" {
+		encoding = opts.Encoding
+	}
```
dorzepowski: why you're doing the same again?

examples/utils/keyDerivation.go
```diff
@@ -0,0 +1,101 @@
...
+
+// DerivationParts creates derivation parts with default prefix and suffix
+// senderIdentityKey is optional - if not provided, a new random key will be generated
+func DerivationParts(senderIdentityKey ...string) DerivationPartsResult {
+	prefix := "someprefix" // set this to any string you want to change the address
+	suffix := "somesuffix" // set this to any string you want to change the address
+	bytes := DerivationBytes(prefix, suffix, nil)
+
+	derivationPrefix := base64.StdEncoding.EncodeToString(bytes.DerivationPrefix)
+	derivationSuffix := base64.StdEncoding.EncodeToString(bytes.DerivationSuffix)
```
dorzepowski: in TS the idea of this part was to:
1. if prefix is not empty then decode it to bytes
2. if prefix is empty then use default base64 prefix `SfKxPIJNgdI=`
	
3. if suffix is not empty, then decode it to bytes
4. if suffix is empty then use default base64 suffix := 'NaGLC6fMH50='

examples/utils/keyDerivation.go
dorzepowski: looks like the opts are redundant here

examples/utils/keyDerivation.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	paymentRemittance := &sdk.Payment{
+		DerivationPrefix:  derivationPrefix,
+		DerivationSuffix:  derivationSuffix,
+		SenderIdentityKey: identityKey,
+	}
+
+	return DerivationPartsResult{
+		KeyID:             KeyID(derivationPrefix, derivationSuffix),
```
dorzepowski: The KeyID is not used anywhere, and the IdentityKey is actually the same thing as SenderIdentityKey, therfore I would suggest to return simply paymentRemittance from here.

examples/example_setup/setup.go
```diff
@@ -0,0 +1,114 @@
...
+	var cfg SetupConfig
+	if err := v.Unmarshal(&cfg); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal config: %w", err)
+	}
+
+	return &cfg, nil
+}
+
+func CreateAlice() *Setup {
+	cfg, err := loadConfig("examples/example_setup/example-config.yaml")
```
dorzepowski: This could be done in second PR, but it would be good to have config-example.yaml which would be committed and use `config.yaml` which would be git ignored. This way no one would need to worry that he commit the changes in his examples

examples/example_setup/setup.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+func (c *SetupConfig) Validate() error {
+	if _, err := defs.ParseBSVNetworkStr(string(c.Network)); err != nil {
+		return fmt.Errorf("invalid BSV network: %w", err)
+	}
+	return nil
+}
+
+func loadConfig(configFile string) (*SetupConfig, error) {
```
dorzepowski: This can be a subject for different PR,

We have [config loader solution](https://github.com/bsv-blockchain/go-wallet-toolbox/blob/f55b7debc10a42c964f3e763d570aa59393de6e1/internal/config/loader.go#L35) that provides a standarized way of setup loading.

examples/example_setup/setup.go
```diff
@@ -0,0 +1,114 @@
...
+	userWallet, err := wallet.New(s.Environment.BSVNetwork, &s.PrivateKey, storageClient)
+	if err != nil {
+		cleanup()
+		return nil, nil, fmt.Errorf("failed to create wallet: %w", err)
+	}
+
+	identityKey := s.IdentityKey.ToDERHex()
+	user, err := storageClient.FindOrInsertUser(ctx, identityKey)
+	if err != nil {
+		cleanup()
```
dorzepowski: nice catch with this ❤️

examples/wallet_examples/faucet_address/faucet_address.go
```diff
@@ -0,0 +1,12 @@
+package main
+
+import (
+	"github.com/4chain-ag/go-wallet-toolbox/examples/example_setup"
+)
+
+func main() {
```
dorzepowski, line 10: Maybe we should add a docs to each example, describing the idea of the example

examples/wallet_examples/faucet_address/faucet_address.go
```diff
@@ -0,0 +1,12 @@
+package main
+
+import (
+	"github.com/4chain-ag/go-wallet-toolbox/examples/example_setup"
+)
+
+func main() {
+	alice := example_setup.CreateAlice()
+
+	example_setup.FaucetAddress(alice)
```
dorzepowski, line 13: maybe it would be good to add some comment describing why it's hidden inside this function.
Something about that, it's custom way - not worth to follow - to make the FaucetAddress and FaucetInternalize smoothly work together but in better words 😄

examples/wallet_examples/faucet_internalize/faucet_internalize.go
```diff
@@ -0,0 +1,56 @@
...
+	alice := example_setup.CreateAlice()
+
+	aliceWallet, cleanup, err := alice.CreateWallet(ctx)
+	if err != nil {
+		panic(fmt.Errorf("failed to create Alice's wallet: %w", err))
+	}
+	defer cleanup()
+
+	if beef == "" {
+		beef, err = utils.WocAPIGetBeefForTX(string(defs.NetworkTestnet), *txID)
```
dorzepowski: you could use BSVNetwork as an argument - so the type conversion won't be needed here

examples/wallet_examples/faucet_internalize/faucet_internalize.go
```diff
@@ -0,0 +1,56 @@
...
+		}
+
+		show.Step("Alice", "Retrieving BEEF data for transaction")
+		show.Transaction(*txID)
+
+	}
+
+	show.Step("Alice", "Internalizing transaction from faucet")
+
+	err = example_setup.InternalizeFromFaucet(ctx, alice.Environment, beef, aliceWallet)
```
dorzepowski: similar to the FaucedAddress - it would be good to place some comment that it's custom solution to make faucet address and this example work smoothly together.

examples/utils/wocGetBeefFromTx.go
```diff
@@ -0,0 +1,30 @@
...
+	"fmt"
+	"io"
+	"net/http"
+)
+
+// WocAPIGetBeefForTX fetches a beef from the WhatsonChain API
+func WocAPIGetBeefForTX(network, txid string) (string, error) {
+	url := fmt.Sprintf("https://api.whatsonchain.com/v1/bsv/%s/tx/%s/beef", network, txid)
+
+	resp, err := http.Get(url)
```
dorzepowski: it would be nice, to use resty instead

examples/example_setup/internalize_tx.go
```diff
@@ -0,0 +1,58 @@
...
+	iar, err := wallet.InternalizeAction(ctx, internalizeArgs, "originator")
+	if err != nil {
+		return sdk.InternalizeActionResult{}, err
+	}
+
+	return *iar, nil
+}
+
+// InternalizeFromFaucet is a helper function to internalize a transaction from the faucet
+func InternalizeFromFaucet(ctx context.Context, env Environment, beefHex string, wallet sdk.Interface) error {
```
dorzepowski: env argument looks like not use

examples/example_setup/internalize_tx.go
```diff
@@ -0,0 +1,58 @@
...
+	"context"
+	"encoding/hex"
+	"fmt"
+
+	"github.com/4chain-ag/go-wallet-toolbox/examples/utils"
+	sdk "github.com/bsv-blockchain/go-sdk/wallet"
+)
+
+// InternalizeTxHandler is a helper function to internalize a transaction
+func InternalizeTxHandler(ctx context.Context, internalizeArgs sdk.InternalizeActionArgs, wallet sdk.Interface) (sdk.InternalizeActionResult, error) {
```
dorzepowski: this helper function is not needed, please inline it and remove.

examples/example_setup/example-config.yaml
```diff
@@ -0,0 +1,8 @@
+network: test
+server_url: "http://localhost:8100"
+alice:
+  identity_key: "020c0ca23c75f7312bad0c5d81bff858bdcf468d3ad69a60b46ae90cafef557b03"
```
dorzepowski: this is probably for other PR, but it would be good to have the keys generated (like it is in ts example)

examples/example_setup/faucet_address.go
dorzepowski, line 1: please add an `internal` package in `examples` directory
and move this directory to `examples/internal/example_setup`

examples/show/show.go
dorzepowski, line 1: please add an `internal` package in `examples` directory
and move this directory to `examples/internal/show`

examples/utils/keyDerivation.go
dorzepowski, line 1: please add an `internal` package in `examples` directory
and move this directory to `examples/internal/utils`

examples/example_setup/example-config.yaml
dorzepowski: please add a task for that

examples/example_setup/setup.go
cain4chain: ok yes makes sense. I will add this as part of the config loader task.

examples/internal/example_setup/setup.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/defs"
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/storage"
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/wallet"
+	"github.com/spf13/viper"
+)
+
+type Setup struct {
+	Environment Environment
+	IdentityKey ec.PublicKey
+	PrivateKey  ec.PrivateKey
```
dorzepowski: I just realize that you're using values for Public and Private keys, I'm not sure if this is best idea, those structs are rather accessed as pointer.

First drawback is that: instead of simply passing them to any method, One need to always get a pointer of it.

I would suggest to change keys to be pointers

IMHO good rule of thumb is to use as a field type:
1. struct as value - when:
   * it has good zero value 
   * it is small struct 
   * it is struct with that shouldn't be modified (config structs often falls into this and the one above case)
2. struct as pointer - when:
  * large structs
  * mutable 
  * recursive operations
  * optional fields

examples/internal/example_setup/setup.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	userWallet, err := wallet.New(s.Environment.BSVNetwork, &s.PrivateKey, storageClient)
+	if err != nil {
+		cleanup()
+		return nil, nil, fmt.Errorf("failed to create wallet: %w", err)
+	}
+
+	identityKey := s.IdentityKey.ToDERHex()
+	user, err := storageClient.FindOrInsertUser(ctx, identityKey)
```
dorzepowski: Sorry, I just realized, that this is actually not needed, because (in both go and ts): 
on the first call of the wallet to storage on any method wallet will call the FindOrInsertUser on the storage, to make storage ready for handling any user operation.

Therefore I would recommend to remove this call from the CreateWallet to not mislead the adopters that they must make this call by themselves.

examples/internal/example_setup/internalize_tx.go
```diff
@@ -0,0 +1,50 @@
...
+					DerivationPrefix:  string(paymentRemittance.DerivationPrefix),
+					DerivationSuffix:  string(paymentRemittance.DerivationSuffix),
+					SenderIdentityKey: paymentRemittance.SenderIdentityKey,
+				},
+			},
+		},
+		Description: "internalize from faucet",
+	}
+
+	fmt.Println(internalizeArgs)
```
dorzepowski: printing any args is good candidate for method in show package.

So it could also put some title what is it.

examples/internal/example_setup/internalize_tx.go
```diff
@@ -0,0 +1,50 @@
...
+
+	iar, err := wallet.InternalizeAction(ctx, internalizeArgs, "originator")
+	if err != nil {
+		fmt.Println(err)
+		return err
+	}
+
+	fmt.Println(*iar)
```
dorzepowski: maybe there could be a functions (two) in show package for printing success and error result of wallet call, which would accept the method name, args and result (in one function) or error (in second function)

examples/wallet_examples/faucet_internalize/faucet_internalize.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/defs"
+	"github.com/spf13/pflag"
+)
+
+// Add beef hex directly here or use the txid to get the beef hex from the API
+var beef = ""
+
+var txID = pflag.String("txid", "15f47f2db5f26469c081e8d80d91a4b0f06e4a97abcc022b0b5163ac5f6cc0c8", "pass the chosen txid or simply change the default value when running the example")
+
+func main() {
```
dorzepowski, line 22: please add the description also here in this comment.
please mention about faucet_address example.

dorzepowski (approved):

examples/wallet_examples/faucet_internalize/faucet_internalize.go
```diff
@@ -0,0 +1,59 @@
...
+
+	if beef == "" {
+		beef, err = utils.WocAPIGetBeefForTX(defs.NetworkTestnet, txID)
+		if err != nil {
+			panic(fmt.Errorf("failed to get beef for tx: %w", err))
+		}
+
+		show.Step("Alice", "Retrieving BEEF data for transaction")
+		show.Transaction(txID)
+
```
chris-4chain: unnecessary empty line

examples/internal/utils/key_derivation.go
```diff
@@ -0,0 +1,68 @@
...
+	var derivationSuffix []byte
+
+	if prefix == "" {
+		derivationPrefix, _ = base64.StdEncoding.DecodeString(defaultBase64Prefix)
+	} else {
+		derivationPrefix = []byte(prefix)
+	}
+
+	if suffix == "" {
+		derivationSuffix, _ = base64.StdEncoding.DecodeString(defaultBase64Suffix)
```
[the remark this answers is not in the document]
chris-4chain: since it's a utility for the "examples" - maybe just panic with descriptive error would be enough

chris-4chain (approved):

examples/internal/utils/key_derivation.go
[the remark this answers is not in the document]
dorzepowski: I would go with panic here, in case of error, so it will be easy to find the reason for this failure, especially as this is only utility for examples

PR #292, opened 2025-06-18
feat(#184): wallet create action with no send

dorzepowski:


pkg/wallet/internal/mapping/mapping_transaction.go
```diff
@@ -0,0 +1,213 @@
...
+	sourceTxID, err := chainhash.NewHashFromHex(it.SourceTxID)
+	if err != nil {
+		return nil, fmt.Errorf("cannot parse input %d source txid returned from storage: %w", it.Vin, err)
+	} else if sourceTxID == nil {
+		return nil, fmt.Errorf("cannot parsed input %d  source tx id is nil: %w", it.Vin, err)
+	}
+
+	if a.isInputFromArgs(it) {
+		return a.toTxInputFromArgs(it, sourceTxID, err)
+	}
```
chris-4chain: I think the error as an argument is mistakenly provided to `a.toTxInputFromArgs` - it is always `nil`, because of an early return couple lines above (`if err != nil {`)

pkg/wallet/internal/mapping/mapping_transaction.go
```diff
@@ -0,0 +1,213 @@
...
+		return nil, fmt.Errorf("cannot create unlocking template for input %d: %w", it.Vin, err)
+	}
+
+	return input, nil
+}
+
+func (a *CreateActionTransactionAssembler) toTxInputFromArgs(it *wdk.StorageCreateTransactionSdkInput, sourceTxID *chainhash.Hash, err error) (*transaction.TransactionInput, error) {
+	argsInput := a.args.Inputs[it.Vin]
+	if !argsInput.Outpoint.Txid.Equal(*sourceTxID) || argsInput.Outpoint.Index != it.SourceVout {
+		return nil, fmt.Errorf("unexpected input (outpoint: %s.%d) on index %d: %w", it.SourceTxID, it.SourceVout, it.Vin, err)
```
chris-4chain: It seems to me that the thing with the error-passed-as-argument is kinda overcomplicated. 
- It's not readable to me

pkg/wallet/internal/mapping/mapping_transaction.go
```diff
@@ -0,0 +1,213 @@
...
+	)
+	if err != nil {
+		return nil, fmt.Errorf("cannot create unlocking template for input %d: %w", it.Vin, err)
+	}
+
+	return input, nil
+}
+
+func (a *CreateActionTransactionAssembler) toTxInputFromArgs(it *wdk.StorageCreateTransactionSdkInput, sourceTxID *chainhash.Hash, err error) (*transaction.TransactionInput, error) {
+	argsInput := a.args.Inputs[it.Vin]
```
chris-4chain, line 148: Please make sure that the `it.Vin` is properly validated against negative values (`it.Vin` is a "signed" int)

pkg/internal/testabilities/asserttx/assertion_transaction.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+func (t *txAssertion) HasMinimalFee() BSVTransactionAssertion {
+	totalInputs, ok := t.totalInputSatoshis()
+	if !ok {
+		return t
+	}
+
+	size := t.tx.Size()
+	var expectedFee = must.ConvertToUInt64(to.IfThen(size%1000 == 0, size/1000).ElseThen(size/1000 + 1))
```
chris-4chain: `1000` - potential candidate for a local constant

pkg/internal/testabilities/asserttx/assertion_transaction.go
```diff
@@ -0,0 +1,148 @@
...
+
+type Index int
+
+func (i Index) MatchesOutput(index int, _ *transaction.TransactionOutput) bool {
+	return index == int(i)
+}
+
+func (i Index) MatchesInput(index int, _ *transaction.TransactionInput) bool {
+	return index == int(i)
+}
```
chris-4chain: This Matcher tools seem to be unused

pkg/wallet/internal/mapping/mapping_outpoints.go
```diff
@@ -0,0 +1,32 @@
...
+		return nil, fmt.Errorf("failed to prepare outpoints from indexes: %w", err)
+	}
+	return outpoints, nil
+}
+
+func indexToOutpointMapper(txID *chainhash.Hash) func(it int) (transaction.Outpoint, error) {
+	return func(it int) (transaction.Outpoint, error) {
+		index, err := to.UInt32(it)
+		if err != nil {
+			return transaction.Outpoint{}, fmt.Errorf("invalid no sand change output index: %w", err)
```
chris-4chain:
```suggestion
			return transaction.Outpoint{}, fmt.Errorf("invalid no send change output index: %w", err)
```

BUT:
Since the function can be used everywhere in the code - I would not add "no send change" text in the error. 

This should be rather added in error wrapping on some upper level.

pkg/wallet/internal/mapping/mapping_signable_transaction_result.go
```diff
@@ -0,0 +1,39 @@
...
+	txID := tx.TxID()
+
+	atomicBytes, err := beefutils.ToAtomicBEEFBytes(tx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create atomic tx bytes: %w", err)
+	}
+
+	result := &sdk.CreateActionResult{
+		// TODO: make an issue/PR, that it should be pointer
+		Txid: to.Value(txID),
```
chris-4chain, line 23: Why? :) Because it's optional?

pkg/wallet/internal/mapping/mapping_transaction.go
```diff
@@ -0,0 +1,213 @@
...
+func NewCreateActionTransactionAssembler(keyDeriver *wallet.KeyDeriver, createActionResult *wdk.StorageCreateActionResult, args wallet.CreateActionArgs) *CreateActionTransactionAssembler {
+	return &CreateActionTransactionAssembler{
+		keyDeriver:         keyDeriver,
+		createActionResult: createActionResult,
+		args:               args,
+		tx:                 &transaction.Transaction{},
+	}
+}
+
+func (a *CreateActionTransactionAssembler) Assemble() (*transaction.Transaction, error) {
```
chris-4chain, line 35: Where did the idea for the name "Assembler" come from? I'm just curious :)

pkg/wallet/internal/mapping/mapping_transaction.go
```diff
@@ -0,0 +1,213 @@
...
+		DerivationPrefix: a.createActionResult.DerivationPrefix,
+		DerivationSuffix: to.Value(it.DerivationSuffix),
+	}
+
+	err := keyID.Validate()
+	if err != nil {
+		return nil, fmt.Errorf("cannot create change locking script for output %d: %w", it.Vout, err)
+	}
+
+	lockingScript, err := brc29.Lock(a.keyDeriver, keyID, a.keyDeriver)
```
chris-4chain, line 212: TO be honest I didn't expect that the locking procedure was done ine `mapping` package. 

The whole "Assembler" doesn't fit to the `mapping` package to me.

pkg/wallet/internal/testabilities/fixture_wallet_faucet.go
```diff
@@ -0,0 +1,98 @@
...
+	targetPubKey, err := userWallet.GetPublicKey(t.Context(), sdk.GetPublicKeyArgs{
+		ForSelf: true,
+		EncryptionArgs: sdk.EncryptionArgs{
+			Counterparty: sdk.Counterparty{
+				Type:         sdk.CounterpartyTypeOther,
+				Counterparty: senderKey,
+			},
+			ProtocolID: sdk.Protocol{
+				SecurityLevel: 2,
+				Protocol:      "3241645161d8",
```
chris-4chain: Potential candidates for constants

pkg/wallet/wallet_rebuild_transaction.go
```diff
@@ -0,0 +1,19 @@
...
+	"fmt"
+
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/wallet/internal/mapping"
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/wdk"
+	"github.com/bsv-blockchain/go-sdk/transaction"
+	sdk "github.com/bsv-blockchain/go-sdk/wallet"
+)
+
+func (w *Wallet) rebuildTransaction(createActionResult *wdk.StorageCreateActionResult, args sdk.CreateActionArgs) (*transaction.Transaction, error) {
+	txAssempler := mapping.NewCreateActionTransactionAssembler(w.keyDeriver, createActionResult, args)
```
chris-4chain:
```suggestion
	txAssembler := mapping.NewCreateActionTransactionAssembler(w.keyDeriver, createActionResult, args)
```

pkg/wallet/wallet.go
```diff
@@ -181,6 +144,33 @@ func (w *Wallet) CreateAction(ctx context.Context, args sdk.CreateActionArgs, or
 		return nil, fmt.Errorf("invalid create action args: %w", err)
 	}
 
+	if wdkArgs.IsNewTx {
+		storageCreateActionResult, err := w.storage.CreateAction(ctx, wdkArgs)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create action: %w", err)
+		}
+
+		tx, err := w.rebuildTransaction(storageCreateActionResult, args)
```
chris-4chain: IMHO This name `rebuildTransaction` is kinda misleading because this transaction has never been built. So:
- `buildTransaction`
- `assembleTransaction`
- `*FromSpec`
- `*FromCreateActionResult`

pkg/wallet/wallet.go
```diff
@@ -181,6 +144,33 @@ func (w *Wallet) CreateAction(ctx context.Context, args sdk.CreateActionArgs, or
 		return nil, fmt.Errorf("invalid create action args: %w", err)
 	}
 
+	if wdkArgs.IsNewTx {
+		storageCreateActionResult, err := w.storage.CreateAction(ctx, wdkArgs)
```
chris-4chain, line 148: The code under `wdkArgs.IsNewTx` could be moved to a separate method.

pkg/internal/testabilities/asserttx/assertion_transaction.go
dorzepowski: yes, I forgot to remove it after realizing that for now it's not something that is needed.

pkg/wallet/internal/mapping/mapping_signable_transaction_result.go
dorzepowski: Because txID by default is a pointer - it makes it less smooth to use with integration with transaction API

pkg/wallet/internal/mapping/mapping_transaction.go
dorzepowski: so what's your proposition?

pkg/wallet/internal/testabilities/fixture_wallet_faucet.go
dorzepowski: I forgot that I already made something for that.

pkg/wallet/wallet.go
dorzepowski: Yes, I totally agree,
but soon I will proceed with broadcasting part, and then I will extract proper methods, because I don't know yet, what is really needed in which part.

pkg/internal/testabilities/asserttx/assertion_transaction.go
```diff
@@ -0,0 +1,133 @@
...
+	"testing"
+
+	"github.com/bsv-blockchain/go-sdk/transaction"
+	"github.com/go-softwarelab/common/pkg/must"
+	"github.com/go-softwarelab/common/pkg/to"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+const kilobyte = 1000
```
chris-4chain, line 13: `1024` 🤔 ?

pkg/internal/testabilities/asserttx/assertion_transaction.go
```diff
@@ -0,0 +1,133 @@
...
+	if !ok {
+		return t
+	}
+
+	assert.EqualValues(t, sats, int(totalInputs), "expect transaction to have total input value")
+	return t
+}
+
+func (t *txAssertion) Outputs() BSVOutputsAssertion {
+	return t
```
ac4ch, line 116:
```suggestion
	return &outputsAssertion{
		TB: t.TB,
		outputs: t.tx.Outputs,
	}
```

should not be like this ? since you saying txAssertion is a BSVOutputsAssertion ? but it is not ... hm ..

jakubmkowalski (approved):

pkg/internal/testabilities/asserttx/assertion_transaction.go
dorzepowski: no, it was intentional,
the whole api is driven by interfaces, so that you can call it only in specific way, 

But this doesn't suggest that you will always get a new struct. IMHO new struct should be returned when it is really needed and gives some benefits (especially when you need additional context for assertions), and this is not the case.

mgosek-4chain (approved):

chris-4chain (approved):

PR #295, opened 2025-06-20
feat: sync known transactions for user's backup (proven and req)

chris-4chain:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter


pkg/wdk/table_certificate_field.go
```diff
@@ -0,0 +1,24 @@
...
+	UpdatedAt     time.Time               `json:"updated_at"`
+	UserID        int                     `json:"userId"`
+	CertificateID uint                    `json:"certificateId"`
+	FieldName     string                  `json:"fieldName"`
+	FieldValue    string                  `json:"fieldValue"`
+	MasterKey     primitives.Base64String `json:"masterKey"`
+}
+
+// TableCertificateX extends TableCertificate with optional fields
+type TableCertificateX struct {
```
dorzepowski: I'm not sure if those structs really need to be moved to the separate file, especially as this one is using TableCertificate

pkg/internal/fixtures/default_request_sync_chunk_args.go
```diff
@@ -0,0 +1,31 @@
+package fixtures
+
+import (
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/wdk"
+)
+
+func DefaultRequestSyncChunkArgs(userIdentityKey, fromIdentityKey string) wdk.RequestSyncChunkArgs {
+	return wdk.RequestSyncChunkArgs{
+		FromStorageIdentityKey: fromIdentityKey,
+		ToStorageIdentityKey:   "to_storage",
```
dorzepowski: shouldn't this also be some how provided or be valid storage identity key?

pkg/internal/storage/database/models/output.go
```diff
@@ -13,7 +13,7 @@ type Output struct {
 	Vout          uint32 `gorm:"index"`
 	Satoshis      int64
 
-	LockingScript      *string `gorm:"type:string"`
+	LockingScript      []byte
```
dorzepowski, line 16: why not hex?

pkg/storage/internal/actions/list_actions_mapping.go
```diff
@@ -158,14 +158,19 @@ func (l *listActions) mapToWalletActionOutputs(outputs []*wdk.TableOutput, tags
 
 		// TODO: Implement tags mapping if needed, currently empty as CreateActionArgs does not support tags
 
+		lockingScript := ""
+		if o.LockingScript != nil {
```
dorzepowski: no need for this check now, we can simply inline the call to o.LockingScript.Hex()

mgosek-4chain (approved):

pkg/wdk/table_certificate_field.go
chris-4chain: I would like to have clear one-file-per-one-table structure.

pkg/internal/fixtures/default_request_sync_chunk_args.go
chris-4chain: This is provided by Manager/SyncToWriter algorithm
The `DefaultRequestSyncChunkArgs` is supposed to be used for tests **only**

About validation: There is no validation of this field in `TS` version. 
- I'm not sure if we want to add such check and pass a public-key-format only.
- But we can - 😄 - let me know

pkg/internal/storage/database/models/output.go
chris-4chain: This was a blocker for me, because during tests with TS I realized that it sends LockingScripts as numbers (`"ExplicitByteArray"`).
- Then I realized that I, some time ago, (by mistake) set for this field a `string` type (instead of `ExplicitByteArray`)
- So, In this PR I needed to fix this. 

BTW: []byte is even more aligned with our code base (there is less conversion needed and it is more performant) 
NOTE: JetBrains still shows it as hex content - so, no harm for readability and debugability 🙈

pkg/wdk/table_certificate_field.go
dorzepowski: so maybe the TableCertificateX structure is in wrong file?

pkg/internal/fixtures/default_request_sync_chunk_args.go
dorzepowski: 1. I know it's for tests only, but I like to have data in tests close to the thing that you can see in production use.
That's why, if this value should be set and should be identity key, I would put here some example key or use some already existing fixture const.

pkg/internal/storage/database/models/output.go
dorzepowski: It's db type so it doesn't need to be ExplicitByteArray, 
but if changing it to []bytes is reducing number of conversion then it seems ok.

pkg/wdk/table_certificate_field.go
chris-4chain: Yes, It should be rather in the `table_certificate` file as it is in `TS`: https://github.com/bitcoin-sv/wallet-toolbox/blob/0fe18f75d93eabb9d0b2b743b501f1d974974379/src/storage/schema/tables/TableCertificate.ts#L4

pkg/internal/fixtures/default_request_sync_chunk_args.go
chris-4chain: Done ✅

pkg/wdk/table_certificate_field.go
chris-4chain: Done ✅

dorzepowski (approved):

PR #296, opened 2025-06-20
feat: persist last block height to skip duplicate syncs

chris-4chain:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter


pkg/storage/internal/actions/synchronize_tx_statuses.go
```diff
@@ -57,15 +60,27 @@ func (s *synchronizeTxStatuses) SynchronizeTxStatuses(ctx context.Context) error
...
+	checkedForCurrentBlock, currentHeight, err := s.alreadyCheckedForCurrentBlock(ctx)
+	if err != nil {
+		s.logger.Warn("failed to check if already checked for this block", slog.Any("err", err))
+		// We still want to proceed with the synchronization, so we log the error and continue
+	} else {
+		if checkedForCurrentBlock {
+			return nil
+		}
+
+		if err = s.setLastBlockHeight(ctx, currentHeight); err != nil {
```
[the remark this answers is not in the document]
chris-4chain: That's a good point - but, I'm not sure if it's what we want. Let's discuss :)

mgosek-4chain (approved):

pkg/storage/provider_synchronize_tx_statuses_test.go
```diff
@@ -42,6 +49,108 @@ func TestSynchronizeTx(t *testing.T) {
...
+		)
+
+	// when:
+	err = activeStorage.SynchronizeTransactionStatuses(context.Background())
+
+	// then:
+	require.NoError(t, err)
+
+	// and:
+	// NOTE: The second call should also trigger a request to WhatsOnChain, because the block height is different
```
chris-4chain: Wrong comment - It should be rather: "request for the transaction"

pkg/internal/storage/repo/key_value.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+
+	return model.Value, true, nil
+}
+
+func (kv *KeyValue) Set(ctx context.Context, key string, value []byte) error {
+	err := kv.db.WithContext(ctx).Model(&models.KeyValue{}).
+		Where("key = ?", key).
+		Assign(models.KeyValue{Value: value}).
+		FirstOrCreate(&models.KeyValue{Key: key}).Error
```
dorzepowski: why not simply Save?

pkg/storage/provider_synchronize_tx_statuses_test.go
```diff
@@ -42,6 +49,110 @@ func TestSynchronizeTx(t *testing.T) {
...
+	givenProvider.ARC().WhenQueryingTx(txSpec.ID()).WillReturnWithMindedTx()
+
+	// NOTE: WhatsOnChain is unreachable, so we simulate that the chain tip is not available
+	_ = givenProvider.WhatsOnChain().WillBeUnreachable()
+
+	// when:
+	err := activeStorage.SynchronizeTransactionStatuses(context.Background())
+
+	// then:
+	require.NoError(t, err)
```
dorzepowski, line 76: maybe you should call for proven tx or something and make some assertion about transaction?

pkg/internal/testabilities/fixture_provider.go
```diff
@@ -71,10 +76,21 @@ func (p *providerFixture) withServices() ProviderFixture {
 
 	p.servicesFixture.ARC().IsUpAndRunning()
 
-	p.services = services.New(p.logger, defs.DefaultServicesConfig(p.network), services.WithRestyClient(p.servicesFixture.ARC().HttpClient()))
+	mockTransport := p.servicesFixture.Transport()
```
dorzepowski, line 79: I would prefer to add opts with transport in GivenServicesWithNetwork

This way, there would be single point of creation services in tests, so in case of changes of the New function, it would be easier to change this in all tests, by changing single places.

pkg/storage/internal/actions/synchronize_tx_statuses.go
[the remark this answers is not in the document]
dorzepowski: I just want to add same comment 😀
We should probably update the last synced block at the end.

pkg/storage/internal/actions/repos.interface.go
```diff
@@ -56,3 +56,8 @@ type ProvenTxRepo interface {
 	SetStatusForProvenTxAboveAttempts(ctx context.Context, attempts uint64, status wdk.ProvenTxReqStatus) error
 	FindProvenTxRawTXs(ctx context.Context, txIDs []string) (map[string][]byte, error)
 }
+
+type KeyValueRepo interface {
```
dorzepowski, line 60: maybe we could treat it (name it) here simply as Cache, not KeyValueRepo?

pkg/storage/internal/actions/synchronize_tx_statuses.go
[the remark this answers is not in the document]
chris-4chain: So 2:1 - bot+human convinced me 🙈

pkg/storage/internal/actions/repos.interface.go
chris-4chain: Hmmm - I wanted to have here `KeyValue` and (in the future) a `Cache` interface/implementations with TTL handling and stuff - on some upper layer.

pkg/internal/storage/repo/key_value.go
chris-4chain: Right :D 
But, note that this also works 😄 

Apparently, the `Save` alone doesn't work. But with `Where` clause, it's ok:
```golang
	err := kv.db.WithContext(ctx).Model(&models.KeyValue{}).
		Where("key = ?", key).
		Save(&models.KeyValue{
			Key:   key,
			Value: value,
		}).Error
```

pkg/storage/provider_synchronize_tx_statuses_test.go
chris-4chain: I cannot, because the `FindKnownTx` is in another PR already
Let's keep this comment open - and when I merge that PR I will add assertions you mention.

pkg/internal/testabilities/fixture_provider.go
chris-4chain: Sorry, don't understand - it's already a single place where it's defined.
Maybe it's not perfect but my point was to extend this with `siffer` - not to change how it was designed. 
However, of course I can to play around with it 😄 - but please explain your idea please.

pkg/storage/internal/actions/synchronize_tx_statuses.go
```diff
@@ -70,9 +70,14 @@ func (s *synchronizeTxStatuses) SynchronizeTxStatuses(ctx context.Context) error
...
 		}
 
-		if err = s.setLastBlockHeight(ctx, currentHeight); err != nil {
-			return err
-		}
+		defer func() {
+			if resultErr != nil {
+				return
+			}
+			if err := s.setLastBlockHeight(ctx, currentHeight); err != nil {
```
dorzepowski, line 77: it previously looks ugly,
and now it looks even worse 🙈 

why not simply put it at the end of this whole function, and simplify this If?

pkg/storage/provider_synchronize_tx_statuses_test.go
chris-4chain: Ok, Done ✅.
I used the `DBStateAssertion` introduced in the previous PR.

dorzepowski (approved):

PR #298, opened 2025-06-23
Feat/wallet list action

cain4chain:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter


pkg/storage/provider.go
```diff
@@ -381,7 +381,7 @@ func (p *Provider) ListActions(ctx context.Context, auth wdk.AuthID, args wdk.Li
 		return nil, ErrAuthorization
 	}
 
-	if err := validate.ListActionsArgs(&args); err != nil {
+	if err := validate.ValidListActionsArgs(&args); err != nil {
```
dorzepowski: I don't like this renaming,

In Go, package name is part of the whole name, therefore now it is:
`Validate Valid List Actions Args` while those args are simply ListActionsArgs

pkg/wallet/wallet_list_actions_test.go
```diff
@@ -0,0 +1,243 @@
...
+
+			// then:
+			then.Result(result).HasError(err)
+
+			then.Storage().HadNoInteraction()
+		})
+	}
+}
+
+func (s *WalletTestSuite) TestWalletListActions() {
```
dorzepowski, line 79: my proposition would be to rather:
1. have here one test for empty result
2. add call to list actions and assertions for it to tests for create action and internalize action - so that test would confirm that whole wallet as a solution is working as expected.

pkg/wallet/internal/mapping/mapping_list_actions.go
```diff
@@ -0,0 +1,187 @@
...
+		labelQueryMode := primitives.LabelQueryModeString("any")
+		result.LabelQueryMode = &labelQueryMode
+	default:
+		labelQueryMode := primitives.LabelQueryModeString("any")
+		result.LabelQueryMode = &labelQueryMode
+	}
+
+	if args.SeekPermission != nil {
+		seekPermission := primitives.BooleanDefaultTrue(*args.SeekPermission)
+		result.SeekPermissions = &seekPermission
```
dorzepowski: small simplification for this and all the next.
```suggestion
		result.SeekPermissions =  to.Ptr(primitives.BooleanDefaultTrue(*args.SeekPermission))
```

pkg/wallet/internal/mapping/mapping_list_actions.go
```diff
@@ -0,0 +1,187 @@
...
+	return &sdk.ListActionsResult{
+		TotalActions: result.TotalActions,
+		Actions:      slices.Map(result.Actions, mapListActionsAction),
+	}
+}
+
+// mapListActionsAction maps wdk.WalletAction to sdk.Action
+func mapListActionsAction(action wdk.WalletAction) sdk.Action {
+	var txidHash chainhash.Hash
+	if action.TxID != "" {
```
dorzepowski: this check can be skipped, because for empty string it still will work

pkg/wallet/internal/mapping/mapping_list_actions.go
```diff
@@ -0,0 +1,187 @@
...
+		if hash, err := chainhash.NewHashFromHex(action.TxID); err == nil {
```
dorzepowski: I would return an error if there is any, instead of swallowing it

pkg/wallet/internal/mapping/mapping_list_actions.go
```diff
@@ -0,0 +1,187 @@
...
+			txidHash = *hash
+		}
+	}
+
+	if action.Satoshis < 0 {
+		panic("negative satoshis value encountered: data integrity issue")
```
dorzepowski: use `satoshis, err := to.UInt64(action.Satoshis)` and return wrapped errror if there is any

pkg/wallet/internal/mapping/mapping_list_actions.go
```diff
@@ -0,0 +1,187 @@
...
+// MapListActionsResult maps *wdk.ListActionsResult to *sdk.ListActionsResult
+func MapListActionsResult(result *wdk.ListActionsResult) *sdk.ListActionsResult {
+	return &sdk.ListActionsResult{
+		TotalActions: result.TotalActions,
+		Actions:      slices.Map(result.Actions, mapListActionsAction),
+	}
+}
+
+// mapListActionsAction maps wdk.WalletAction to sdk.Action
+func mapListActionsAction(action wdk.WalletAction) sdk.Action {
```
dorzepowski: You can use `slices.MapOrError` to allow for returning an error from this function

pkg/wallet/internal/mapping/mapping_list_actions.go
```diff
@@ -0,0 +1,187 @@
...
+	case "unproven":
+		return sdk.ActionStatusUnproven
+	case "unsigned":
+		return sdk.ActionStatusUnsigned
+	case "nosend":
+		return sdk.ActionStatusNoSend
+	case "nonfinal":
+		return sdk.ActionStatusNonFinal
+	default:
+		return sdk.ActionStatusUnprocessed
```
dorzepowski: shouldn't we return an error here, that unknown status was received.

pkg/wallet/internal/mapping/mapping_list_actions.go
```diff
@@ -0,0 +1,187 @@
...
+
+// mapActionInput maps wdk.WalletActionInput to sdk.ActionInput
+func mapActionInput(input wdk.WalletActionInput) sdk.ActionInput {
+	result := sdk.ActionInput{
+		SourceSatoshis:   input.SourceSatoshis,
+		InputDescription: input.InputDescription,
+		SequenceNumber:   input.SequenceNumber,
+	}
+
+	parts := strings.Split(input.SourceOutpoint, ":")
```
dorzepowski: I'm pretty sure, it's not separated by ":" but rather by "."

pkg/wallet/internal/mapping/mapping_list_actions.go
```diff
@@ -0,0 +1,187 @@
...
+	if len(parts) == 2 {
```
dorzepowski: there is transaction.OutpointFromString method, maybe it's better to use it

pkg/wallet/internal/mapping/mapping_list_actions.go
```diff
@@ -0,0 +1,187 @@
...
+				}
+			}
+		}
+	}
+
+	if input.SourceLockingScript != "" {
+		if lockingScript, err := script.NewFromHex(input.SourceLockingScript); err == nil {
+			result.SourceLockingScript = lockingScript.Bytes()
+		}
+	}
```
dorzepowski, line 177: maybe all those conversions into script bytes could be changed into simple function

like `result.SourceLockingScript, err = scriptBytes(input.SourceLockingScript)

pkg/wallet/internal/mapping/mapping_list_actions.go
```diff
@@ -0,0 +1,187 @@
...
+				result.SourceOutpoint = transaction.Outpoint{
+					Txid:  *txidHash,
+					Index: uint32(vout),
+				}
+			}
+		}
+	}
+
+	if input.SourceLockingScript != "" {
+		if lockingScript, err := script.NewFromHex(input.SourceLockingScript); err == nil {
```
dorzepowski: please, wrap and return an error, instead of swallowing it

pkg/wallet/internal/mapping/mapping_list_actions.go
```diff
@@ -34,68 +35,77 @@ func MapListActionsArgs(args sdk.ListActionsArgs) wdk.ListActionsArgs {
...
-	return &sdk.ListActionsResult{
-		TotalActions: result.TotalActions,
-		Actions:      slices.Map(result.Actions, mapListActionsAction),
+func MapListActionsResult(result *wdk.ListActionsResult) (*sdk.ListActionsResult, error) {
+	actions, err := slices.MapOrError(result.Actions, mapListActionsAction)
+	if err != nil {
+		return nil, fmt.Errorf("failed to map actions: %w", err)
+	}
+
+	if result.TotalActions > math.MaxUint32 {
```
dorzepowski: ### very minor
to.UInt32FromUnsigned(result.TotalActions) could be used instead

pkg/wallet/internal/mapping/mapping_list_actions.go
```diff
@@ -34,68 +35,77 @@ func MapListActionsArgs(args sdk.ListActionsArgs) wdk.ListActionsArgs {
...
-	if action.TxID != "" {
-		if hash, err := chainhash.NewHashFromHex(action.TxID); err == nil {
-			txidHash = *hash
-		}
+
+	hash, err := chainhash.NewHashFromHex(action.TxID)
+	if err != nil {
+		return sdk.Action{}, fmt.Errorf("failed to convert txid to hash: %w", err)
+	}
+	txidHash = *hash
```
dorzepowski: no need for this variable, you can simply assign *hash to Txid field of sdk.Action

pkg/wallet/internal/mapping/mapping_list_actions.go
```diff
@@ -105,29 +115,40 @@ func mapListActionsAction(action wdk.WalletAction) sdk.Action {
...
+		return "", fmt.Errorf("unknown action status: %s", status)
+	}
+}
+
+// scriptBytes converts a hex string to script bytes, returns nil if conversion fails
+func scriptBytes(hexString string) []byte {
+	if hexString == "" {
+		return nil
+	}
+	if script, err := script.NewFromHex(hexString); err == nil {
```
dorzepowski: please, return the error also if there is one.

pkg/wallet/internal/mapping/mapping_list_actions.go
```diff
@@ -138,29 +159,14 @@ func mapActionInput(input wdk.WalletActionInput) sdk.ActionInput {
...
-	if len(parts) == 2 {
-		if txidHash, err := chainhash.NewHashFromHex(parts[0]); err == nil {
-			if vout, err := strconv.ParseUint(parts[1], 10, 32); err == nil {
-				result.SourceOutpoint = transaction.Outpoint{
-					Txid:  *txidHash,
-					Index: uint32(vout),
-				}
-			}
+	if input.SourceOutpoint != "" {
+		if outpoint, err := transaction.OutpointFromString(input.SourceOutpoint); err == nil {
```
dorzepowski: return the error if there is one

pkg/wallet/wallet_list_actions_test.go
```diff
@@ -104,10 +96,11 @@ func (s *WalletTestSuite) TestWalletListActions() {
 		// then:
 		assert.NoError(t, err)
 		require.NotNil(t, result)
-		assert.GreaterOrEqual(t, int(result.TotalActions), 0, "Should have non-negative total actions")
+		assert.Equal(t, uint32(0), result.TotalActions, "Should have zero total actions")
+		assert.Empty(t, result.Actions, "Actions list should be empty")
 	})
 
-	s.Run("list actions with all includes", func() {
+	s.Run("list actions after create action", func() {
```
dorzepowski: this isn't what I mean 🙈 

There are already tests for CreateAction, just add another `when:` `then:` section there that will check the list actions result after that call

pkg/wallet/wallet_list_actions_test.go
```diff
@@ -118,57 +111,34 @@ func (s *WalletTestSuite) TestWalletListActions() {
...
+		assert.Equal(t, createArgs.Description, action.Description)
+		assert.Equal(t, createArgs.Version, action.Version)
+		assert.Equal(t, createArgs.LockTime, action.LockTime)
+		assert.ElementsMatch(t, createArgs.Labels, action.Labels)
+		assert.NotNil(t, action.Inputs, "Inputs should be included")
+		assert.NotNil(t, action.Outputs, "Outputs should be included")
 	})
 
-	s.Run("list actions with label query mode all", func() {
+	s.Run("list actions after internalize action", func() {
```
dorzepowski: Same as for create action.

Add another `when:` `then` section to create action tests.

pkg/wallet/wallet_list_actions_test.go
cain4chain: ahhh ok, i got it and will fix thanks :D

pkg/wallet/internal/mapping/mapping_list_actions.go
mgosek-4chain: As @dorzepowski  mentioned in the previous comment, we should return the missing error.

pkg/wallet/internal/mapping/mapping_list_actions.go
```diff
@@ -0,0 +1,213 @@
...
+
+	if args.IncludeOutputLockingScripts != nil {
+		result.IncludeOutputLockingScripts = to.Ptr(primitives.BooleanDefaultFalse(*args.IncludeOutputLockingScripts))
+	}
+
+	return result
+}
+
+// MapListActionsResult maps *wdk.ListActionsResult to *sdk.ListActionsResult
+func MapListActionsResult(result *wdk.ListActionsResult) (*sdk.ListActionsResult, error) {
```
chris-4chain, line 69: The mapping should not be necessary, because the storage should return `sdk.ListActionsResult` already. 
- But this is not possible right now, because of the issue with "negative" satoshis value. 
- go-sdk guys are going to fix this - then we will be able to proceed

pkg/wallet/wallet.go
```diff
@@ -188,8 +188,27 @@ func (w *Wallet) AbortAction(ctx context.Context, args sdk.AbortActionArgs, orig
 
 // ListActions lists all transactions matching the specified labels.
 func (w *Wallet) ListActions(ctx context.Context, args sdk.ListActionsArgs, originator string) (*sdk.ListActionsResult, error) {
-	// TODO implement me
-	panic("implement me")
+	if err := validate.Originator(originator); err != nil {
+		return nil, fmt.Errorf("invalid originator: %w", err)
+	}
+
+	wdkArgs := mapping.MapListActionsArgs(args)
```
chris-4chain, line 195: This mapping would not be necessary when we switched to sdk's version of the args type.

Maybe - in the meantime - when we're waiting for go-sdk to fix the result struct - you could try to switch the storage method to use the sdk version.
- ... and remove the wdk.ListActionsArgs - as at this point it won't be used anywhere

pkg/wallet/internal/mapping/mapping_list_actions.go
```diff
@@ -90,7 +90,7 @@ func mapListActionsAction(action wdk.WalletAction) (sdk.Action, error) {
 		return sdk.Action{}, fmt.Errorf("failed to convert txid to hash: %w", err)
 	}
 
-	satoshis, err := to.UInt64(action.Satoshis)
+	satoshis, err := to.Int64(action.Satoshis)
```
chris-4chain: No need to do a conversion - it's already in `int64`

pkg/wallet/wallet_list_actions_test.go
```diff
@@ -0,0 +1,102 @@
...
+
+			// then:
+			then.Result(result).HasError(err)
+
+			then.Storage().HadNoInteraction()
+		})
+	}
+}
+
+func (s *WalletTestSuite) TestWalletListActions() {
```
chris-4chain, line 79: Please also add testcases for given:
1. InternalizeAction + CreateAction
2. n*InternalizeAction
3. n*(InternalizeAction + CreateAction)

cain4chain: Actually the tests are added to the createAction and InternalizeAction tests as per this comment
https://github.com/bsv-blockchain/go-wallet-toolbox/pull/298#discussion_r2161776727

Let me know what you would like to do though :)

chris-4chain: Ok I get it 👍 
So several listActions calls are missing:
- [`TestWalletInternalizeAction->basket insertion protocol`](https://github.com/bsv-blockchain/go-wallet-toolbox/blob/573a1e2e04b561a9e205cf8a724a8aaa5aea581a/pkg/wallet/wallet_internalize_action_test.go#L140)
- [`TestWalletCreateActionSuccess->return signable transaction when sign&process is false`](https://github.com/bsv-blockchain/go-wallet-toolbox/blob/c10c969644f40c08c19957fef1da878c5dc643ff/pkg/wallet/wallet_create_action_test.go#L103)

cain4chain: Thanks will update those two.

pkg/wallet/wallet_create_action_test.go
```diff
@@ -144,6 +144,17 @@ func (s *WalletTestSuite) TestWalletCreateActionSuccess() {
...
+		// when: verify that ListActions shows the created action
+		listArgs := fixtures.DefaultWalletListActionsArgsWithIncludes()
+		listArgs.Labels = args.Labels
+		listResult, listErr := aliceWallet.ListActions(t.Context(), listArgs, fixtures.DefaultOriginator)
+
+		// then:
+		assert.NoError(t, listErr)
+		require.NotNil(t, listResult)
+		assert.Equal(t, uint32(1), listResult.TotalActions, "Should have one action after create")
+		require.Len(t, listResult.Actions, 1)
```
chris-4chain: The fields of `listResult.Actions[0]` should be asserted - at least the crucial ones.
Similar comment for other testcases.

chris-4chain (commented):
I created a suggestion-PR, if you find it ok, please merge it into this one.
https://github.com/bsv-blockchain/go-wallet-toolbox/pull/316

chris-4chain (approved):

PR #307, opened 2025-06-24
feat: create WoC broadcaster fixing outputs

ac4ch:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter


pkg/services/internal/whatsonchain/broadcast.go
```diff
@@ -26,60 +25,114 @@
...
 }
 
-// broadcast submits a raw transaction to WhatsOnChain.
+type txBroadcastResponse struct {
+	TxID string `json:"txid"`
+}
+
+type txInfoResponse struct {
+	Blockhash   string `json:"blockhash"`
+	Blockheight int64  `json:"blockheight"`
```
chris-4chain:
```suggestion
	BlockHeight int64  `json:"blockheight"`
```

pkg/services/internal/whatsonchain/service.go
```diff
@@ -172,81 +172,32 @@ func (woc *WhatsOnChain) PostBEEF(ctx context.Context, beef *transaction.Beef, t
...
+	rawTxs, err := extractRawTransactions(beef, txIDs)
+	if err != nil {
+		return nil, err
 	}
 
-	txResults := make([]wdk.PostedTxID, 0, len(txIDs))
-	delayDuration := 3 * time.Second
-	var delay bool
+	var (
+		txResults     []wdk.PostedTxID
```
chris-4chain: Consider preallocating `txResults`

pkg/services/internal/whatsonchain/broadcast.go
```diff
@@ -26,60 +25,114 @@ type broadcastRequest struct {
...
+	}
+}
+
+func (woc *WhatsOnChain) tryFetchTxInfo(ctx context.Context, txid string, notes *[]string) (string, int64, error) {
+	info, err := woc.fetchTxInfo(ctx, txid)
+	if err != nil {
+		errMsg := fmt.Sprintf("failed to fetch tx info: %v", err)
+		*notes = append(*notes, errMsg)
+		return "", 0, fmt.Errorf(errMsg)
+	}
```
chris-4chain: Warn: printf: non-constant format string in call to fmt.Errorf
Besides, no error wrapping. 

I'm suggesting sth like this:
```suggestion
	if err != nil {
		*notes = append(*notes, fmt.Sprintf("failed to fetch tx info: %v", err))
		return "", 0, fmt.Errorf("failed to fetch tx info: %w", err)
	}
```

pkg/services/internal/whatsonchain/broadcast.go
```diff
@@ -26,60 +25,114 @@ type broadcastRequest struct {
...
 
-// broadcast submits a raw transaction to WhatsOnChain.
+type txBroadcastResponse struct {
+	TxID string `json:"txid"`
+}
+
+type txInfoResponse struct {
+	Blockhash   string `json:"blockhash"`
+	Blockheight int64  `json:"blockheight"`
+	Timestamp   string `json:"time"`
```
chris-4chain: Couldn't it be `time.Time`?
I know that it's not important right now, because this field is not even used anywhere - but, just asking :)

pkg/services/internal/whatsonchain/broadcast.go
```diff
@@ -26,60 +25,113 @@ type broadcastRequest struct {
...
+		DoubleSpend:  status == StatusDoubleSpend || status == StatusMissingInputs,
+		AlreadyKnown: status == StatusAlreadyBroadcasted,
+		BlockHash:    blockHash,
+		BlockHeight:  blockHeight,
+		Error:        firstNonNilError(fetchErr),
+		Notes:        convertNotes(notes),
+	}
+}
+
+func (woc *WhatsOnChain) tryFetchTxInfo(ctx context.Context, txid string, notes *[]string) (string, int64, error) {
```
ac4ch: i know its `notes *[]string` slice reference type, this way i am changing inside function and reflect it back to caller the change.

pkg/services/internal/whatsonchain/broadcast.go
```diff
@@ -26,60 +25,113 @@ type broadcastRequest struct {
...
+func (woc *WhatsOnChain) fetchTxInfo(ctx context.Context, txid string) (*txInfoResponse, error) {
+	var info txInfoResponse
+
+	req := woc.httpClient.
+		R().
+		SetContext(ctx).
+		SetResult(&info).
+		AddRetryCondition(retryOnTooManyRequestsStatus)
+
+	url := fmt.Sprintf("%s/tx/%s/info", woc.url, txid)
```
chris-4chain: Hmmm, Why cannot I find this `/tx/.../info` endpoint on WoC documentation?: https://docs.taal.com/core-products/whatsonchain/transaction

pkg/services/internal/whatsonchain/broadcast.go
chris-4chain: Already addressed 👍 
Resolving... :)

pkg/services/internal/whatsonchain/service.go
chris-4chain: Already addressed 👍
Resolving... :)

pkg/services/internal/whatsonchain/broadcast.go
```diff
@@ -26,60 +25,113 @@ type broadcastRequest struct {
...
+			TxID:   txid,
+			Error:  broadcastErr,
+			Notes:  convertNotes([]string{fmt.Sprintf("broadcast error: %v", broadcastErr)}),
 		}
 	}
-	return converted
+
+	resultStatus, notes := classifyBroadcastStatus(status)
+
+	blockHash, blockHeight, fetchErr := woc.tryFetchTxInfo(ctx, returnedTxid, &notes)
```
chris-4chain: Consider avoiding passing a pointer to notes and instead add the note on this level.  
```suggestion
	blockHash, blockHeight, fetchErr := woc.tryFetchTxInfo(ctx, returnedTxid)
	if fetchErr != nil {
		notes = append(notes, fetchErr.Error())
	}
```

pkg/services/internal/whatsonchain/service.go
```diff
@@ -197,56 +187,15 @@ func (woc *WhatsOnChain) PostBEEF(ctx context.Context, beef *transaction.Beef, t
 
 	for i, txid := range txIDs {
 		if delay {
-			select {
-			case <-ctx.Done():
-				return nil, fmt.Errorf("context canceled while waiting to broadcast transaction %s: %w", txid, ctx.Err())
-			case <-time.After(delayDuration):
+			if err := waitOrCancel(ctx, delayDuration, txid); err != nil {
```
chris-4chain: *Minor suggestion*
```go
if i != 0 {
    if err := waitOrCancel(ctx, delayDuration, txid); err != nil {
```
To me it's more straightforward and self-documenting but it's a minor.

pkg/services/internal/whatsonchain/broadcast.go
```diff
@@ -26,60 +25,136 @@ type broadcastRequest struct {
...
 }
 
-// broadcast submits a raw transaction to WhatsOnChain.
+type txBroadcastResponse struct {
+	TxID string `json:"txid"`
+}
+
+type txInfoResponse struct {
+	BlockHash   string `json:"blockhash"`
+	BlockHeight int64  `json:"blockheight"`
```
chris-4chain: I think, `json` tags are not necessary anymore for this struct.
Also I would rename the name of the struct to `txInfoResult` because it is not a "response" from WoC and rather our internal struct how we return a "result"

pkg/services/internal/whatsonchain/broadcast.go
```diff
@@ -26,60 +25,136 @@ type broadcastRequest struct {
...
+		DoubleSpend:  status == StatusDoubleSpend || status == StatusMissingInputs,
+		AlreadyKnown: status == StatusAlreadyBroadcasted,
+		BlockHash:    blockHash,
+		BlockHeight:  blockHeight,
+		Error:        firstNonNilError(fetchErr),
+		Notes:        convertNotes(notes),
+	}
+}
+
+func (woc *WhatsOnChain) tryFetchTxInfo(ctx context.Context, txid string) (string, int64, error) {
```
chris-4chain: For consistency (and small simplification) - you could also return here your struct with Hash&Height:

```golang
func (woc *WhatsOnChain) tryFetchTxInfo(ctx context.Context, txid string) (*txInfoResponse, error) {

// or after renaming:
func (woc *WhatsOnChain) tryFetchTxInfo(ctx context.Context, txid string) (*txInfoResult, error) {
```

pkg/services/internal/whatsonchain/utils.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+	return converted
+}
+
+func containsI(a string, b string) bool {
+	return strings.Contains(
+		strings.ToLower(a),
+		strings.ToLower(b),
+	)
+}
```
chris-4chain: Consider checking a bunch of strings in a single call:
```suggestion
func containsI(subject string, contains ...string) bool {
	subject = strings.ToLower(subject)

	for _, c := range contains {
		if strings.Contains(subject, strings.ToLower(c)) {
			return true
		}
	}

	return false
}
```

This way, you could simplify checking for multiple strings:
```golang
switch {
		case containsI(responseText, "already in mempool", "txn-already-known"):
```

pkg/services/internal/whatsonchain/postbeef_test.go
```diff
@@ -72,55 +72,68 @@ func TestWhatsOnChain_PostBEEF(t *testing.T) {
...
 			httpResponse:      "unexpected response code 500: unknown failure",
 			expectErrorResult: true,
-			expectTxID:        "f036a074ace427c5ebc3d0de89a63d4a5e7aabeff9fbc77435eb58f8dbfd59a9",
+			expectTxID:        calculatedTxID,
 		},
 	}
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
-			// Given:
```
chris-4chain, line 95: Why did you remove `given/when/then` comments?

pkg/services/internal/whatsonchain/postbeef_test.go
```diff
@@ -72,55 +72,68 @@ func TestWhatsOnChain_PostBEEF(t *testing.T) {
...
 			fixture.WhatsOnChain().WillRespondWithBroadcast(tc.httpStatus, tc.httpResponse, nil)
```
chris-4chain: Do we have a case when the third argument (err) is not nil?
Maybe it would be beneficial to have one... or remove the unused argument.

pkg/services/internal/whatsonchain/postbeef_test.go
```diff
@@ -72,55 +72,68 @@ func TestWhatsOnChain_PostBEEF(t *testing.T) {
...
-			// When:
 			result, err := woc.PostBEEF(t.Context(), beef, []string{calculatedTxID})
 
-			// Then:
 			require.NoError(t, err)
 			require.NotNil(t, result)
 			require.Len(t, result.TxIDResults, 1)
 
-			assertTxIDResult(t, result.TxIDResults[0], tc.expectTxID, tc.expectErrorResult, tc.expectDoubleSpend, tc.expectAlreadyKnown, tc.expectNotes)
+			assertTxIDResult(t, result.TxIDResults[0], tc)
```
chris-4chain, line 127: Why not inline those assertions?

```golang
            got := result.TxIDResults[0]
			require.Equal(t, tc.expectTxID, got.TxID)

			if tc.expectErrorResult {
				require.Equal(t, wdk.PostedTxIDResultError, got.Result)
			} else {
				require.NotEqual(t, wdk.PostedTxIDResultError, got.Result)
			}
// (...)
```

pkg/services/internal/whatsonchain/postbeef_test.go
```diff
@@ -72,55 +72,68 @@ func TestWhatsOnChain_PostBEEF(t *testing.T) {
...
+	expectAlreadyKnown bool
+	expectNotes        []string
+}) {
+	require.Equal(t, tc.expectTxID, got.TxID)
 
-	if expectErrorResult {
+	if tc.expectErrorResult {
 		require.Equal(t, wdk.PostedTxIDResultError, got.Result)
 	} else {
 		require.NotEqual(t, wdk.PostedTxIDResultError, got.Result)
```
chris-4chain, line 149: Could you also check the `.Error` field?
```golang
		require.Nil(t, got.Error)
```

pkg/services/internal/whatsonchain/postbeef_test.go
```diff
@@ -72,55 +72,68 @@ func TestWhatsOnChain_PostBEEF(t *testing.T) {
...
+	name               string
+	httpStatus         int
+	httpResponse       string
+	expectTxID         string
+	expectErrorResult  bool
+	expectDoubleSpend  bool
+	expectAlreadyKnown bool
+	expectNotes        []string
+}) {
+	require.Equal(t, tc.expectTxID, got.TxID)
```
chris-4chain, line 144: `BlockHeight` & `BlockHash` fields are not verified.

pkg/services/internal/whatsonchain/postbeef_test.go
```diff
@@ -72,55 +72,68 @@ func TestWhatsOnChain_PostBEEF(t *testing.T) {
...
+	for _, expectedNote := range tc.expectNotes {
+		found := false
+		for _, actualNote := range got.Notes {
+			if strings.Contains(actualNote.What, expectedNote) {
+				found = true
+				break
+			}
+		}
+		require.True(t, found, "expected note to contain: %q", expectedNote)
 	}
```
chris-4chain: Just saying :)

```suggestion
	strNotes := slices.Map(got.Notes, func (note wdk.ReqHistoryNote) string {
		return note.What
	})
	for _, expectedNote := range tc.expectNotes {
		assert.Contains(t, strNotes, expectedNote)
	}
```

BTW - `assert` doesn't break the test execution - and I think it fits better to check an array of expectedNotes

pkg/services/internal/whatsonchain/postbeef_test.go
```diff
@@ -13,7 +14,6 @@ import (
 )
 
 func TestWhatsOnChain_PostBEEF(t *testing.T) {
```
chris-4chain, line 18: Hmmm.. the tests execute significantly slower than other tests in this project
- Maybe it's because of the `time.Sleep(3*time.Seconds)` in the `PostBEEF` function. 

I know that it would require more work - but please consider adding:
1. New config param to `defs/whats_on_chain.go` like `BroadcastDelay`
2. Remember to set the default to 3 secs
3. Use the config in `woc/PostBEEF()` method. 

This way - for test purposes we could set this value to extremely low value - to not slow down the tests

... And also - this way, we allow user to customize it which also can be beneficial .

ac4ch: nah more was due to call the status endpoint .. but i also addressed this delay as it was good suggestion to not do it in tests....

pkg/services/internal/whatsonchain/postbeef_test.go
ac4ch: :( i know diff between require and assert ... added this suggestion to the test too :)

pkg/services/internal/whatsonchain/postbeef_test.go
ac4ch: usually helper functions are the better solution since they improve readability of the test and is easier to maintain this ... i would prefer to keep it this helper func...

pkg/services/internal/whatsonchain/postbeef_test.go
```diff
@@ -72,28 +84,62 @@ func TestWhatsOnChain_PostBEEF(t *testing.T) {
...
 		},
 	}
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
 			// Given:
-			fixture.WhatsOnChain().WillRespondWithBroadcast(tc.httpStatus, tc.httpResponse, nil)
+			fixture.WhatsOnChain().WillRespondWithBroadcast(tc.httpStatus, tc.httpResponse)
+
+			fixture.WhatsOnChain().Transport().RegisterResponder("POST",
```
chris-4chain: It would be good to hide this under a testability method like:
```golang
fixture.WhatsOnChain().WillRespondOnStatus(...)
```

pkg/defs/whats_on_chain.go
```diff
@@ -10,6 +10,7 @@ type WhatsOnChain struct {
 	APIKey            string          `mapstructure:"api_key"`
 	BSVExchangeRate   BSVExchangeRate `mapstructure:"bsv_exchange_rate"`
 	BSVUpdateInterval *time.Duration  `mapstructure:"bsv_update_interval"`
+	BroadcastDelay    *time.Duration  `mapstructure:"broadcast_delay"`
```
chris-4chain: Please make this "not-optional" (without pointer)
- and add default value here: https://github.com/bsv-blockchain/go-wallet-toolbox/blob/b8746f6df2aa9512856107b31b4d1d8ae7ea1586/pkg/defs/services.go#L76-L83

pkg/services/internal/whatsonchain/service.go
```diff
@@ -51,13 +52,19 @@ func New(httpClient *resty.Client, logger *slog.Logger, network defs.BSVNetwork,
 		SetLogger(logging.RestyAdapter(logger)).
 		SetDebug(logger.Enabled(context.Background(), slog.LevelDebug))
 
+	delay := 3 * time.Second
+	if config.BroadcastDelay != nil {
+		delay = *config.BroadcastDelay
+	}
```
chris-4chain: This wouldn't be required if you addressed my suggestion above ☝️

pkg/services/internal/whatsonchain/utils.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+	return nil
+}
+
+// WithBroadcastDelay allows setting custom broadcast delay for WhatsOnChain service
+func WithBroadcastDelay(delay time.Duration) func(*WhatsOnChain) {
+	return func(woc *WhatsOnChain) {
+		woc.broadcastDelay = delay
+	}
+}
```
chris-4chain: If it's used/applicable only in fixtures, please move this to `fixture_woc.go`

chris-4chain (changes requested):
Code looks good 👍 
But manual tests are required. I'm leaving "request changes" to prevent from auto merge.

chris-4chain (approved):

PR #309, opened 2025-06-26
feat(#160): implement BHS service to fetch longest block header tip

### LINKED ISSUE #160: [FEATURE] create block-headers-service chaintracker

mgosek-4chain:
## Linked Issues / Tickets
Closes #160 

## Testing Procedure

- [X] All tests pass locally

## Checklist:

- [X] I have run the linter


pkg/internal/testabilities/testservices/fixture_bhs.go
```diff
@@ -0,0 +1,113 @@
...
+	)
+}
+
+func (b *bhsFixture) OnLongestTipBlockHeaderResponseWith(opts ...LongestChainTipOptions) {
+	for _, o := range opts {
+		o(b.longestChainTip)
+	}
+}
+
+var bhsAnyEndpointRegexFixture = regexp.MustCompile(fmt.Sprintf(`^http:\/\/%s(?:\/.*)?$`, defs.BHSTestURL[7:]))
```
ac4ch: if i am correct you could simply do
```suggestion
var bhsAnyEndpointRegexFixture = regexp.MustCompile(`^http://(?:localhost:\d+)(?:/.*)?$`)
```
what is this magic number 7 ?

pkg/internal/testabilities/testservices/fixture_bhs.go
```diff
@@ -0,0 +1,113 @@
...
+		longestChainTip: &longestChainTipResponse{
+			Height:        800000,
+			Hash:          "0000000000000000000a7b8c9d0e1f2a3b4c5d6e7f8a9b0c1d2e3f4a5b6c7d8e",
+			Version:       536870912,
+			MerkleRoot:    "3a4b5c6d7e8f90123456789abcdef0123456789abcdef0123456789abcdef01",
+			Timestamp:     1719427200,
+			Bits:          386136923,
+			Nonce:         2083236893,
+			PreviousBlock: "00000000000000000008e7b8c6d5f4e3d2c1b0a987654321fedcba9876543210",
+		},
```
ac4ch:
```suggestion
		longestChainTip: defaultLongestChainTipResponse(),
```
consider adding helper func 
```go
func defaultLongestChainTipResponse() *longestChainTipResponse {
	return &longestChainTipResponse{
		Height:        800000,
		Hash:          "0000000000000000000a7b8c9d0e1f2a3b4c5d6e7f8a9b0c1d2e3f4a5b6c7d8e",
		Version:       536870912,
		MerkleRoot:    "3a4b5c6d7e8f90123456789abcdef0123456789abcdef0123456789abcdef01",
		Timestamp:     1719427200,
		Bits:          386136923,
		Nonce:         2083236893,
		PreviousBlock: "00000000000000000008e7b8c6d5f4e3d2c1b0a987654321fedcba9876543210",
	}
}
```

pkg/services/internal/bhs/service.go
```diff
@@ -0,0 +1,100 @@
...
+	var block dto.BlockHeader
+	url := fmt.Sprintf("%s/chain/tip/longest", b.cfg.URL)
+	res, err := b.
+		httpClient.
+		R().
+		SetContext(ctx).
+		SetResult(&block).
+		AddRetryCondition(func(res *resty.Response, err error) bool {
+			return res.StatusCode() == http.StatusTooManyRequests
+		}).
```
ac4ch:
```suggestion
		AddRetryCondition(retryOnTooManyRequestsStatus).
```
```go
// retryOnTooManyRequestsStatus defines retry condition for Too Many Requests response.
func retryOnTooManyRequestsStatus(res *resty.Response, err error) bool {
	return res.StatusCode() == http.StatusTooManyRequests
}
```

pkg/services/internal/bhs/service.go
```diff
@@ -0,0 +1,100 @@
...
+		httpClient: newRestyHTTPClient(httpClient, logger, network, config),
+		cfg:        &config,
+	}
+}
+
+func newRestyHTTPClient(httpClient *resty.Client, logger *slog.Logger, network defs.BSVNetwork, config defs.BHS) *resty.Client {
+	const (
+		retries         = 2
+		retriesWaitTime = 2 * time.Second // TODO: Move this section to shared pkg.
+	)
```
ac4ch:
```suggestion
	const (
	retries         = httpx.DefaultRetryCount
	retriesWaitTime = httpx.DefaultRetryInterval
)
```

```go
// pkg/services/internal/httpx/retry.go
const (
	DefaultRetryCount    = 2
	DefaultRetryInterval = 2 * time.Second
)
```

why TODO ? i think is simple change which can be done even now WDYT ?

pkg/internal/testabilities/testservices/fixture_bhs.go
```diff
@@ -0,0 +1,113 @@
...
+	b.transport.RegisterResponder(
+		http.MethodGet, defs.BHSTestURL+"/chain/tip/longest",
+		httpmock.NewJsonResponderOrPanic(http.StatusOK, longestChainTipResponse{}),
+	)
+}
+
+func (b *bhsFixture) OnLongestTipBlockHeaderResponseWith(opts ...LongestChainTipOptions) {
+	for _, o := range opts {
+		o(b.longestChainTip)
+	}
```
ac4ch, line 58:
```suggestion
	cloned := *b.longestChainTip
	for _, o := range opts {
		o(&cloned)
	}
	b.longestChainTip = &cloned
```
think is safer approach would be to clone the struct to avoid direct mutates ?

pkg/internal/testabilities/testservices/fixture_bhs.go
```diff
@@ -0,0 +1,113 @@
...
+	OnLongestTipBlockHeaderResponseWith(opts ...LongestChainTipOptions)
+}
+
+type bhsFixture struct {
+	testing.TB
+	transport       *httpmock.MockTransport
+	longestChainTip *longestChainTipResponse
+}
+
+func (b bhsFixture) WillRespondWithEmptyLongestTipBlockHeader() {
```
chris-4chain: Struct bhsFixture has methods on both value and pointer receivers. Such usage is not recommended by the Go Documentation.

```suggestion
func (b *bhsFixture) WillRespondWithEmptyLongestTipBlockHeader() {
```

pkg/services/services_find_chain_tip_header_test.go
```diff
@@ -41,15 +66,17 @@ func TestFindChainTipHeader(t *testing.T) {
...
 		require.Nil(t, actualBlock)
 	})
 
 	t.Run("return an error when all block header services are unreachable", func(t *testing.T) {
 		// given:
-
 		given := testservices.GivenServices(t)
-		target := given.WhatsOnChain().WillBeUnreachable()
+		target1 := given.BHS().WillBeUnreachable()
+		target2 := given.WhatsOnChain().WillBeUnreachable()
```
chris-4chain, line 79: Both WillBeUnreachable functions return an error which is then checked with require.ErrorIs. However, if both errors are identical (net.UnknownNetworkError("tests defined this endpoint as unreachable")), the checks:

```golang
require.ErrorIs(t, err, target1)
require.ErrorIs(t, err, target2)
```

will not effectively distinguish if both endpoints were indeed unreachable or only one, since they may return equivalent error instances (with the same message but not necessarily the same object).

pkg/services/internal/bhs/service.go
```diff
@@ -0,0 +1,100 @@
...
+			return res.StatusCode() == http.StatusTooManyRequests
+		}).
+		Get(url)
+
+	if err != nil {
+		return nil, fmt.Errorf("error while fetching block header from Block Headers Servcie API (URL: %s): %w", url, err)
+	}
+
+	if res.StatusCode() != http.StatusOK {
+		return nil, fmt.Errorf("unexpected response from  Block Headers Servcie API (URL: %s): status code %d", url, res.StatusCode())
```
chris-4chain: `Servcie` type - and in several places. :D I guess some copy-paste took place here 🙈

pkg/services/internal/bhs/internal/dto/dto.go
```diff
@@ -0,0 +1,31 @@
...
+	Hash          string `json:"hash"`
+	Version       uint32 `json:"version"`
+	MerkleRoot    string `json:"merkleRoot"`
+	Timestamp     uint64 `json:"creationTimestamp"`
+	Bits          uint64 `json:"bits"`
+	Nonce         uint32 `json:"nonce"`
+	PreviousBlock string `json:"prevBlockHash"`
+}
+
+func (b BlockHeader) IsZero() bool { return b == BlockHeader{} }
```
chris-4chain: Struct BlockHeader has methods on both value and pointer receivers. Such usage is not recommended by the Go Documentation.

pkg/services/internal/bhs/internal/dto/dto.go
```diff
@@ -0,0 +1,31 @@
+package dto
+
+import "github.com/4chain-ag/go-wallet-toolbox/pkg/wdk"
+
+type BlockHeader struct {
```
chris-4chain: Just saying, that in `BHS` repository there is actual struct defined. Maybe it would be good to align the types :shrug:

```golang
type TipResponse struct {
	Hash             string   `json:"hash"`
	Version          int32    `json:"version"`
	PreviousBlock    string   `json:"prevBlockHash"`
	MerkleRoot       string   `json:"merkleRoot"`
	Timestamp        uint32   `json:"creationTimestamp"`
	DifficultyTarget uint32   `json:"difficultyTarget"`
	Nonce            uint32   `json:"nonce"`
	Work             *big.Int `json:"work" swaggertype:"string"`
}
```

pkg/services/internal/bhs/service.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	err := network.Validate()
+	if err != nil {
+		panic(err)
+	}
+
+	err = config.Validate()
+	if err != nil {
+		panic(err)
```
chris-4chain: Consider add some context to the error:
```suggestion
panic(fmt.Errorf("invalid BHS configuration: %w", err))
```

pkg/services/internal/bhs/service.go
```diff
@@ -0,0 +1,100 @@
...
+	return &BlockHeadersService{
+		httpClient: newRestyHTTPClient(httpClient, logger, network, config),
+		cfg:        &config,
+	}
+}
+
+func newRestyHTTPClient(httpClient *resty.Client, logger *slog.Logger, network defs.BSVNetwork, config defs.BHS) *resty.Client {
+	const (
+		retries         = 2
+		retriesWaitTime = 2 * time.Second // TODO: Move this section to shared pkg.
```
chris-4chain: Are you going to address this TODO during this PR?

I think, this could be a config param for BHS config category you've just added.

pkg/defs/network.go
```diff
@@ -18,7 +21,7 @@ func ParseBSVNetworkStr(network string) (BSVNetwork, error) {
 
 // Validate checks if the value underneath is within valid BSVNetwork values.
 func (n BSVNetwork) Validate() error {
-	switch n {
+	switch BSVNetwork(strings.ToLower(string(n))) {
```
chris-4chain: I don't like this change. The enum should have accurate value at this point. 
If the "enum" value originates form some string (e.g. from config file) - there is `PatseBSVNetworkStr` to do the job.

pkg/services/internal/bhs/service.go
```diff
@@ -0,0 +1,100 @@
...
+
+func NewBlockHeadersService(httpClient *resty.Client, logger *slog.Logger, network defs.BSVNetwork, config defs.BHS) *BlockHeadersService {
+	if httpClient == nil {
+		panic("httpClient is required")
+	}
+	if logger == nil {
+		panic("logger is required")
+	}
+
+	err := network.Validate()
```
chris-4chain, line 58: To keep things consistent - could you add similar check to the `whatsonchain.New` method?

pkg/services/internal/bhs/service.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/internal/logging"
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/services/internal/bhs/internal/dto"
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/services/internal/httpx"
+	"github.com/4chain-ag/go-wallet-toolbox/pkg/wdk"
+	"github.com/go-resty/resty/v2"
+)
+
+const ServiceName = "BlockHeadersService"
+
+type BlockHeadersService struct {
```
chris-4chain, line 19: Could you look into the `bhs` package that we've made in `spv-wallet` - maybe you could find something interesting there. 

https://github.com/bitcoin-sv/spv-wallet/tree/8cf33cacb70d5906b45552a3ae1e9012874cd0a1/engine/chain/internal/bhs

pkg/defs/bhs.go
```diff
@@ -0,0 +1,24 @@
...
+type BHS struct {
+	URL    string `mapstructure:"url"`     // URL is the base endpoint of the BHS service.
+	APIKey string `mapstructure:"api_key"` // APIKey is the authentication key used for accessing the BHS service.
+}
+
+// Validate checks if the BHS configuration is valid.
+// It ensures that both the URL and APIKey fields are not empty.
+// Returns an error if any required field is missing.
+func (b *BHS) Validate() error {
+	if len(b.APIKey) == 0 {
```
chris-4chain: As far as I know, the `BHS` can work without `APIKey`.
I wonder if we should allow for this.

- https://github.com/bsv-blockchain/block-headers-service/blob/9e0dcc327a7578b802c31e39124e66f4b95fe8f0/config.example.yaml#L67
- https://github.com/bitcoin-sv/spv-wallet/blob/8d44c38b9a64d9881551c26501a9eb1dff178e85/engine/chain/internal/bhs/get_merkleroots.go#L29-L33

pkg/internal/testabilities/testservices/fixture_bhs.go
ac4ch: good catch missed this :)

pkg/internal/testabilities/testservices/fixture_bhs.go
mgosek-4chain: The idea was to allow customization of the response state using the functional options. Each test should work its own dataset, including an instance of the BHS fixture. Could you provide an example use case for your approach, and explain the benefits compared to the current one? 

@chris-4chain  🤔 ?

chris-4chain: For current state (how it's used right now), I don't see a reason to make a clone. 
They are tests anyway 🤷

pkg/services/internal/bhs/internal/dto/dto.go
chris-4chain: I suggested you to use the same response struct but not the imported one.
We should avoid making a dependency to BHS in `go.mod`

ac4ch (approved):

pkg/services/internal/bhs/internal/dto/dto.go
```diff
@@ -0,0 +1,26 @@
...
+type ExtendedTipStateResponse tips.TipStateResponse
+
+func (e *ExtendedTipStateResponse) IsZero() bool {
+	return *e == ExtendedTipStateResponse{}
+}
+
+func (e *ExtendedTipStateResponse) ConvertToChainBlockHeader() *wdk.ChainBlockHeader {
+	return &wdk.ChainBlockHeader{
+		ChainBaseBlockHeader: wdk.ChainBaseBlockHeader{
+			Version:      uint32(e.Header.Version), //nolint:gosec
```
chris-4chain:
```suggestion
			Version:      must.ConvertToUInt32(e.Header.Version),
```

pkg/services/internal/bhs/internal/dto/dto.go
```diff
@@ -0,0 +1,26 @@
...
+			PreviousHash: e.Header.PreviousBlock,
+			MerkleRoot:   e.Header.MerkleRoot,
+			Time:         uint64(e.Header.Timestamp),
```
chris-4chain:
```suggestion
			Time:         must.ConvertToUInt64(e.Header.Timestamp),
```

chris-4chain (approved):

PR #310, opened 2025-06-27
feat: backup users transactions

chris-4chain:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter


pkg/storage/internal/sync/chunk_processor.go
```diff
@@ -161,31 +200,171 @@ func (p *chunkProcessor) upsertProvenTx(chunkProvenTx *wdk.TableProvenTx) error
...
+		ProvidedBy:         chunkOutput.ProvidedBy,
+		Purpose:            chunkOutput.Purpose,
+		Type:               chunkOutput.Type,
+		SenderIdentityKey:  chunkOutput.SenderIdentityKey,
+		Tags:               nil, //TODO: Implement it along with tags backup support.
+		BasketName:         basketName,
+	}
+
+	if chunkOutput.Spendable && basketName != nil && *basketName == wdk.BasketNameForChange {
+		satoshis, err := to.UInt64(chunkOutput.Satoshis)
```
cain4chain, line 300: After the recent update should this be now int64?

pkg/storage/internal/sync/chunk_processor.go
```diff
@@ -104,22 +130,35 @@ func (p *chunkProcessor) mergeUser() error {
...
+	}
+
+	isNew, basketNumID, err := p.parent.repo.UpsertOutputBasketForSync(p.ctx, entity.OutputBasket{
+		Name:                    string(chunkBasket.Name),
+		UserID:                  p.user.ID,
+		CreatedAt:               chunkBasket.CreatedAt,
+		UpdatedAt:               chunkBasket.UpdatedAt,
+		NumberOfDesiredUTXOs:    chunkBasket.NumberOfDesiredUTXOs,
+		MinimumDesiredUTXOValue: chunkBasket.MinimumDesiredUTXOValue,
+	})
```
mgosek-4chain, line 149: It looks like we'll have here a panic when `User` field is nil.

pkg/storage/internal/sync/chunk_processor.go
chris-4chain: No, because the change was for "action" (which is basically a transaction entity)
- In this case, this is a UTXO which cannot have negative number.

mgosek-4chain (approved):

pkg/storage/internal/sync/chunk_processor.go
chris-4chain: User won't be `nil` because not-nil instance is passed to the constructor. 
BTW: This is a "internal" utility algorithm implementation - so I didn't want to be overly defensive programmer - that's why I didn't add additional check.

cain4chain (approved):

PR #327, opened 2025-07-04
feat(#299): createAction with signAndProcess of new tx.

dorzepowski:


pkg/internal/validate/valid_create_action_args.go
```diff
@@ -32,9 +32,11 @@ func ValidCreateActionArgs(args *wdk.ValidCreateActionArgs) error {
 		return fmt.Errorf("create action is meant to create a new transaction")
 	}
 
-	deducedIsSignAction := args.IsNewTx && (!args.Options.SignAndProcess.Value() || containsNilUnlockingScript(args.Inputs))
-	if args.IsSignAction != deducedIsSignAction {
-		return fmt.Errorf("inconsistent IsSignAction with IsNewTx and Options.SignAndProcess and Inputs.UnlockingScript")
+	deducedIsSignAction := args.IsNewTx && !args.Options.SignAndProcess.Value()
+	if deducedIsSignAction && !args.IsSignAction {
```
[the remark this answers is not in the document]
dorzepowski: Unfortunatelly at the storage side, we cannot make such expectation, there is a comment in the code about this.

jakubmkowalski (approved):

pkg/internal/validate/validate_not_delayed_process_action_result.go
```diff
@@ -0,0 +1,26 @@
...
+	allSend := seq.Every(seq.FromSlice(result.SendWithResults), func(it wdk.SendWithResult) bool {
+		return it.Status == wdk.SendWithResultStatusUnproven
+	})
+
+	if allSend {
+		return nil
+	}
+
+	// TODO: add more context to the error
+	//   and allow for making errors.Is on it
```
chris-4chain, line 24: It would be good to address this during this PR.

pkg/wallet/internal/actions/wallet_create_action.go
```diff
@@ -0,0 +1,134 @@
...
+	KeyDeriver *wallet.KeyDeriver
+	Storage    WalletStorageCreateAction
+	WalletOpts *wallet_opts.Opts
+
+	wdkArgs    wdk.ValidCreateActionArgs
+	originator string
+}
+
+func (a *CreateAction) CreateAction(ctx context.Context, args wallet.CreateActionArgs, originator string) (*wallet.CreateActionResult, error) {
+	// TODO: mapping.MapCreateActionArgs should handle known tx ids - needs some merging and validation of BEEF
```
chris-4chain, line 30: Shouldn't it be done during this PR?

pkg/wallet/internal/actions/wallet_create_action.go
```diff
@@ -0,0 +1,134 @@
...
+
+	if err := a.validate(); err != nil {
+		return nil, err
+	}
+
+	if a.isNotNewTX() {
+		return a.handleNotNewTX(ctx)
+	}
+	return a.handleNewTX(ctx, args)
+	// TODO: merge BEEF Party ??
```
chris-4chain, line 42: What about this `merge BEEF`?

pkg/wallet/internal/actions/wallet_create_action.go
```diff
@@ -0,0 +1,134 @@
...
+	processActionArgs := mapping.MapProcessActionArgs(txID, tx, createActionResult.Reference, a.wdkArgs)
+
+	processActionResult, err := a.Storage.ProcessAction(ctx, processActionArgs)
+	if err != nil {
+		return nil, fmt.Errorf("failed to process created action: %w", err)
+	}
+
+	err = a.validateProcessActionResult(processActionResult)
+	if err != nil {
+		return nil, err
```
chris-4chain, line 92: Shouldn't we do `abortAction` when the `wallet` decides that the response from storage was "invalid"?

- `abortAction` or any other "cleanup" method.

pkg/wallet/internal/mapping/mapping_process_action_args.go
```diff
@@ -0,0 +1,25 @@
...
+	"github.com/go-softwarelab/common/pkg/to"
+)
+
+func MapProcessActionArgs(txid *chainhash.Hash, tx *transaction.Transaction, reference string, wdkArgs wdk.ValidCreateActionArgs) wdk.ProcessActionArgs {
+	processActionArgs := wdk.ProcessActionArgs{
+		IsNewTx:    wdkArgs.IsNewTx,
+		IsSendWith: wdkArgs.IsSendWith,
+		IsNoSend:   wdkArgs.IsNoSend,
+		IsDelayed:  wdkArgs.IsDelayed,
+		SendWith:   to.IfThen(wdkArgs.IsSendWith, wdkArgs.Options.SendWith).ElseThen(nil),
```
chris-4chain, line 17: Why not inline?
```suggestion
		IsNoSend:   wdkArgs.IsNoSend,
		Reference:  &reference,
		TxID:       to.Ptr(primitives.TXIDHexString(txid.String())),
		RawTx:      tx.Bytes(),
```

pkg/wallet/internal/mapping/mapping_send_with_results.go
```diff
@@ -0,0 +1,49 @@
...
+	return result, nil
+}
+
+func mapSendWithResultWDKtoSDK(sendWithResult wdk.SendWithResult) (sdk.SendWithResult, error) {
+	txID, err := chainhash.NewHashFromHex(sendWithResult.TxID.String())
+	if err != nil {
+		return sdk.SendWithResult{}, fmt.Errorf("cannot restore tx id from hex %s, %w", sendWithResult.TxID, err)
+	}
+
+	status, err := mapSendWithResultStatusWDKtoSDK(sendWithResult.Status)
```
chris-4chain, line 27: Hmmm... just thinking

```suggestion
	status, err := SendWithResultStatusMapper.Map(status)
```

and:

```golang
type EnumMapper[From, To comparable] struct {
	lookup map[From]To
	name string
}

func NewStatusMapper[From, To comparable](name string, mappings map[From]To) *EnumMapper[From, To] {
	return &EnumMapper[From, To]{
		lookup: mappings,
		name:   name,
	}
}

func (mapper *EnumMapper[From, To]) Map(status From) (To, error) {
	if mappedStatus, ok := mapper.lookup[status]; ok {
		return mappedStatus, nil
	}
	return to.ZeroValue[To](), fmt.Errorf("unexpected %s: [%T] %v", mapper.name, status, status)
}

var SendWithResultStatusMapper = NewStatusMapper(
	"sendWithResult status from storage",
	map[wdk.SendWithResultStatus]sdk.ActionResultStatus{
		wdk.SendWithResultStatusUnproven: sdk.ActionResultStatusUnproven,
		wdk.SendWithResultStatusSending:   sdk.ActionResultStatusSending,
		wdk.SendWithResultStatusFailed:    sdk.ActionResultStatusFailed,
	},
)

```

pkg/wallet/internal/testabilities/fixture_wallet.go
```diff
@@ -4,63 +4,69 @@ import (
...
+		usersSetups:    make(map[testusers.User]*userWalletSetup),
+		usersFaucets:   make(map[string]*faucetFixture),
+		cleanupFuncs:   []func(){storageCleanup},
+		storageFixture: storageFixture,
 	}
+
+	cleanup = func() {
+		for i := len(w.cleanupFuncs) - 1; i >= 0; i-- {
+			w.cleanupFuncs[i]()
+		}
```
chris-4chain:
```suggestion
		for cleanupFunc := range seq.FromSliceReversed(w.cleanupFuncs) {
			cleanupFunc()
		}
```

pkg/wallet/wallet.go
```diff
@@ -49,15 +50,17 @@ func New[KeySource PrivateKeySource](chain defs.BSVNetwork, keySource KeySource,
 
 	storageManager := storage.NewWalletStorageManager(keyDeriver.IdentityKey().ToDERHex(), activeStorage)
 
+	opts := &wallet_opts.Opts{
```
chris-4chain, line 53: Hmmm... the `"Opts"` name suggests that it is a vararg options pattern `New(sth, opts ...Opts)`. But here, some other pattern is used.

Maybe some other name :shrug:, like `Configuration/Conifg`, `Flags`, `Settings`

pkg/wallet/wallet_create_action_test.go
```diff
@@ -255,16 +257,159 @@ func (s *WalletTestSuite) TestWalletCreateActionSuccess() {
...
+		thenCreatedAction := thenState.ActionAtIndex(1)
+		thenCreatedAction.
+			WithTxID(result.Txid.String()).
+			WithStatus(sdk.ActionStatusUnproven).
+			WithDescription(args.Description).
+			WithLabels(fixtures.CreateActionTestLabel).
+			WithSatoshis(-int64(args.Outputs[0].Satoshis) - fee) // Pay attention that this is negative value (user spends balance).
+
+		thenCreatedAction.OutputAtIndex(0).
+			WithSatoshis(args.Outputs[0].Satoshis).
```
chris-4chain, line 314: It would be good to check the listoutputs also:
```suggestion
			ListActionsAlignsListOutputs().
			WithSatoshis(args.Outputs[0].Satoshis).
```

pkg/wallet/wallet_create_action_test.go
```diff
@@ -295,3 +440,61 @@ func (s *WalletTestSuite) TestWalletCreateActionError() {
...
+		// when:
+		args := fixtures.DefaultWalletCreateActionArgs(t, walletargs.WithInput(input))
+		result, err := aliceWallet.CreateAction(t.Context(), args, fixtures.DefaultOriginator)
+
+		// then:
+		assert.Nil(t, result, "Wallet shouldn't return result when all services are down")
+		require.Error(t, err, "Wallet should return error when not delayed broadcast failed")
+
+		// and:
+		// TODO: replacace with better assertions for error - when we will have custom type for it
```
chris-4chain:
```suggestion
		// TODO: replace with better assertions for error - when we will have custom type for it
```

pkg/internal/validate/validate_not_delayed_process_action_result.go
dorzepowski: I created the task for that. To not make this PR too big.

pkg/wallet/internal/actions/wallet_create_action.go
dorzepowski: I'm still afraid of the stability of BEEF, therefore I would like to make this handling separately - there is a task for that.

pkg/wallet/internal/actions/wallet_create_action.go
dorzepowski: this is some weird struct BEEF party, I don't really get the idea what is it for, and if it gives any benefits - there is a task for analysing and for introducing this.

pkg/wallet/internal/actions/wallet_create_action.go
dorzepowski: There is no such a logic in TS, so I don't want to be too eager with such a tweak, because this actually forces the adopter to know the difference between Go and TS implementation - while there is a public AbortAction method.

pkg/wallet/internal/mapping/mapping_process_action_args.go
dorzepowski: good catch, thx.

pkg/wallet/internal/mapping/mapping_send_with_results.go
dorzepowski: 1. Doesn't look like a big benefit - from "effective" 5 lines of code - now there is 20 and the reduction in future doesn't seem to be greater
2. Using short map instead of switch has impact on the speed of this mapping - I'm learning pre-optimisation from you 😂

pkg/wallet/internal/testabilities/fixture_wallet.go
dorzepowski: Thanks, this forced me to check the implementation, and I don't know why, but previously I was sure that it will create additional slice - that was my totally wrong assumption that lead me to bad code.

pkg/wallet/wallet.go
dorzepowski: This is bigger problem, and I don't have yet solution/idea for it.

IMHO those flags should be set with the opts pattern - or with some config struct at the moment of creation.

BUT in ts - those flags are accessible and modifiable - they are public - and I'm not sure if we should expose them like it is done now, or maybe there should be a setter or maybe we shouldn't expose them at all, and only allow for configuring them on the start - which seem safer in case of concurrent access.

What do you think about it?

pkg/wallet/internal/mapping/mapping_send_with_results.go
chris-4chain: The reason I'm suggesting this is because I've already seen similar mapping functions. 
I'm also guessing, that the number of such mappings will grow - so handy function might be beneficial.
But it's minor anyway, and doesn't directly concerns your PR so - do whatever you decide :) 

https://github.com/bsv-blockchain/go-wallet-toolbox/blob/4841f9e24ba5ec46bf8044d32f0272203a18af51/pkg/wallet/internal/mapping/mapping_send_with_results.go#L38

https://github.com/bsv-blockchain/go-wallet-toolbox/blob/62488744b7486d522fd98d9514aeac0595f61db3/pkg/wallet/internal/mapping/mapping_list_actions.go#L125 
- In the second example, it also should be wdk type (instead of string)

pkg/wallet/wallet.go
chris-4chain: I would make them as standard opts pattern. 
And expose methods from wallet like:
```
SetIncludeAllSourceTransactions(value bool)
SetAutoKnownTxIDs(value bool)
SetTrustSelf(weird type of wallet.TrustSelf)
```

pkg/wallet/internal/mapping/mapping_send_with_results.go
dorzepowski: Sorry, I don't want to change it, I don't really see a big benefit from making it that way.

chris-4chain (approved):

PR #333, opened 2025-07-07
[FEATURE] isValidRootForHeight (SPV) on WoC

### LINKED ISSUE #315: [FEATURE] WalletService: isValidRootForHeight (SPV) on WoC
## Desired Solution

## Suggested Implementation

## Alternatives Considered

## Additional Context

## Categorize Feature

ac4ch:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter


examples/is_valid_root/is_valid_root_for_height.go
```diff
@@ -0,0 +1,113 @@
...
+	printRow := func(cells []string) {
+		for i, c := range cells {
+			fmt.Printf("%-*s  ", colW[i], c)
+		}
+		fmt.Println()
+	}
+
+	printRow(headers)
+	for i := range headers {
+		fmt.Printf("%s  ", stringRepeat("-", colW[i]))
```
chris-4chain:
```suggestion
		fmt.Printf("%s  ", strings.Repeat("-", colW[i]))
```
You don't need to define your own function. There is similar in std library already.

examples/is_valid_root/is_valid_root_for_height.go
```diff
@@ -0,0 +1,113 @@
...
+	const (
+		height  = uint32(903321)
+		baseURL = "https://api.whatsonchain.com/v1/bsv/main"
+	)
+
+	cfg := defs.DefaultServicesConfig(defs.NetworkMainnet)
+	ws := services.New(slog.Default(), cfg)
+
+	rootHex, err := fetchRoot(baseURL, height)
+	dieIf(err, "fetch header")
```
chris-4chain: just panic would be enough 
```golang
	if err != nil {
		panic(fmt.Errorf("something: %w", err))
	}
```

examples/is_valid_root/is_valid_root_for_height.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+func stringRepeat(ch string, n int) string {
+	if n <= 0 {
+		return ""
+	}
+	return fmt.Sprintf("%s%s", ch, stringRepeat(ch, n-1))
+}
+
+func dieIf(err error, what string) {
```
chris-4chain: Is it in "deutsch"? 🙈
```golang
derIf
dieIf
dasIf
```

examples/is_valid_root/is_valid_root_for_height.go
```diff
@@ -0,0 +1,113 @@
...
+)
+
+func main() {
+	const (
+		height  = uint32(903321)
+		baseURL = "https://api.whatsonchain.com/v1/bsv/main"
+	)
+
+	cfg := defs.DefaultServicesConfig(defs.NetworkMainnet)
+	ws := services.New(slog.Default(), cfg)
```
chris-4chain: Maybe `srv` because `ws` reminds me of `WebSocket` :)

examples/is_valid_root/is_valid_root_for_height.go
```diff
@@ -0,0 +1,113 @@
...
+
+	rootHex, err := fetchRoot(baseURL, height)
```
chris-4chain: I think, in the example, we should not load the root automatically. 
- To, just hardcoding some root hex would be enough.
- Of course with some comment and/or a url to whatsonchain.

examples/is_valid_root/is_valid_root_for_height.go
```diff
@@ -0,0 +1,113 @@
...
+
+	root, err := chainhash.NewHashFromHex(rootHex)
+	dieIf(err, "NewHashFromHex")
+
+	ok, err := ws.IsValidRootForHeight(context.Background(), root, height)
+	dieIf(err, "IsValidRootForHeight")
+
+	bad := *root
+	bad[0] ^= 0xff
+	okBad, err := ws.IsValidRootForHeight(context.Background(), &bad, height)
```
chris-4chain: I'm not sure if showing that we handle "bad" roots correctly adds any value. 

It seems to me that in the examples only happy paths should be presented. 
- When a user wanted to check what happens "if" some wrong root is provided, they would just modify the input root hex and check.

pkg/defs/services.go
```diff
@@ -17,6 +17,12 @@ const (
 	// DefaultWoCBroadcastDelay is the delay after which the WhatsOnChain service will broadcast a transaction
 	DefaultWoCBroadcastDelay = 3 * time.Second
 
+	// DefaultRootForHeightValidationTimeout is the timeout for fetching the root for height validation
+	DefaultRootForHeightValidationTimeout = 1 * time.Second
```
chris-4chain: Isn't it too small value? Especially as a default...?

pkg/internal/testabilities/testservices/fixture_services.go
```diff
@@ -59,6 +59,8 @@ func givenServicesWithNetwork(t testing.TB, network defs.BSVNetwork) ServicesFix
 
 	servicesConfig := defs.DefaultServicesConfig(network)
 	servicesConfig.WhatsOnChain.BroadcastDelay = 0
+	servicesConfig.WhatsOnChain.RootForHeightValidationTimeout = 0
```
chris-4chain: Isn't `zero` too small value (even for tests) :thinking:

pkg/services/internal/whatsonchain/testabilities/fixtures.go
```diff
@@ -34,8 +34,10 @@ func (f *wocServiceFixture) NewWoCService(opts ...func(*whatsonchain.WhatsOnChai
 	network := f.Network()
 
 	config := to.OptionsWithDefault(defs.WhatsOnChain{
```
chris-4chain: Hmmm, It's not directly related to your current PR, but:
the function `to.OptionsWithDefault` is not necessary if no "functional" option is passed as vararg.

pkg/services/internal/whatsonchain/is_valid_root.go
```diff
@@ -0,0 +1,74 @@
...
+	"fmt"
+	"net/http"
+	"time"
+
+	"github.com/bsv-blockchain/go-sdk/chainhash"
+)
+
+// IsValidRootForHeight checks if the provided Merkle root is valid for the given block height.
+func (woc *WhatsOnChain) IsValidRootForHeight(ctx context.Context, root *chainhash.Hash, height uint32) (bool, error) {
+
```
chris-4chain: Unnecessary empty line.

pkg/services/internal/whatsonchain/service.go
```diff
@@ -27,9 +28,12 @@ type WhatsOnChain struct {
...
 
-	bsvExchangeRate   defs.BSVExchangeRate // TODO: possibly handle by some caching structure/redis
-	bsvUpdateInterval time.Duration
-	broadcastDelay    time.Duration
+	bsvExchangeRate                defs.BSVExchangeRate // TODO: possibly handle by some caching structure/redis
+	bsvUpdateInterval              time.Duration
+	broadcastDelay                 time.Duration
+	rootForHeightValidationTimeout time.Duration
+	rootForHeightValidationRetries int
+	rootCache                      map[uint32]*chainhash.Hash
```
chris-4chain:
```suggestion
	rootCache                      map[uint32]*chainhash.Hash // TODO: possibly handle by some caching structure/redis
```

pkg/services/internal/whatsonchain/is_valid_root.go
```diff
@@ -0,0 +1,74 @@
...
+		ok, err := woc.fetchAndCompare(ctx, root, height)
+		if err == nil {
+			return ok, nil
+		}
+		lastErr = err
+
+		select {
+		case <-ctx.Done():
+			return false, fmt.Errorf("context canceled while validating Merkle root for height %d: %w", height, ctx.Err())
+		case <-time.After(woc.rootForHeightValidationTimeout):
```
chris-4chain: I would prefer to handle it this way:

```golang
	timeoutCtx, cancel := context.WithTimeout(ctx, woc.rootForHeightValidationTimeout)
	defer cancel()
```

pkg/services/internal/whatsonchain/is_valid_root.go
```diff
@@ -0,0 +1,74 @@
...
+
+	for range woc.rootForHeightValidationRetries {
+		ok, err := woc.fetchAndCompare(ctx, root, height)
+		if err == nil {
+			return ok, nil
+		}
+		lastErr = err
+
+		select {
+		case <-ctx.Done():
```
chris-4chain, line 28: This: 
```golang
		if err = timeoutCtx.Err(); err != nil {
			return fmt.Errorf("context canceled, aborting synchronizeTxStatuses: %w", err)
		}
```

along with my previous comment would make the `select` unnecessary.

examples/is_valid_root/is_valid_root_for_height.go
ac4ch: not really more like in this context `he will die`

pkg/defs/services.go
nonus25:
```     
error = sdk.WalletError.fromUnknown(eu)
        if (tryCount > retries) {
          throw error
        }
        await wait(1000)
      }
```

in TS :)

pkg/services/internal/whatsonchain/testabilities/fixtures.go
ac4ch: i think i did without of this before and Damian was pointing that i used it in a strange way w/o of this default call .. so i started using that method ...

pkg/internal/testabilities/testservices/fixture_services.go
ac4ch: then used the default value which is 1s

pkg/services/internal/whatsonchain/is_valid_root_test.go
```diff
@@ -0,0 +1,135 @@
...
+)
+
+func TestIsValidRootForHeight(t *testing.T) {
+	type want struct {
+		ok  bool
+		err error
+	}
+
+	goodRoot, _ := chainhash.NewHashFromHex(tst.TestMerkleRootHex)
+	badRoot := func() *chainhash.Hash { h := *goodRoot; h[0] ^= 0xff; return &h }()
```
cain4chain: For this variable I suggest to change "goodRoot" to "validRoot" and "badRoot" to "invalidRoot" what do you think?

cain4chain (approved):

pkg/services/internal/whatsonchain/is_valid_root.go
```diff
@@ -0,0 +1,77 @@
...
+		if err == nil {
+			return ok, nil
+		}
+
+		if ce := attemptCtx.Err(); ce != nil {
+			return false, fmt.Errorf("context canceled while validating Merkle root for height %d: %w", height, ce)
+		}
+
+		lastErr = err
+	}
```
chris-4chain: Sorry, but the `woc.rootForHeightValidationTimeout` is not a timeout value for **HTTP request** but rather a `delay` between retries:

```suggestion
		for range woc.rootForHeightValidationRetries {
		ok, err := woc.fetchAndCompare(ctx, root, height)
		if err == nil {
			return ok, nil
		}
		
		select {
		case <-ctx.Done():
			return false, fmt.Errorf("context canceled while validating Merkle root for height %d: %w", height, ctx.Err())
		case <-time.After(woc.rootForHeightValidationTimeout):
			// continue to retry
		}

		lastErr = err
	}
```

I also suggest to rename the `rootForHeightValidationTimeout` and the matching config param.

pkg/defs/services.go
chris-4chain: See my comment in `is_valid_root.go`

chris-4chain (changes requested):

pkg/defs/whats_on_chain.go
```diff
@@ -7,10 +7,12 @@ import (
...
 type WhatsOnChain struct {
-	APIKey            string          `mapstructure:"api_key"`
-	BSVExchangeRate   BSVExchangeRate `mapstructure:"bsv_exchange_rate"`
-	BroadcastDelay    time.Duration   `mapstructure:"broadcast_delay"`
-	BSVUpdateInterval *time.Duration  `mapstructure:"bsv_update_interval"`
+	APIKey                     string          `mapstructure:"api_key"`
+	BSVExchangeRate            BSVExchangeRate `mapstructure:"bsv_exchange_rate"`
+	BroadcastDelay             time.Duration   `mapstructure:"broadcast_delay"`
+	BSVUpdateInterval          *time.Duration  `mapstructure:"bsv_update_interval"`
+	RootForHeightRetryInterval time.Duration   `mapstructure:"root_for_height_validation_timeout"`
```
chris-4chain: You forgot to update the name in the `mapstructure` tag

pkg/services/internal/whatsonchain/is_valid_root_test.go
```diff
@@ -0,0 +1,135 @@
...
+				f.WhatsOnChain().Transport().RegisterResponder(http.MethodGet, pat,
+					func(_ *http.Request) (*http.Response, error) {
+						cancel(context.Canceled)
+						return nil, context.Canceled
+					})
+			},
+			root:  validRoot,
+			ctx:   nil, // replaced below
+			want1: want{ok: false, err: context.Canceled},
+			want2: want{ok: false, err: context.Canceled},
```
chris-4chain: For this single case (with the special context and stuff) I would suggest a separated TestFunction (no "table-test)
- This would also simplify other cases in test-table.

pkg/services/services_is_valid_root_test.go
```diff
@@ -0,0 +1,125 @@
...
+			},
+			root: validRoot,
+			ctx:  t.Context(),
+			want: want{ok: false, expectErr: true},
+		},
+		{
+			name: "context canceled during request",
+			setup: func(f testservices.ServicesFixture) {
+				localCtx, cancel := context.WithCancelCause(t.Context())
+				ctxDuringRequest = localCtx
```
chris-4chain: Similar comment about separated test case. ☝️

chris-4chain (changes requested):

chris-4chain (commented):
Looks Good To Me 👍 💯 
I'll do manual tests :)

chris-4chain (approved):

PR #350, opened 2025-07-10
Feat(#342)config generator for example setup

### LINKED ISSUE #342: [FEATURE] Wallet Examples : create config key generator for wallet example setup

cain4chain:
## Description of Changes

This code contains Config generator for examples setup. It will create the file and provide the required fields for example setup testing.

## Linked Issues / Tickets

Closes #342 

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter


examples/internal/example_setup/setup.go
```diff
@@ -24,52 +24,82 @@ type Environment struct {
...
+		return fmt.Errorf("server_url is required")
+	}
+
+	if c.Alice.IdentityKey == "" || c.Alice.PrivateKey == "" {
+		return fmt.Errorf("alice.identity_key and alice.private_key are required")
+	}
+
+	if c.Bob.IdentityKey == "" || c.Bob.PrivateKey == "" {
+		return fmt.Errorf("bob.identity_key and bob.private_key are required")
+	}
```
mgosek-4chain: I would suggest to add an extra verify function to the `UserConfig` structure:

```
func (u *UserConfig) Verify() error {
	if len(u.IdentityKey) == 0 {
		return fmt.Errorf("identity key value is required")
	}
	if len(u.PrivateKey) == 0 {
		return fmt.Errorf("private key value is required")
	}
	return nil
}
```


And then simplify the `Validate` codebase to:

```
func (c *SetupConfig) Validate() error {
	if _, err := defs.ParseBSVNetworkStr(string(c.Network)); err != nil {
		return fmt.Errorf("invalid BSV network: %w", err)
	}

	if c.Network == "" {
		return fmt.Errorf("network is required")
	}

	if c.ServerURL == "" {
		return fmt.Errorf("server_url is required")
	}

	err := c.Alice.Verify()
	if err != nil {
		return fmt.Errorf("alice user config is invalid: %w", err)
	}

	err = c.Bob.Verify()
	if err != nil {
		return fmt.Errorf("bob user config is invalid: %w", err)
	}

	return nil
}

```

mgosek-4chain (changes requested):

mgosek-4chain (approved):

examples/examples_config_gen/main.go
```diff
@@ -0,0 +1,72 @@
...
+	flag.StringVar(network, "n", string(defs.NetworkTestnet), "BSV network (shorthand)")
+
+	serverURL := flag.String("server-url", "http://localhost:8100", "Server URL")
+	flag.StringVar(serverURL, "s", "http://localhost:8100", "Server URL (shorthand)")
+
+	flag.Parse()
+
+	_, err := defs.ParseBSVNetworkStr(*network)
+	if err != nil {
+		log.Fatalf("Invalid network '%s': %v", *network, err)
```
jakubmkowalski: We agreed to use panic's in fatal cases in the examples.

examples/examples_config_gen/main.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	alice, err := generateUserConfig()
+	if err != nil {
+		log.Fatalf("Error generating Alice config: %v", err)
```
jakubmkowalski: Same as above.

examples/examples_config_gen/main.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	bob, err := generateUserConfig()
+	if err != nil {
+		log.Fatalf("Error generating Bob config: %v", err)
```
jakubmkowalski: Same as above.

examples/examples_config_gen/main.go
```diff
@@ -0,0 +1,72 @@
...
+	cfg := example_setup.SetupConfig{
+		Network:   defs.BSVNetwork(*network),
+		ServerURL: *serverURL,
+		Alice:     alice,
+		Bob:       bob,
+	}
+
+	err = cfg.Validate()
+	if err != nil {
+		log.Fatalf("Configuration validation failed: %v", err)
```
jakubmkowalski: Same as above.

examples/examples_config_gen/main.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	err = cfg.ToYAMLFile(*outputFile)
+	if err != nil {
+		log.Fatalf("Error writing configuration: %v", err)
```
jakubmkowalski: Same as above.

examples/examples_config_gen/main.go
```diff
@@ -0,0 +1,72 @@
...
+
+	ec "github.com/bsv-blockchain/go-sdk/primitives/ec"
+	"github.com/bsv-blockchain/go-wallet-toolbox/examples/internal/example_setup"
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/defs"
+)
+
+func main() {
+
+	outputFile := flag.String("output-file", "examples/internal/example_setup/examples-config.yaml", "Output configuration file path")
+	flag.StringVar(outputFile, "o", "examples/internal/example_setup/examples-config.yaml", "Output configuration file path (shorthand)")
```
ac4ch: I am not sure we wanna have flags / args passed in examples but if we do use the flag pkg different way 
```go
var outputFile string
flag.StringVar(&outputFile, "output-file", "examples/internal/example_setup/examples-config.yaml", "Output configuration file path")
flag.StringVar(&outputFile, "o", "examples/internal/example_setup/examples-config.yaml", "Output configuration file path (shorthand)")
``` 
use `StringVar` directly to avoid dups ...

ac4ch (approved):

jakubmkowalski (approved):

examples/examples_config_gen/main.go
```diff
@@ -0,0 +1,75 @@
...
+
+import (
+	"encoding/hex"
+	"flag"
+	"fmt"
+
+	ec "github.com/bsv-blockchain/go-sdk/primitives/ec"
+	"github.com/bsv-blockchain/go-wallet-toolbox/examples/internal/example_setup"
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/defs"
+	"github.com/bsv-blockchain/go-wallet-toolbox/examples/internal/show"
```
chris-4chain: Imports are not sorted

```suggestion
	ec "github.com/bsv-blockchain/go-sdk/primitives/ec"
	"github.com/bsv-blockchain/go-wallet-toolbox/examples/internal/example_setup"
	"github.com/bsv-blockchain/go-wallet-toolbox/examples/internal/show"
	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/defs"
```

Apparently our lint formatter has no power in the `examples` directory

examples/examples_config_gen/main.go
```diff
@@ -0,0 +1,75 @@
...
+)
+
+func main() {
+
+	var outputFile string
+	flag.StringVar(&outputFile, "output-file", "examples/internal/example_setup/examples-config.yaml", "Output configuration file path")
```
chris-4chain: To be honest, I don't like the idea that the config is buried somewhere in the internal directories. 

I would rather expect that it is right in the `examples` directory. 
Of course, it's good that utilities for loading/generating config are in internal package. 

I think, we need to discuss it internally :)

chris-4chain (changes requested):

examples/examples_config_gen/main.go
dorzepowski: yes, I agree, 
IMHO the best would to have config in the root examples directory.

examples/examples_config_gen/main.go
```diff
@@ -0,0 +1,75 @@
...
+
+func generateUserConfig() (example_setup.UserConfig, error) {
+	privKey, err := ec.NewPrivateKey()
+	if err != nil {
+		return example_setup.UserConfig{}, fmt.Errorf("failed to generate private key: %w", err)
+	}
+
+	return example_setup.UserConfig{
+		IdentityKey: privKey.PubKey().ToDERHex(),
+		PrivateKey:  hex.EncodeToString(privKey.Serialize()),
```
dorzepowski: maybe we could use WIF instead of plain hex, it would be shorter here, and WIF is safer to copy as it contains a checksum

cain4chain: Just had a thought about consistency with TS wallet toolbox ENV for testing uses private key [hex values](https://github.com/bsv-blockchain/wallet-toolbox/blob/3029ea8f2e62ac31fa524fedcfc8c1f2b68e3e43/.env.template#L18C76-L18C106). 

Should we consider keeping is hex values in that case? The server private key for the infra-config is based in hex also.

dorzepowski (approved):
LGTM

examples/examples_config_gen/main.go
dorzepowski: Maybe you're right, maybe we should use hex, to make it easier to copy&paste the keys from ts here.
Sorry for messing with this 🙈

examples/internal/example_setup/load_config.go
```diff
@@ -0,0 +1,98 @@
+package example_setup
+
+import (
+	"fmt"
+	"os"
+	"encoding/hex"
```
chris-4chain: Imports are not sorted:
```golang
	"encoding/hex"
	"fmt"
	"os"
```

FYI: We have a task to fix our linter to format (--fix) also the "examples" directory.

examples/examples-config-example.yaml
chris-4chain, line 1: I'm not sure if this file is needed at all.
Because the config will be autogenerated anyway - and then, someone will be able to adjust it.

And because of this file is not autogenerated by `go generate` it would be hard to keep it up-to-date.

examples/internal/example_setup/load_config.go
```diff
@@ -0,0 +1,98 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("error generating Alice config: %w", err)
+	}
+
+	bob, err := generateUserConfig()
+	if err != nil {
+		return nil, fmt.Errorf("error generating Bob config: %w", err)
+	}
+
+	cfg := &SetupConfig{
```
chris-4chain, line 50: There is one confusion here.
The `defaultConfig()` function returns other content than the one created here in `generateDefaultConfig`.

examples/examples-config-example.yaml
cain4chain: ok makes sense. I will remove it, thanks

examples/internal/example_setup/load_config.go
cain4chain: yes good point. Will change to generateConfig.

chris-4chain (approved):

PR #340, opened 2025-07-08
feat(#173): add script hash history

jakubmkowalski:
This pull request introduces functionality for retrieving script history, including both confirmed and unconfirmed transactions, for a given script hash. It includes validation, API integration, and extensive test coverage to ensure the correctness of the implementation. The changes span across fixtures, service methods, and data structures, enabling robust handling of script history queries.

### Script History Retrieval Functionality:
* **Validation of Script Hashes**: Added a `validateScriptHash` function in `pkg/services/internal/whatsonchain/script_history.go` to ensure script hashes meet length and format requirements.
* **API Integration**: Implemented methods `getConfirmedScriptHistory` and `getUnconfirmedScriptHistory` in `pkg/services/internal/whatsonchain/script_history.go` to interact with the WhatsOnChain API and retrieve script history. Combined results are handled by the `GetScriptHistory` method.
* **Service Exposure**: Added `ScriptHistory` method to `WalletServices` in `pkg/services/services.go` for external access to script history retrieval.

### Fixture Enhancements for Testing:
* **Fixture Methods**: Extended `WhatsOnChainFixture` in `pkg/internal/testabilities/testservices/fixture_woc.go` with methods for mocking script history responses, including error handling and pagination. [[1]](diffhunk://#diff-0a498ce2b803514ad0826deb0f2d90e0d7664b8166232bf949179dda7f93cac9R37-R43) [[2]](diffhunk://#diff-0a498ce2b803514ad0826deb0f2d90e0d7664b8166232bf949179dda7f93cac9R357-R625)
* **Script History Data Builder**: Introduced `ScriptHistoryDataBuilder` to easily construct mock script history data for testing.

### Data Structure Updates:
* **Script History Models**: Added new types in `pkg/wdk/services_script_hash_history.go` to represent script history items, results, and options for querying confirmed history. Includes validation for query options.

### Test Coverage:
* **Unit Tests**: Added comprehensive tests in `pkg/services/services_get_script_history_test.go` to validate script history retrieval, including builder pattern usage, query fixture methods, validation errors, pagination, and large history scenarios.

These changes collectively enhance the ability to query, validate, mock, and test script history functionality within the application.## Description of Changes

Provide a brief description of the changes you've made.

## Linked Issues / Tickets

Closes #173 

## Testing Procedure

Describe the tests you've added or any testing steps you've taken.

- [ ] I have added new unit tests
- [ ] All tests pass locally
- [ ] I have tested manually in my local environment

## Checklist:

[...]

PR #351, opened 2025-07-10
suggestion: use iterators in results of ScriptHistory getters

dorzepowski:
@jakubmkowalski suggestion how to iterators should be used to combine results


jakubmkowalski:
added in #340

PR #362, opened 2025-07-14
[FEATURE] isValidRootForHeight (SPV) on Bitalis and BHS

### LINKED ISSUE #364: [FEATURE] isValidRootForHeight (SPV) on Bitalis
## Desired Solution

## Suggested Implementation

## Alternatives Considered

## Additional Context

## Categorize Feature

ac4ch:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter


pkg/services/internal/whatsonchain/service.go
```diff
@@ -204,3 +240,20 @@ func (woc *WhatsOnChain) PostBEEF(ctx context.Context, beef *transaction.Beef, t
 
 	return &wdk.PostedBEEF{TxIDResults: txResults}, nil
 }
+
+// IsValidRootForHeight checks if the provided Merkle root is valid for the given block height.
```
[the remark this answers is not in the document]
ac4ch: `AddRetryCondition(httpx.RetryOnErrOr5xx).`

pkg/services/internal/bitails/service.go
```diff
@@ -87,6 +90,19 @@ func (b *Bitails) PostBEEF(ctx context.Context, beef *transaction.Beef, txIDs []
 	return &wdk.PostedBEEF{TxIDResults: results}, nil
 }
 
+// IsValidRootForHeight checks if the supplied merkle-root belongs to the block at `height`.
+func (b *Bitails) IsValidRootForHeight(ctx context.Context, root *chainhash.Hash, height uint32) (bool, error) {
+	if cached, ok := b.rootCache[height]; ok {
+		return cached.IsEqual(root), nil
+	}
+
+	ok, err := b.fetchAndCompareRoot(ctx, root, height)
```
chris-4chain: Consider just fetch the root by separate function and  the `remoteRoot` here. 
To me, it would be more readable to have `remoteRoot.IsEqual(root)` and `b.rootCache[height] = remoteRoot` right here.

pkg/services/internal/bitails/helpers.go
```diff
@@ -7,9 +7,58 @@ import (
...
+
+	resp, err := b.httpClient.R().SetContext(ctx).SetResult(&dto).AddRetryCondition(httpx.RetryOnErrOr5xx).Get(url)
+	if err != nil {
+		return false, fmt.Errorf("failed to fetch header: %w", err)
+	}
+
+	switch resp.StatusCode() {
+	case http.StatusOK:
+	case http.StatusNotFound:
+		b.rootCache[height] = new(chainhash.Hash)
```
chris-4chain: To me, we shouldn't store "empty hash" if not found, because it can change in time, like: "now the header is not found, but after 2 minutes it appears"

pkg/services/internal/bhs/internal/dto/dto.go
```diff
@@ -39,3 +40,27 @@ func (t *TipStateResponse) ConvertToChainBlockHeader() *wdk.ChainBlockHeader {
...
+type MerkleRootVerifyItem struct {
+	BlockHeight uint32 `json:"blockHeight"`
+	MerkleRoot  string `json:"merkleRoot"`
+}
+
+type merkleRootConfirmationState string
+
+const (
+	confirmed      merkleRootConfirmationState = "CONFIRMED"
+	unableToVerify merkleRootConfirmationState = "UNABLE_TO_VERIFY"
```
chris-4chain, line 53: It seems unused.

pkg/services/internal/whatsonchain/service.go
```diff
@@ -204,3 +240,20 @@ func (woc *WhatsOnChain) PostBEEF(ctx context.Context, beef *transaction.Beef, t
...
+func (woc *WhatsOnChain) IsValidRootForHeight(ctx context.Context, root *chainhash.Hash, height uint32) (bool, error) {
+	if err := ctx.Err(); err != nil {
+		return false, fmt.Errorf("context canceled while validating Merkle root for height %d: %w", height, err)
+	}
+
+	if cached, ok := woc.rootCache[height]; ok {
```
chris-4chain: ## This is serious.
The `rootCache` should be concurrency safe. 
Use `sync.Map` or `sync.RWMutex`

pkg/services/internal/bitails/helpers.go
```diff
@@ -7,9 +7,58 @@ import (
...
+	if err != nil {
+		return false, err
+	}
+
+	remoteRoot, err := chainhash.NewHashFromHex(hdr.MerkleRoot)
+	if err != nil {
+		return false, fmt.Errorf("parse remote root: %w", err)
+	}
+
+	b.rootCache[height] = remoteRoot
```
chris-4chain: ## This is serious.
The `rootCache` should be concurrency safe.
Use `sync.Map` or `sync.RWMutex`

chris-4chain (changes requested):

pkg/services/internal/bhs/service.go
```diff
@@ -53,32 +53,37 @@ func New(httpClient *resty.Client, logger *slog.Logger, network defs.BSVNetwork,
...
 }
 
-func (b *BlockHeadersService) FindChainTipHeader(ctx context.Context) (*wdk.ChainBlockHeader, error) {
-	var block dto.TipStateResponse
-	url, err := tipLongestURL(b.cfg.URL)
+func (b *BlockHeadersService) IsValidRootForHeight(ctx context.Context, root *chainhash.Hash, height uint32) (bool, error) {
+	url, err := verifyMerkleRootURL(b.cfg.URL)
 	if err != nil {
-		return nil, fmt.Errorf("error while building tip URL for Block Headers Service API: %w", err)
+		return false, fmt.Errorf("failed for service %s: build verify URL: %w", ServiceName, err)
```
chris-4chain: Service name is added on the top level by https://github.com/bsv-blockchain/go-wallet-toolbox/blob/a1c9948ebba0c8dc5224cc037290635f1a2f331b/pkg/services/internal/servicequeue/queues.go#L234.

So no need to add `failed for service %s`.
This is a general comment. Applicable for other places also.

pkg/services/internal/bhs/service_test.go
```diff
@@ -155,3 +156,79 @@ func TestBlockHeadersService_ErrorCase(t *testing.T) {
...
+	)
+
+	validRoot, err := chainhash.NewHashFromHex(root)
+	require.NoError(t, err)
+
+	type setupFn func(fix testservices.BHSFixture)
+
+	tests := []struct {
+		name      string
+		setup     setupFn
```
chris-4chain: Why not inline the setupFn?

```suggestion
		setup func(fix testservices.BHSFixture)
```

pkg/services/internal/bhs/service_test.go
```diff
@@ -155,3 +156,79 @@ func TestBlockHeadersService_ErrorCase(t *testing.T) {
...
+	// given:
+	const (
+		height  = uint32(900_123)
+		rootHex = "abcdef0123456789abcdef0123456789abcdef0123456789abcdef0123456789"
+	)
+
+	root, err := chainhash.NewHashFromHex(rootHex)
+	require.NoError(t, err)
+
+	fix := bhsTst.Given(t)
```
chris-4chain:
```suggestion
	given := bhsTst.Given(t)
```

pkg/services/internal/whatsonchain/service.go
```diff
@@ -204,3 +242,25 @@ func (woc *WhatsOnChain) PostBEEF(ctx context.Context, beef *transaction.Beef, t
 
 	return &wdk.PostedBEEF{TxIDResults: txResults}, nil
 }
+
+// IsValidRootForHeight checks if the provided Merkle root is valid for the given block height.
+func (w *WhatsOnChain) IsValidRootForHeight(ctx context.Context, root *chainhash.Hash, height uint32) (bool, error) {
```
chris-4chain:
```suggestion
func (woc *WhatsOnChain) IsValidRootForHeight(ctx context.Context, root *chainhash.Hash, height uint32) (bool, error) {
```

Please keep the receivers consistent.

pkg/internal/testabilities/testservices/fixture_bitails.go
```diff
@@ -234,6 +241,22 @@ func (b *bitailsFixture) WillReturnTxStatus(txid string, blockHeight int) {
 	)
 }
 
+func (f *bitailsFixture) WillRespondWithBlockHeaderByHeight(status int, height uint32, headerHex string) {
```
chris-4chain: Please rename to `b *bitailsFixture` to keep receiver names consistent.

pkg/services/internal/bhs/service.go
```diff
@@ -87,29 +93,41 @@ func (b *BlockHeadersService) CurrentHeight(ctx context.Context) (uint32, error)
...
-		SetContext(ctx).
-		SetResult(&tip).
-		Get(url)
+	_, err = b.doGET(ctx, url, &tip)
 	if err != nil {
-		return 0, fmt.Errorf("failed for service %s: height query failed (%s): %w", ServiceName, url, err)
-	}
-	if res.StatusCode() != http.StatusOK {
-		return 0, fmt.Errorf("failed for service %s: unexpected HTTP %d for %s", ServiceName, res.StatusCode(), url)
+		return 0, err
```
chris-4chain: :)

pkg/services/internal/bhs/service.go
```diff
@@ -87,29 +93,41 @@ func (b *BlockHeadersService) CurrentHeight(ctx context.Context) (uint32, error)
 	var tip dto.TipStateResponse
 	url, err := tipLongestURL(b.cfg.URL)
 	if err != nil {
-		return 0, fmt.Errorf("failed for service %s to build tip URL: %w", ServiceName, err)
+		return 0, errors.Join(err, fmt.Errorf("error from service %s", ServiceName))
```
chris-4chain: **General question** 
Why have you switched to `errors.Join` ? (not only here but in many places?)
To me, major (or all) of the cases you modified are better suited for wrapping. 

Besides, in this project we decided to use wrapping as primary choice - and, if you want to change it - feel free to open some discussion. 

BTW: As I mentioned somewhere above - you don't need to add the text `error from service <service_name>` because it is added on the upper layer:
https://github.com/bsv-blockchain/go-wallet-toolbox/blob/a1c9948ebba0c8dc5224cc037290635f1a2f331b/pkg/services/internal/servicequeue/queues.go#L234
... and there `errors.Join` is used in proper (to me) case

chris-4chain (changes requested):

pkg/services/internal/bhs/service.go
dorzepowski: I also would like to hear the motivation to switch from simple Errorf to errors.Join.

I don't really see the reason to use Join for those cases, for me:
1. errors.Join is for collecting errors that have equal importance, like when you make a whole validation and collecting all the errors from the validation instead of returning first wrong thing - then Join seem to be good method to show that all those errors are equal and that are just collected.
2. fmt.Errorf used as a wrapper is like passing a cause to some error in other languages
3. It's changing the flow of reading the code, to which we're used to because of common Errorf wrapping approach in other places
4. even if there is some performance gain (which I don't believe that there is) I don't think that this is good moment to optimise anything now - when we don't have good measures nor a known usage paths - also I don't think that error handling would be the first target for the optimisation - as this is rather the unhappy path.

ac4ch: lol

ac4ch: image

pkg/services/internal/bhs/service.go
```diff
@@ -87,29 +92,41 @@ func (b *BlockHeadersService) CurrentHeight(ctx context.Context) (uint32, error)
...
-		SetContext(ctx).
-		SetResult(&tip).
-		Get(url)
+	_, err = b.doGET(ctx, url, &tip)
 	if err != nil {
-		return 0, fmt.Errorf("failed for service %s: height query failed (%s): %w", ServiceName, url, err)
-	}
-	if res.StatusCode() != http.StatusOK {
-		return 0, fmt.Errorf("failed for service %s: unexpected HTTP %d for %s", ServiceName, res.StatusCode(), url)
+		return 0, fmt.Errorf("unexpected HTTP for %s", url)
```
dorzepowski: ### minor
```suggestion
		return 0, fmt.Errorf("unexpected HTTP error for %s", url)
```

pkg/services/internal/bhs/service.go
```diff
@@ -87,29 +92,41 @@ func (b *BlockHeadersService) CurrentHeight(ctx context.Context) (uint32, error)
...
+		return 0, fmt.Errorf("error building URL: %w", err)
 	}
 
-	res, err := b.
-		httpClient.
-		R().
-		SetContext(ctx).
-		SetResult(&tip).
-		Get(url)
+	_, err = b.doGET(ctx, url, &tip)
```
dorzepowski: Maybe simply use FindChainTipHeader to get tip header, and extract height from it, instead of duplicating logic in both methods?

pkg/services/internal/bhs/service.go
```diff
@@ -87,29 +92,41 @@ func (b *BlockHeadersService) CurrentHeight(ctx context.Context) (uint32, error)
...
 	if err != nil {
-		return 0, fmt.Errorf("failed for service %s: height query failed (%s): %w", ServiceName, url, err)
-	}
-	if res.StatusCode() != http.StatusOK {
-		return 0, fmt.Errorf("failed for service %s: unexpected HTTP %d for %s", ServiceName, res.StatusCode(), url)
+		return 0, fmt.Errorf("unexpected HTTP for %s", url)
 	}
+
 	if tip.IsZero() || tip.Height == 0 {
```
dorzepowski: maybe IsZero should also check the Height if it is zero?

pkg/services/services_is_valid_root_test.go
```diff
@@ -114,3 +115,245 @@ func TestWalletServices_IsValidRootForHeight_ContextCancelled(t *testing.T) {
...
+	require.False(t, ok)
+
+	// when:
+	given.BHS().Transport().Reset()
+	ok, err = svc.IsValidRootForHeight(ctx, root, bhsHeight)
+
+	// then:
+	require.True(t, errors.Is(err, context.Canceled))
+	require.False(t, ok)
+	require.Equal(t, 0, given.BHS().Transport().GetTotalCallCount())
```
dorzepowski: Why do you check it twice in the same test?
If you want to test for already canceled context, then please split it into 2 test cases, so it's easier to fix single failing testcase

pkg/services/internal/bitails/is_valid_root_test.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"net/http"
+	"testing"
+
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/services/internal/bitails/testabilities"
+	"github.com/stretchr/testify/require"
+)
+
+func TestBitails_IsValidRootForHeight(t *testing.T) {
```
dorzepowski, line 12: AFAIU this is a test case for caching, 
so please make it a separated test for caching
and add some more test cases, like - when the call will fail at the bitails side,
and when the Root is invalid

pkg/services/internal/whatsonchain/is_valid_root_test.go
```diff
@@ -83,12 +70,12 @@ func TestIsValidRootForHeight(t *testing.T) {
 	for _, tc := range cases {
 		t.Run(tc.name, func(t *testing.T) {
 			// given:
-			fix := tst.Given(t)
+			given := tst.Given(t)
 			if tc.setup != nil {
```
dorzepowski: this if is not needed, and I would rather assume that tc.setup must be always available - it can simply be empty function

pkg/services/internal/bhs/service.go
chris-4chain: That comment wasn't a request "to switch from wrappin to Join".
- I've just pointed out that the `ServiceName` is already added on top level mechanizm (using `errors.Join`)
- And I've provided a link to it.

pkg/services/internal/bitails/is_valid_root_test.go
ac4ch: more cases added

pkg/services/internal/bhs/helpers.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"context"
+	"fmt"
+	"net/http"
+
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/services/internal/httpx"
+	"github.com/go-resty/resty/v2"
+)
+
+func (b *BlockHeadersService) doGET(ctx context.Context, url string, result any) (*resty.Response, error) {
```
chris-4chain: Since the `doGET` and `doPOST` helper functions are used only once, I would suggest getting rid of them and make the logic inline.

pkg/services/internal/whatsonchain/service.go
```diff
@@ -142,6 +144,42 @@ func (woc *WhatsOnChain) UpdateBsvExchangeRate() (defs.BSVExchangeRate, error) {
...
 
+// MerklePath retrieves the merkle path for a transaction using WoC TSC proof.
+func (woc *WhatsOnChain) MerklePath(ctx context.Context, txID string) (*wdk.MerklePathResult, error) {
+	proof, err := woc.getTscProof(ctx, txID)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get TSC proof: %w", err)
+	}
+	if proof == nil {
+		// Proof not found
+		return &wdk.MerklePathResult{Name: ServiceName}, nil
```
chris-4chain: General comment:
SInce my PR has been merged, there is also the "note" returned in such situations.

See for reference:
https://github.com/bsv-blockchain/go-wallet-toolbox/blob/fa68ba9cfebcabf8a91db64c7ae3650e265f76c2/pkg/services/internal/bitails/service.go#L109-L112

Please adjust wherever applicable
...also for "success" cases, like: https://github.com/bsv-blockchain/go-wallet-toolbox/blob/fa68ba9cfebcabf8a91db64c7ae3650e265f76c2/pkg/services/internal/bitails/service.go#L147-L152

chris-4chain (changes requested):

pkg/services/internal/bhs/internal/dto/dto.go
```diff
@@ -24,7 +25,9 @@ type TipStateResponse struct {
 	Height    uint        `json:"height"`
 }
 
-func (t *TipStateResponse) IsZero() bool { return *t == TipStateResponse{} }
+func (t *TipStateResponse) IsZero() bool {
+	return *t == TipStateResponse{} || t.Height == 0
```
chris-4chain: There is some redundancy here:

The original expression is logically redundant. It's like saying "if it's raining heavily OR if it's wet outside". If it's raining heavily, it's definitely wet outside, so you could have just asked if it's wet outside. 

Consider:
1. `return t.Height == 0`
2. `return *t == TipStateResponse{}`

pkg/services/internal/bhs/service_test.go
chris-4chain: I don't see any changes regarding this. And also you haven't answer my comment.

ac4ch: must missed this .. probably i did other change and this got mark as outdated ... now is inline

pkg/services/internal/bhs/internal/dto/dto.go
ac4ch: reverted to the original `IsZero` check as suggested.... the extra `Height == 0` condition was logically redundant...

chris-4chain (approved):

dorzepowski (approved):

jakubmkowalski (approved):

PR #365, opened 2025-07-14
feat: support syncing tags for backup & refactored backup mechanism 

chris-4chain:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter


pkg/internal/storage/repo/outputs.go
```diff
@@ -238,7 +238,7 @@ func (o *Outputs) FindInputsAndOutputsWithBaskets(ctx context.Context, txIDs []u
 }
 
 func (o *Outputs) SaveOutput(ctx context.Context, output *entity.Output) error {
-	tags := slices.Map(output.Tags, func(tag string) *models.Tag {
+	tags := slices.Map(output.Tags, func(tag string) any {
```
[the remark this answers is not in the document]
chris-4chain: No can do - the `association.Replace` expects a vararg of `any` type (`...any`)

pkg/internal/storage/database/models/output_tags.go
```diff
@@ -1,7 +1,17 @@
...
+)
+
+type OutputTag struct {
+	CreatedAt time.Time
+	UpdatedAt time.Time
+	DeletedAt gorm.DeletedAt `gorm:"index"`
+
+	OutputID  uint   `gorm:"primary_key"`
+	TagName   string `gorm:"primary_key"`
+	TagUserID int    `gorm:"primary_key"`
```
mgosek-4chain, line 16: Is that expected?

mgosek-4chain (changes requested):

pkg/internal/storage/database/models/output_tags.go
chris-4chain: YES ✅ 
> It's generally a good and common practice to have a composite primary key for a many-to-many table, consisting of the two foreign key columns from the tables it links. 

In here - because, the tags have composite primary keys, we need three columns to connect `outputs` and `tags`

mgosek-4chain (approved):

pkg/internal/storage/repo/syncrepo/label_tag_commons.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+func (f *labelTagCommons[_, _, _]) Delete(ctx context.Context, userID int, name string) (deleted bool, err error) {
+	err = f.db.WithContext(ctx).Transaction(func(tx *gorm.DB) error {
+		txDelete := tx.Delete(
+			f.zeroModelPtr(),
+			"user_id = ? AND name = ?", userID, name,
+		)
+		if txDelete.Error != nil {
+			return fmt.Errorf("failed to delete: %w", err)
```
jakubmkowalski:
```suggestion
			return fmt.Errorf("failed to delete: %w", txDelete.Error)
```

pkg/internal/storage/repo/syncrepo/label_tag_map_commons.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+func (f *labelTagMapCommons[_, _]) Delete(ctx context.Context, parentID uint, userID int, name string) (deleted bool, err error) {
+	txDelete := f.db.WithContext(ctx).Delete(
+		f.zeroModelPtr(),
+		fmt.Sprintf("%s = ? AND %s = ? AND %s = ?", f.relationParentIDColumn, f.relationUserIDColumn, f.relationNameColumn),
+		parentID, userID, name,
+	)
+	if txDelete.Error != nil {
+		return false, fmt.Errorf("failed to delete many2many relation %q: %w", f.relationTableName, err)
```
jakubmkowalski:
```suggestion
		return false, fmt.Errorf("failed to delete many2many relation %q: %w", f.relationTableName, txDelete.Error)
```

jakubmkowalski (approved):

pkg/internal/storage/repo/syncrepo/label_tag_commons.go
chris-4chain: Good catch 👍

pkg/internal/storage/repo/syncrepo/find_chunk.go
```diff
@@ -0,0 +1,49 @@
...
+	scopes []gormScope,
+) ([]*ResultModel, error) {
+	var resultModels []*ResultModel
+
+	var model Model
+
+	err := db.Transaction(func(tx *gorm.DB) error {
+		if err := upsertNumericIDLookup(ctx, db, tx, func(db *gorm.DB) *gorm.DB {
+			return db.
+				Select(fmt.Sprintf("?, %s", stringID), tableName).
```
ac4ch, line 26: what is `stringID` cannot be this named more intuitive ?  `foreignKeyField` or similar ?

pkg/internal/storage/repo/syncrepo/find_chunk.go
```diff
@@ -0,0 +1,49 @@
...
+	db *gorm.DB,
+	tableName, stringID string,
+	scopes []gormScope,
+) ([]*ResultModel, error) {
+	var resultModels []*ResultModel
+
+	var model Model
+
+	err := db.Transaction(func(tx *gorm.DB) error {
+		if err := upsertNumericIDLookup(ctx, db, tx, func(db *gorm.DB) *gorm.DB {
```
ac4ch, line 24: `func(db *gorm.DB)` think it should be `func(tx *gorm.DB)`  for consistency with the transaction WDYT ?

pkg/internal/storage/repo/syncrepo/find_chunk.go
```diff
@@ -0,0 +1,49 @@
...
+
+func findChunk[Model, ResultModel any](
+	ctx context.Context,
+	db *gorm.DB,
+	tableName, stringID string,
+	scopes []gormScope,
+) ([]*ResultModel, error) {
+	var resultModels []*ResultModel
+
+	var model Model
```
ac4ch, line 21:
```suggestion
	modelPtr := new(Model)
```
then later `Model(modelPtr)` would not be safer ?

pkg/internal/storage/repo/syncrepo/label_tag_map_commons.go
```diff
@@ -0,0 +1,115 @@
...
+	db                     *gorm.DB
+	subjectTableName       string
+	relationTableName      string
+	relationUserIDColumn   string
+	relationNameColumn     string
+	relationParentIDColumn string
+}
+
+func (f *labelTagMapCommons[_, ReadModel]) FindChunk(ctx context.Context, userID int, opts ...queryopts.Options) ([]*ReadModel, error) {
+	labelStringIDClause := fmt.Sprintf("CONCAT(%s, '.', %s)", f.relationUserIDColumn, f.relationNameColumn)
```
ac4ch, line 25: since u building sql strings dynamically here i wonder if this is not risky per potential SQL injections `clause.Expr{}` would fit here better and in other places where u building query  WDYT ?

pkg/internal/storage/repo/syncrepo/numeric_id.go
```diff
@@ -0,0 +1,78 @@
...
+// It executes custom INSERT ... SELECT ... ON CONFLICT DO NOTHING based on the result of the provided stringIDsQuery function.
+func upsertNumericIDLookup(ctx context.Context, db, tx *gorm.DB, stringIDsQuery func(db *gorm.DB) *gorm.DB) error {
+	dry := db.Session(&gorm.Session{DryRun: true, Initialized: true}) // NOTICE: Initialized to separate the dry run from the actual transaction (this makes the Session to clone the Statement)
+	query := stringIDsQuery(dry)
+
+	insertSelect := &gorm.Statement{DB: db}
+	clause.Expr{
+		SQL:  fmt.Sprintf("INSERT INTO %s (table_name, string_id) %s ON CONFLICT DO NOTHING", genquery.NumericIDLookup.TableName(), query.Statement.SQL.String()),
+		Vars: query.Statement.Vars,
+	}.Build(insertSelect)
```
ac4ch, line 23: nice !!!

ac4ch, line 23: why not use `clause.Expr` for interpolation strings ? 
```suggestion
	clause.Expr{
	SQL: "INSERT INTO ?? (table_name, string_id) " + query.Statement.SQL.String() + " ON CONFLICT DO NOTHING",
	Vars: append(
		[]interface{}{clause.Table{Name: genquery.NumericIDLookup.TableName()}},
		query.Statement.Vars...,
	),
}.Build(insertSelect)
```

pkg/internal/storage/repo/syncrepo/sync_basket.go
```diff
@@ -0,0 +1,136 @@
...
+	NumID int
+}
+
+func (s *SyncBasket) tableName() string {
+	return genquery.OutputBasket.TableName()
+}
+
+func (s *SyncBasket) stringIDClause() string {
+	return "CONCAT(user_id, '.', name)"
+}
```
ac4ch, line 38: why do not use `clause.Expr`
```go
func (s *SyncBasket) stringIDClauseExpr() clause.Expr {
	return clause.Expr{
		SQL:  "CONCAT(??, '.', ??)",
		Vars: []any{
			clause.Column{Name: "user_id"},
			clause.Column{Name: "name"},
		},
	}
}
``` 
consider this as this should protect us more from injections

pkg/internal/storage/repo/syncrepo/sync_basket.go
```diff
@@ -0,0 +1,136 @@
...
+func (s *SyncBasket) FindBasketNameByNumIDForSync(ctx context.Context, basketNumID uint) (string, error) {
+	var basketName string
+
+	err := s.db.WithContext(ctx).Model(&models.OutputBasket{}).
+		Scopes(joinWithNumericIDLookupScope(s.stringIDClause(), s.tableName(), clause.InnerJoin)).
+		Where("num.num_id = ?", basketNumID).
+		Select("name").
+		Scan(&basketName).Error
+	if err != nil {
+		return "", fmt.Errorf("failed to find output basket name by numeric ID: %w", err)
```
ac4ch, line 107:
```suggestion
		if errors.Is(err, gorm.ErrRecordNotFound) {
	             return "", nil // or custom NotFound error
                }
```
think this way u could distinguishing missing data vs DB issues ?

pkg/internal/storage/repo/syncrepo/sync_basket.go
```diff
@@ -0,0 +1,136 @@
...
+		UserID:                  entity.UserID,
+		Name:                    entity.Name,
+		NumberOfDesiredUTXOs:    entity.NumberOfDesiredUTXOs,
+		MinimumDesiredUTXOValue: entity.MinimumDesiredUTXOValue,
+	}
+
+	err = s.db.WithContext(ctx).Transaction(func(tx *gorm.DB) error {
+		numID, err := s.saveNumericIDForOutputBasket(ctx, tx, entity.UserID, entity.Name)
+		if err != nil {
+			return err
```
ac4ch, line 64:
```suggestion
			return fmt.Errorf("failed to save numeric ID for output basket [%d.%s]: %w", entity.UserID, entity.Name, err)
```

pkg/internal/storage/repo/syncrepo/sync_knowntx.go
```diff
@@ -0,0 +1,171 @@
...
+	if err != nil {
+		return false, fmt.Errorf("transaction failed: %w", err)
+	}
+
+	return isNew, nil
+}
+
+func (s *SyncKnownTx) whereExistsScope(userID int) func(*gorm.DB) *gorm.DB {
+	return func(db *gorm.DB) *gorm.DB {
+		whereExistClause := fmt.Sprintf(
```
ac4ch, line 93: consider using `clause.Expr` :)

pkg/internal/storage/repo/syncrepo/sync_label.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+func (s *SyncLabel) DeleteLabelForSync(ctx context.Context, entity *entity.Label) (deleted bool, err error) {
+	return s.common.Delete(ctx, entity.UserID, entity.Name)
+}
+
+func (s *SyncLabel) FindLabelByNumIDForSync(ctx context.Context, numID uint) (*entity.Label, error) {
+	model, err := s.common.FindByNumID(ctx, numID)
+	if err != nil {
+		return nil, err
```
ac4ch, line 65:
```suggestion
		return nil, fmt.Errorf("failed to find known transaction by numeric ID %d: %w", numID, err)
```

pkg/internal/storage/repo/syncrepo/sync_label_map.go
```diff
@@ -0,0 +1,67 @@
...
+
+type LabelsMapReadModel struct {
+	models.TransactionLabel
+	NumID uint
+}
+
+func (s *SyncLabelMap) FindLabelsMapForSync(ctx context.Context, userID int, opts ...queryopts.Options) ([]*wdk.TableTxLabelMap, error) {
+	result, err := s.common.FindChunk(ctx, userID, opts...)
+	if err != nil {
+		return nil, err
```
ac4ch, line 43:
```suggestion
		return nil, fmt.Errorf("failed to find chunk for user %d with opts %+v: %w", userID, opts, err)
```

pkg/internal/storage/repo/syncrepo/sync_output.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+
+	err = tx.Create(&model).Error
+	if err != nil {
+		err = fmt.Errorf("failed to create output: %w", err)
+		return
+	}
+
+	isNew = true
+	outputID = model.ID
```
ac4ch, line 174:
```go
if outputID == 0 {
		return false, 0, fmt.Errorf("output ID was not retrieved after upsert")
	}
```
i would be here more defencive .. but this should never happen .. so i think u can decide if you think is good to add this :)

pkg/internal/storage/repo/syncrepo/find_chunk.go
chris-4chain: This doesn't change anything. 
Is it safer?: `no`

pkg/internal/storage/repo/syncrepo/label_tag_map_commons.go
chris-4chain: Where exactly do you see potential risk for `SQL Injection` ?
FYI: `f.relationUserIDColumn, f.relationNameColumn` originate from statically generated columns.

pkg/internal/storage/repo/syncrepo/numeric_id.go
chris-4chain: 1. I do use `Vars` to pass arguments to this expression
2. TableName is static (not provided from external)

pkg/internal/storage/repo/syncrepo/sync_basket.go
chris-4chain: Auggie, why??
This is just static text. Your suggestion doesn't protect anything.

pkg/internal/storage/repo/syncrepo/sync_basket.go
chris-4chain: In this case - no.

pkg/internal/storage/repo/syncrepo/sync_basket.go
chris-4chain: The error already is provided with context and originates from the same package (already wrapped). 
- See implementation of: `(s *SyncBasket) saveNumericIDForOutputBasket`

pkg/internal/storage/repo/syncrepo/sync_knowntx.go
chris-4chain: 1. SQL injection is not a case here.
2. This won't work.

pkg/internal/storage/repo/syncrepo/sync_label.go
chris-4chain: 1. The error is already provided with context. (the message is almost the same as you're suggesting)
2. The error originates from the same package

pkg/internal/storage/repo/syncrepo/sync_label_map.go
chris-4chain: Same comment ☝️

pkg/internal/storage/repo/syncrepo/sync_output.go
chris-4chain: Actually, it's a good idea 👍 
- I've added your suggestion and it helped me find (and fix) one edge case.

pkg/internal/storage/repo/syncrepo/find_chunk.go
ac4ch: sure `modelPtr := new(Model)` can make the intent clearer to me .. but ok in behave i agree with u

pkg/internal/storage/repo/syncrepo/label_tag_map_commons.go
ac4ch: i pointed this out as to construct query u are using `fmt.Sprintf` and `CONCAT` which create dynamic SQL .. that why i recommended to use `clause.Expr` if possible

pkg/internal/storage/repo/syncrepo/sync_knowntx.go
ac4ch: yes in this case for sure not .. was just pointing if it would be possible to use clause.Expr here .. if you think no .. that ok

pkg/internal/storage/repo/syncrepo/sync_label.go
ac4ch: ya, but having them wrapped is more than just about adding context ... it preserves the full error chain, which helps with:
`errors.Is` to work at higher layers .. easier tracing in logs and consistent propagation patterns (especially across package boundaries in the future) :) .. that why i pointing unwrapped errors.

pkg/internal/storage/repo/syncrepo/sync_label_map.go
ac4ch: same answer as above ....

pkg/internal/storage/repo/syncrepo/sync_basket.go
ac4ch: ya, but having them wrapped is more than just about adding context ... it preserves the full error chain, which helps with:
errors.Is to work at higher layers .. easier tracing in logs and consistent propagation patterns (especially across package boundaries in the future) :) .. that why i pointing unwrapped errors.

pkg/internal/storage/repo/syncrepo/sync_basket.go
ac4ch: :) fair...  yeah, right now its just static text and not vulnerable... but using `clause.Expr` here builds the habit of expressing SQL structure explicitly and lets *GORM* handle escaping and dialect differences....
but if these column names were ever made dynamic or reused in other places,... the `Expr` form would already be safer and clearer... and is more about building consistent and safer habits i would say :+1:

pkg/internal/storage/repo/syncrepo/find_chunk.go
chris-4chain: But it's not foreign key field. This is a `stringID` which is stored in the table: 
image

pkg/internal/storage/repo/syncrepo/find_chunk.go
chris-4chain: The point is, a value passed to this function is not `tx`.
I could rename it to `dry` or `dryDB` or `drySession` - but I'm not sure if it makes it more explanatory

pkg/internal/storage/repo/syncrepo/sync_basket.go
chris-4chain: My reasoning for not wrapping here is based on our team's convention, which is primarily to wrap errors that come from **external packages** (as enforced by our [wrapcheck linter](https://github.com/tomarrell/wrapcheck)).

The error from `saveNumericIDForOutputBasket` originates within **our own package**; it's essentially an internal utility. I'm already providing context, and since we're not checking for a specific underlying error type with `errors.Is` in the calling functions, the added complexity of wrapping doesn't seem necessary for this use case.

Preserving the error chain is most critical when an error crosses an abstraction or package boundary, which isn't happening here.

ac4ch (approved):

PR #395, opened 2025-07-21
feat: add abort action storage

### LINKED ISSUE #345: [Task] Storage: AbortAction

jakubmkowalski:
Closes #345
This pull request introduces significant changes to support an "Abort Action" feature for transactions in progress. It includes new functionality for aborting transactions, validation logic, and updates to transaction management methods. Additionally, several refactorings and enhancements were made to improve code structure and testing capabilities.

### New Abort Action Feature:
* [`pkg/internal/fixtures/default_abort_action.go`](diffhunk://#diff-a531be6724031e5b6929102591c4cddbdc42343a00d5210d947e970999ffac58R1-R45): Added default test fixtures for `AbortActionArgs` and `AbortActionResult` to facilitate testing of the abort feature.
* [`pkg/storage/internal/actions/abort.go`](diffhunk://#diff-9cda57e1464cf412f67e91af2cdd3eb6ff2de6f01548d230293df6d9788e0443R1-R91): Implemented the core logic for aborting transactions, including validation of transaction status and updates to transaction states.
* [`pkg/internal/validate/validate_abort_action.go`](diffhunk://#diff-23182a6f2da432dafd807f195c5c0b532a3ba400712bed5e733c2e3c6e0e421eR1-R19): Added validation logic for `AbortActionArgs` to ensure proper input for aborting transactions.

### Updates to Transaction Management:
* [`pkg/internal/storage/repo/transactions.go`](diffhunk://#diff-f4b526ed9ad7b3f1f6240c510c35ae64860702054261613ab2b51ef58eb8a4dcL319-R480): Refactored transaction status update methods to separate logic for updating by `txID` and internal `transactionID`. Added methods for cleaning up transaction outputs and releasing reserved UTXOs.
* [`pkg/storage/internal/actions/repos.interface.go`](diffhunk://#diff-94f32606e818054d510d6fadc9d60851e34ade46a455a66f8dee6f35f70f6a03R30-R52): Updated the `TransactionsRepo` interface to include new methods for finding transactions by reference and updating transaction status by ID.

### Mock and Client Enhancements:
* [`pkg/internal/mocks/mock_wallet_storage_writer.go`](diffhunk://#diff-d4a15600e36c87de294b6ce1f4a7e7407037840eed3f16d2fc5e596ae1532765R44-R58): Added mock methods for `AbortAction` to support testing of the abort feature.
* [`pkg/storage/client_gen.go`](diffhunk://#diff-43960e1c778173ecad4d9320004cfd58c26530c3e2dc8f012195aed885029b4aR93-R97): Added `AbortAction` method to the `WalletStorageProviderClient` for client-side interaction with the abort feature. [[1]](diffhunk://#diff-43960e1c778173ecad4d9320004cfd58c26530c3e2dc8f012195aed885029b4aR93-R97) [[2]](diffhunk://#diff-43960e1c778173ecad4d9320004cfd58c26530c3e2dc8f012195aed885029b4aR114)

### Testability Improvements:
* [`pkg/storage/internal/testabilities/assertions_db_state.go`](diffhunk://#diff-fc7a5f94d98dfb749d5b4fe804c082f5710837763c7a207efdc6aae5a90cacd5R55): Added a new assertion method `WithTxIDCount` for validating outputs associated with specific transactions in tests. [[1]](diffhunk://#diff-fc7a5f94d98dfb749d5b4fe804c082f5710837763c7a207efdc6aae5a90cacd5R55) [[2]](diffhunk://#diff-fc7a5f94d98dfb749d5b4fe804c082f5710837763c7a207efdc6aae5a90cacd5R246-R258)

### Minor Changes:
* [`pkg/internal/testabilities/dbfixtures/db.go`](diffhunk://#diff-8837b63d4f3f1ba1b8362ad079cbae01c917ebc6b80f249ec5d4e1798c9a4bcfL34-R35): Modified SQLite connection string for tests to remove memory mode.
## Description of Changes

## Linked Issues / Tickets

Reference any related issues or tickets, e.g. "".

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter


pkg/internal/storage/repo/transactions.go
```diff
@@ -225,6 +224,32 @@ func (txs *Transactions) FindTransactionByReference(ctx context.Context, userID
 
 	return txs.mapModelToTransactionEntity(&transaction), nil
 }
+func (txs *Transactions) FindUniqueTransactionByReference(ctx context.Context, userID int, reference string) (*entity.Transaction, error) {
+	transaction := []models.Transaction{}
```
mgosek-4chain: You can also use zero value `var transactions []models.Transaction`

pkg/storage/internal/testabilities/assertions_db_state.go
```diff
@@ -311,13 +312,26 @@ func (d *outputsListAssertion) WithCount(expected int) OutputsListAssertion {
...
 
+func (d *outputsListAssertion) WithTxIDCount(expected int, txID uint) OutputsListAssertion {
+	d.Helper()
+	txIDStr := to.StringFromInteger(txID)
+	count := seq.Count(seq.Filter(seq.FromSlice(d.outputs), func(output *wdk.WalletOutput) bool {
+		id := output.Outpoint.MustGetTxID()
+
+		equal := txIDStr == id
+		return equal
+	}))
```
mgosek-4chain: Why not just use a simple for loop? 🤔

mgosek-4chain (changes requested):

pkg/storage/internal/testabilities/assertions_db_state.go
jakubmkowalski: The implementation is similar to WithCountHavingOutpoint. 
If this is only a suggestion, I would leave it as it is, unless someone else is not satisfied with this implementation.

pkg/internal/storage/repo/transactions.go
jakubmkowalski: changed

pkg/storage/internal/testabilities/assertions_db_state.go
chris-4chain: It seems ok to me - especially if similar thing is already in the code.

pkg/internal/storage/repo/transactions.go
```diff
@@ -315,32 +340,132 @@ func makeOutputsSpendable(tx *gorm.DB, updatedTx entity.UpdatedTx) error {
...
+		return fmt.Errorf("failed to update transaction status for txID %s: %w", txID, err)
+	}
+	return nil
+}
+
+func (txs *Transactions) UpdateTransactionStatusByID(
+	ctx context.Context,
+	transactionID uint,
+	newStatus wdk.TxStatus,
+	provenTxReqStatus wdk.ProvenTxReqStatus,
```
chris-4chain: This `provenTxReqStatus` is not used. Please make sure if it's required (logically).
It not - remove from the args list

pkg/internal/storage/repo/transactions.go
```diff
@@ -315,32 +340,132 @@ func makeOutputsSpendable(tx *gorm.DB, updatedTx entity.UpdatedTx) error {
...
+
+func (txs *Transactions) updateTransactionStatusByTxID(tx *gorm.DB, txID string, newStatus wdk.TxStatus) error {
+	return tx.Model(&models.Transaction{}).
+		Where("tx_id = ?", txID).
+		Updates(map[string]any{
+			"status": newStatus,
+		}).Error
+}
+
+func (txs *Transactions) updateTransactionStatusByInternalID(tx *gorm.DB, transactionID uint, newStatus wdk.TxStatus) error {
```
chris-4chain: Why not `updateTransactionStatusByID` ?
- Why the `"internal"` word?

pkg/internal/storage/repo/transactions.go
```diff
@@ -315,32 +340,132 @@ func makeOutputsSpendable(tx *gorm.DB, updatedTx entity.UpdatedTx) error {
...
+func (txs *Transactions) deleteOutputsByTransactionID(tx *gorm.DB, transactionID uint) error {
+	return tx.Delete(&models.Output{}, "transaction_id = ?", transactionID).Error
+}
+
+func (txs *Transactions) releaseOutputsReservedByTransaction(tx *gorm.DB, transactionID uint) error {
+	return tx.Model(&models.Output{}).
+		Where("spent_by = ?", transactionID).
 		Updates(map[string]any{
-			"status": txStatus,
+			"spent_by":  nil,
```
chris-4chain, line 410: Hmmmm - what if the output is already spent (`spent_by` is not nil)?

pkg/storage/internal/actions/abort.go
```diff
@@ -0,0 +1,88 @@
...
+	if txEntity == nil {
+		return nil, fmt.Errorf("failed to find unique transaction by reference: expected exactly one transaction with reference %s, found 0", *args.Reference)
+	}
+
+	err = validateTxForAbort(txEntity)
+	if err != nil {
+		return nil, err
+	}
+
+	history := history.NewBuilder().AbortAction(*args.Reference)
```
chris-4chain: Maybe `historyNote` or `note` as a variable name, because `"history"` is already a package name

pkg/internal/storage/history/history_note.go
```diff
@@ -16,6 +16,8 @@ const (
 	InternalizeActionHistoryNote = "internalizeAction"
 	ProcessActionHistoryNote     = "processAction"
 	AggregateResultsHistoryNote  = "aggregateResults"
+	AbortActionHistoryNote       = "abortAction"
+	ActionHistoryNote            = "action"
```
chris-4chain: This `ActionHistoryNote` seems unused

pkg/internal/storage/history/history_note.go
```diff
@@ -112,6 +118,12 @@ func (b *builder) PostBeefSuccess(serviceName string, beef []byte, txIDs []strin
 	return b.postBeefBase(PostBeefSuccess, serviceName, beef, txIDs)
 }
 
+func (b *builder) AbortAction(reference string) Builder {
+	return b.WithWhat(AbortActionHistoryNote).
+		WithAttribute(actionAttr, abortActionAttr).
```
chris-4chain: This is weird / inconsistent. 
Why additional `actionAttr`? We have already the information about "what action is made" in `"AbortActionHistoryNote"`

pkg/storage/internal/actions/repos.interface.go
```diff
@@ -28,6 +28,7 @@ type TransactionsRepo interface {
 	CreateTransaction(ctx context.Context, transaction *entity.NewTx) error
 	FindTransactionByUserIDAndTxID(ctx context.Context, userID int, txID string) (*entity.Transaction, error)
 	FindTransactionByReference(ctx context.Context, userID int, reference string) (*entity.Transaction, error)
+	FindUniqueTransactionByReference(ctx context.Context, userID int, reference string) (*entity.Transaction, error)
```
chris-4chain: This seems redundant to `FindTransactionByReference`.
The second method is not necessary. 
However, if you think that there should be also checking of "uniquness" (like, `len(transactions) != 1`) - feel free to add it to `FindTransactionByReference`

pkg/storage/internal/actions/abort.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+}
+
+func (a *abortAction) AbortAction(ctx context.Context, userID int, args *wdk.AbortActionArgs) (*wdk.AbortActionResult, error) {
+	txEntity, err := a.transactionsRepo.FindUniqueTransactionByReference(ctx, userID, *args.Reference)
+	if err != nil {
+		return nil, fmt.Errorf("failed to find unique transaction by reference: %w", err)
+	}
+
+	if txEntity == nil && len(*args.Reference) == 64 {
```
chris-4chain: It seems like magic - IMHO we should provide some `NOTE` comment or separate this logic to another function (also with the comment)

pkg/storage/internal/actions/abort.go
```diff
@@ -0,0 +1,88 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to update transaction status: %w", err)
+	}
+
+	return &wdk.AbortActionResult{
+		Aborted: true,
+	}, nil
+}
+
+func validateTxForAbort(txEntity *entity.Transaction) error {
```
chris-4chain: it could be a private method for `(a *abortAction)`.
Then, the name could be shorter, like: `validateTx`

pkg/storage/internal/actions/abort.go
```diff
@@ -0,0 +1,88 @@
...
+	if !txEntity.IsOutgoing {
+		return fmt.Errorf("txDirectionInvalid: must be an outgoing transaction.")
```
chris-4chain: Warning from my IDE:
`Error string should not be capitalized or end with punctuation mark`.

Could you remove the period character `"."`?

pkg/storage/internal/actions/abort.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	return validateTxStatusForAbort(txEntity.Status)
+}
+
+func validateTxStatusForAbort(txStatus wdk.TxStatus) error {
```
chris-4chain, line 65: I would suggest using a `switch` statement for this.
It is more performant:

```golang
func validateTxStatusForAbort(txStatus wdk.TxStatus) error {
	switch txStatus {
	case wdk.TxStatusCompleted, wdk.TxStatusFailed, wdk.TxStatusSending, wdk.TxStatusUnproven:
		return fmt.Errorf("txNotAbortable: action with status %s cannot be aborted", txStatus)
	case wdk.TxStatusUnprocessed, wdk.TxStatusUnsigned, wdk.TxStatusNoSend, wdk.TxStatusNonFinal, wdk.TxStatusUnfail:
		fallthrough
	default:
		return nil
	}
}
```

The second case is needed, because otherwise the linter screams "exhaustive".

... or you can disable linter-exhaustive for this switch:
```golang
	//nolint:exhaustive /
```

pkg/internal/storage/repo/transactions.go
```diff
@@ -315,32 +340,132 @@ func makeOutputsSpendable(tx *gorm.DB, updatedTx entity.UpdatedTx) error {
...
+}
+
+func (txs *Transactions) UpdateTransactionStatusByID(
+	ctx context.Context,
+	transactionID uint,
+	newStatus wdk.TxStatus,
+	provenTxReqStatus wdk.ProvenTxReqStatus,
+) error {
+	err := txs.db.WithContext(ctx).Transaction(func(tx *gorm.DB) error {
+		if newStatus == wdk.TxStatusFailed {
```
chris-4chain: This function `UpdateTransactionStatusByID` mixes responsibilities. Actually, it doesn't "only" "update transaction status by id".

chris-4chain (changes requested):

pkg/wdk/storage_abort_action_args.go
```diff
@@ -0,0 +1,13 @@
+package wdk
+
+// AbortActionArgs defines the arguments for aborting a wallet action.
+type AbortActionArgs struct {
+	// Reference is the unique identifier for the action to be aborted.
+	Reference *string
```
dorzepowski: 1. this doesn't seem to be optional in typescript, so it shouldn't be a pointer
2. In typescript validation it seems that it is required to be Base64 encoded, so you can use primitives.Base64String type to simplify validation later

pkg/storage/provider.go
```diff
@@ -505,3 +505,20 @@ func (p *Provider) CommissionEntity() crud.Commission {
 func (p *Provider) KnownTxEntity() crud.KnownTx {
 	return crud.NewKnownTx(p.repo.KnownTx)
 }
+
+// AbortAction aborts an action by its reference for the authenticated user.
+func (p *Provider) AbortAction(ctx context.Context, auth wdk.AuthID, args wdk.AbortActionArgs) (*wdk.AbortActionResult, error) {
```
dorzepowski: please move it closer to the rest of the XyzAction methods (for example just after ProcessAction)

pkg/internal/validate/validate_abort_action.go
```diff
@@ -0,0 +1,19 @@
...
+
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/wdk"
+)
+
+func ValidAbortActionArgs(args *wdk.AbortActionArgs) error {
+	if args == nil {
+		return fmt.Errorf("args cannot be nil")
+	}
+
+	if args.Reference == nil {
```
dorzepowski: Check for empty string.

Hint:
```suggestion
	if is.BlankString(args.Reference) {
```

pkg/storage/internal/actions/abort.go
```diff
@@ -0,0 +1,88 @@
...
+)
+
+type abortAction struct {
+	logger           *slog.Logger
+	transactionsRepo TransactionsRepo
+}
+
+func newAbortAction(logger *slog.Logger, transactions TransactionsRepo) *abortAction {
+	return &abortAction{
+		logger:           logging.Child(logger, "abort_action"),
```
dorzepowski: ### minor

We don't have strict convention (worth to talk about it later and adjust all), but in other XyzAction we use camel case for service name

```suggestion
		logger:           logging.Child(logger, "abortAction"),
```

pkg/storage/internal/actions/abort.go
dorzepowski: I would solve this with extracting a method to use here like
```suggestion
if txEntity == nil && a.isPotentiallyTxID(args.Reference) {
```

pkg/storage/internal/actions/abort.go
```diff
@@ -0,0 +1,88 @@
...
+func validateTxStatusForAbort(txStatus wdk.TxStatus) error {
+	var unAbortableStatuses = map[wdk.TxStatus]bool{
+		wdk.TxStatusCompleted: true,
+		wdk.TxStatusFailed:    true,
+		wdk.TxStatusSending:   true,
+		wdk.TxStatusUnproven:  true,
+	}
+
+	if unAbortableStatuses[txStatus] {
+		return fmt.Errorf("txNotAbortable: action with status %s cannot be aborted", txStatus)
```
dorzepowski: It looks like you had some intuition or something that you're trying to distinguish not abortable errors from others.

I would suggest to create a public error like:
```go
var ErrNotAbortableAction = fmt.Errorf("action cannot be aborted")
```


And then use it here:
```suggestion
		return fmt.Errorf("%w: action with status %s cannot be aborted", wdk.ErrNotAbortableAction,  txStatus)
```

pkg/internal/storage/repo/transactions.go
```diff
@@ -315,32 +340,132 @@ func makeOutputsSpendable(tx *gorm.DB, updatedTx entity.UpdatedTx) error {
...
+	return nil
+}
+
+func (txs *Transactions) deleteOutputsByTransactionID(tx *gorm.DB, transactionID uint) error {
+	return tx.Delete(&models.Output{}, "transaction_id = ?", transactionID).Error
+}
+
+func (txs *Transactions) releaseOutputsReservedByTransaction(tx *gorm.DB, transactionID uint) error {
+	return tx.Model(&models.Output{}).
+		Where("spent_by = ?", transactionID).
```
dorzepowski, line 408: @chris-4chain can we somehow use gorm gen result for those new queries?

pkg/storage/internal/actions/abort.go
```diff
@@ -0,0 +1,88 @@
...
+	history := history.NewBuilder().AbortAction(*args.Reference)
+
+	if txEntity.TxID != nil {
+		err = a.transactionsRepo.UpdateTransactionStatusForTxID(ctx, *txEntity.TxID, wdk.TxStatusFailed, wdk.ProvenTxStatusInvalid, history)
+		if err != nil {
+			return nil, fmt.Errorf("failed to update known transaction status: %w", err)
+		}
+	}
+
+	err = a.transactionsRepo.UpdateTransactionStatusByID(ctx, txEntity.ID, wdk.TxStatusFailed, wdk.ProvenTxStatusInvalid)
```
dorzepowski: maybe I miss something, but if the txEntity has TxID, then we're updating the status twice?

IMHO you should use separated method for UpdateKnownTxStatus in case TxID exists.

pkg/storage/provider_abort_action_test.go
```diff
@@ -0,0 +1,272 @@
...
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/wdk"
+	"github.com/go-softwarelab/common/pkg/to"
+	"github.com/stretchr/testify/require"
+)
+
+const NumberOfDesiredUTXOs = 32
+
+func TestAbortActionSuccess(t *testing.T) {
+	// given:
+	given, cleanup := testabilities.GivenCustomStorage(t, fixtures.StorageServerPrivKey, "dbstorage_test")
```
dorzepowski: You don't need to use CustomStorage, it's used for synchronisation where you need 2 storages (look at the tests for CreateActions for example)

pkg/storage/provider_abort_action_test.go
```diff
@@ -0,0 +1,272 @@
...
+	faucet := given.Faucet(activeStorage, testusers.Alice)
+	faucet.TopUp(100_000)
+
+	createResult, err := activeStorage.CreateAction(
+		t.Context(),
+		testusers.Alice.AuthID(),
+		fixtures.DefaultValidCreateActionArgs(),
+	)
+	require.NoError(t, err)
+	testabilities.ThenDBState(t, activeStorage).AllOutputs(testusers.Alice).WithCount(1 + NumberOfDesiredUTXOs)
```
dorzepowski: No need to check the createAction result here - the tests for create action should test it

pkg/storage/internal/testabilities/assertions_db_state.go
```diff
@@ -311,13 +312,26 @@ func (d *outputsListAssertion) WithCount(expected int) OutputsListAssertion {
 	return d
 }
 
+func (d *outputsListAssertion) WithTxIDCount(expected int, txID uint) OutputsListAssertion {
```
dorzepowski: It looks like it is not used anywhere

pkg/storage/internal/testabilities/assertions_db_state.go
dorzepowski: Look at my comment for whole method 😀

https://github.com/bsv-blockchain/go-wallet-toolbox/pull/395/files/e27c98813b962666439321096bd6d906c0da5415#r2221503488

pkg/storage/provider_abort_action_test.go
```diff
@@ -0,0 +1,272 @@
...
+	given, cleanup := testabilities.GivenCustomStorage(t, fixtures.StorageServerPrivKey, "dbstorage_test")
+
+	defer cleanup()
+
+	activeStorage := given.Provider().GORM()
+
+	given.Faucet(activeStorage, testusers.Alice).TopUp(100_000)
+	createdTx, _ := given.ActionCreatedAndSigned(activeStorage)
+	txID := createdTx.Inputs[0].SourceTxID
+	testabilities.ThenDBState(t, activeStorage).AllOutputs(testusers.Alice).WithCount(1 + NumberOfDesiredUTXOs)
```
dorzepowski: again, not needed check

pkg/storage/provider_abort_action_test.go
```diff
@@ -0,0 +1,272 @@
...
+
+func TestAbortActionTransactionNotOutgoing(t *testing.T) {
+	// given:
+	given, cleanup := testabilities.GivenCustomStorage(t, fixtures.StorageServerPrivKey, "dbstorage_test")
+
+	defer cleanup()
+
+	activeStorage := given.Provider().GORM()
+
+	given.Faucet(activeStorage, testusers.Alice).TopUp(100_000)
```
dorzepowski: the top up here is not needed

pkg/storage/provider_abort_action_test.go
```diff
@@ -0,0 +1,272 @@
...
+	createdTx, _ := given.ActionCreatedAndSigned(activeStorage)
```
dorzepowski: Interesting - I would assume that result of this would be exactly the outgoing action, 
Maybe it's worth checking if this is not a bug.

@chris-4chain Was this intentional or we should create an issue to investigate

pkg/storage/provider_abort_action_test.go
```diff
@@ -0,0 +1,272 @@
...
+func TestAbortActionSuccess(t *testing.T) {
+	// given:
+	given, cleanup := testabilities.GivenCustomStorage(t, fixtures.StorageServerPrivKey, "dbstorage_test")
+	defer cleanup()
+
+	activeStorage := given.Provider().GORM()
+	faucet := given.Faucet(activeStorage, testusers.Alice)
+	faucet.TopUp(100_000)
+
+	createResult, err := activeStorage.CreateAction(
```
dorzepowski: It would be great to use 
`given.ActionCreatedAndSigned(activeStorage)` 
In most of the tests here.

pkg/storage/provider_abort_action_test.go
```diff
@@ -0,0 +1,272 @@
...
+		t.Context(),
+		testusers.Alice.AuthID(),
+		wdk.AbortActionArgs{
+			Reference: &txID,
+		},
+	)
+
+	// then:
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "failed to abort action: txDirectionInvalid: must be an outgoing transaction.")
```
dorzepowski: If you will create this public error (as I'm suggesting in other comment), then here you could use require.ErrorIs

pkg/storage/provider_abort_action_test.go
```diff
@@ -0,0 +1,272 @@
...
+
+	thenDBState := testabilities.ThenDBState(t, activeStorage)
+	thenDBState.HasUserTransactionByReference(testusers.Alice, createResult.Reference).
+		WithStatus(wdk.TxStatusFailed)
+	thenDBState.AllOutputs(testusers.Alice).WithCount(1)
+}
+
+func TestAbortActionTransactionNotOutgoing(t *testing.T) {
+	// given:
+	given, cleanup := testabilities.GivenCustomStorage(t, fixtures.StorageServerPrivKey, "dbstorage_test")
```
dorzepowski: again, no need for custom storage

pkg/storage/provider_abort_action_test.go
```diff
@@ -0,0 +1,272 @@
...
+			Reference: to.Ptr("123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890"),
+		},
+	)
+
+	// then:
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "found 0")
+}
+
+func TestAbortActionTransactionStatusFailed(t *testing.T) {
```
dorzepowski: The tests for different statuses could (and should) be changed to table tests

pkg/internal/fixtures/default_abort_action.go
```diff
@@ -0,0 +1,45 @@
...
+		Reference: to.Ptr(ShortReference),
+	}
+}
+
+// DefaultAbortActionResult returns default AbortActionResult for testing
+func DefaultAbortActionResult() *wdk.AbortActionResult {
+	return &wdk.AbortActionResult{
+		Aborted: true,
+	}
+}
```
dorzepowski: all those methods seem to be not used

pkg/internal/validate/validate_abort_action.go
```diff
@@ -0,0 +1,19 @@
+package validate
+
+import (
+	"fmt"
+
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/wdk"
+)
+
+func ValidAbortActionArgs(args *wdk.AbortActionArgs) error {
```
dorzepowski: We should have tests for that

pkg/storage/provider_abort_action_test.go
dorzepowski: @chris-4chain nvm

Ohh, I see that I overlooked next line. So you just taking the parent tx.

I'm suggesting to either 
 use the txID from faucet topup
OR add some better spacing to better expose the line from where the txID is comming from.


Additionally, it seems that you can use this `given.ActionCreatedAndSigned(activeStorage)` in all the other tests, instead of plain call to CreateAction

pkg/internal/fixtures/default_abort_action.go
jakubmkowalski: removed unused

pkg/internal/storage/history/history_note.go
jakubmkowalski: removed

pkg/internal/storage/history/history_note.go
jakubmkowalski: simplified to:
```
		return b.WithWhat(AbortActionHistoryNote).
		WithAttribute(referenceAttr, reference)
```

pkg/internal/validate/validate_abort_action.go
jakubmkowalski: as `Reference` is now type `primitives.Base64String` changed to:
```
if args.Reference == "" {
```

pkg/internal/validate/validate_abort_action.go
jakubmkowalski: tests added

pkg/storage/internal/actions/abort.go
jakubmkowalski: changed

pkg/storage/internal/actions/abort.go
jakubmkowalski: changed to `historyNote`

pkg/storage/internal/actions/abort.go
jakubmkowalski: removed

pkg/storage/internal/actions/abort.go
jakubmkowalski: changed

pkg/storage/internal/actions/abort.go
jakubmkowalski: changed

pkg/storage/internal/actions/abort.go
jakubmkowalski: changed to switch case

pkg/storage/internal/actions/abort.go
jakubmkowalski: method extracted

pkg/internal/storage/repo/transactions.go
jakubmkowalski: It is ID from database, i didn't want to use TransactionID nor TxID but you are right simple ID will be ok.
Changed to `updateTransactionStatusByID`

pkg/internal/storage/repo/transactions.go
jakubmkowalski: removed

pkg/storage/internal/testabilities/assertions_db_state.go
jakubmkowalski: removed

pkg/storage/provider.go
jakubmkowalski: moved

pkg/storage/provider_abort_action_test.go
jakubmkowalski: now using default db

pkg/storage/provider_abort_action_test.go
jakubmkowalski: now using default db

pkg/storage/provider_abort_action_test.go
jakubmkowalski: removed

pkg/wdk/storage_abort_action_args.go
jakubmkowalski: 1. Now `Reference` is not a pointer.
2. I would do this as separate task as this requires some more refactoring

pkg/storage/provider_abort_action_test.go
jakubmkowalski: removed

pkg/storage/provider_abort_action_test.go
jakubmkowalski: changed to table tests

pkg/storage/provider_abort_action_test.go
jakubmkowalski: now I use 
given.Faucet(activeStorage, testusers.Alice).TopUp(100_000) 
or 
given.ActionCreatedAndSigned(activeStorage)
but not both in single test

pkg/storage/provider_abort_action_test.go
jakubmkowalski: tests now looks little different, i don't know if this applies now

pkg/storage/provider_abort_action_test.go
jakubmkowalski: Now I hope that the readability has improved.

pkg/storage/provider_abort_action_test.go
jakubmkowalski: now when I can, I use `given.ActionCreatedAndSigned(activeStorage)`

pkg/internal/storage/repo/transactions.go
jakubmkowalski: You are right, when `spent_by` in not `null` aborting should fail.

pkg/internal/storage/repo/transactions.go
jakubmkowalski: @chris-4chain bump

pkg/storage/internal/actions/abort.go
jakubmkowalski: refactored

pkg/internal/storage/repo/transactions.go
jakubmkowalski: refactored

pkg/storage/internal/actions/repos.interface.go
jakubmkowalski: I added this condition to the `FindTransactionByReference` function and removed `FindUniqueTransactionByReference`. 
I want to be sure that there is no more than one transaction with this reference. 
If you know that such a situation is impossible or very unlikely, I can remove this condition.

pkg/internal/storage/repo/transactions.go
```diff
@@ -485,3 +554,47 @@ func (txs *Transactions) labelFilterScope(tx *gorm.DB, userID int, filter entity
...
+
+		return nil
+	})
+
+	if err != nil {
+		return fmt.Errorf("failed to abort transaction: %w", err)
+	}
+
+	return nil
+}
```
ac4ch:
```suggestion
func (txs *Transactions) AbortTransactionAtomic(ctx context.Context, transactionID uint, txID *string, reference string) error {
	if err := txs.db.WithContext(ctx).Transaction(func(tx *gorm.DB) error {
		txNotes := history.NewBuilder().AbortAction(reference).Note().AsList()
		historyBuilders := make([]history.Builder, len(txNotes))
		for i, note := range txNotes {
			historyBuilders[i] = history.NewBuilderFromNote(note)
		}

		steps := []func(*gorm.DB) error{
			func(tx *gorm.DB) error {
				return txs.releaseReservedUTXOs(tx, transactionID)
			},
			func(tx *gorm.DB) error {
				return txs.checkIfAnyOutputIsSpent(tx, transactionID)
			},
			func(tx *gorm.DB) error {
				return txs.releaseOutputsReservedByTransaction(tx, transactionID)
			},
			func(tx *gorm.DB) error {
				return txs.deleteOutputsByTransactionID(tx, transactionID)
			},
			func(tx *gorm.DB) error {
				return txs.updateTransactionStatusByID(tx, transactionID, wdk.TxStatusFailed)
			},
		}

		for _, step := range steps {
			if err := step(tx); err != nil {
				return err
			}
		}

		if txID != nil && *txID != "" {
			if err := updateKnownTxStatus(tx, *txID, wdk.ProvenTxStatusInvalid, historyBuilders); err != nil {
				return err
			}
		}

		return nil
	}); err != nil {
		return fmt.Errorf("failed to abort transaction: %w", err)
	}

	return nil
}
```

consider to improve the readability ... i think is slightly cleaner coz it reduces nesting by using the `if err`

pkg/storage/internal/actions/abort.go
```diff
@@ -0,0 +1,83 @@
...
+		}
+	}
+
+	if txEntity == nil {
+		return nil, fmt.Errorf("failed to find unique transaction by reference: expected exactly one transaction with reference %s, found 0", args.Reference)
+	}
+
+	err = a.validateTx(txEntity)
+	if err != nil {
+		return nil, err
```
ac4ch:
```suggestion
		return nil, fmt.Errorf("transaction validation failed: %w", err)
```

pkg/storage/internal/actions/abort.go
```diff
@@ -0,0 +1,83 @@
...
+		args.Reference,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("failed to abort transaction: %w", err)
+	}
+
+	return &wdk.AbortActionResult{
+		Aborted: true,
+	}, nil
+}
```
ac4ch:
```suggestion
func (a *abortAction) AbortAction(ctx context.Context, userID int, args *wdk.AbortActionArgs) (*wdk.AbortActionResult, error) {
	txEntity, err := a.transactionsRepo.FindTransactionByReference(ctx, userID, args.Reference)
	if err != nil {
		return nil, fmt.Errorf("failed to find unique transaction by reference: %w", err)
	}

	if txEntity == nil && a.isPotentiallyTxID(args.Reference) {
		txEntity, err = a.transactionsRepo.FindTransactionByUserIDAndTxID(ctx, userID, args.Reference)
		if err != nil {
			return nil, fmt.Errorf("failed to find transaction by txid: %w", err)
		}
	}

	if txEntity == nil {
		return nil, fmt.Errorf("no transaction found with reference or txid %q", args.Reference)
	}

	if err := a.validateTx(txEntity); err != nil {
		return nil, fmt.Errorf("transaction validation failed: %w", err)
	}

	if err := a.transactionsRepo.AbortTransactionAtomic(ctx, txEntity.ID, txEntity.TxID, args.Reference); err != nil {
		return nil, fmt.Errorf("failed to abort transaction: %w", err)
	}

	return &wdk.AbortActionResult{Aborted: true}, nil
}
``` 
or maybe like this WDYT ?

pkg/storage/internal/actions/abort.go
ac4ch: hmm .. what about doing it like this ? 

```go
var abortableStatuses = map[wdk.TxStatus]struct{}{
	wdk.TxStatusUnprocessed: {},
	wdk.TxStatusUnsigned:    {},
	wdk.TxStatusNoSend:      {},
	wdk.TxStatusNonFinal:    {},
	wdk.TxStatusUnfail:      {},
}

func validateTxStatusForAbort(txStatus wdk.TxStatus) error {
	if _, ok := abortableStatuses[txStatus]; ok {
		return nil
	}

	switch txStatus {
	case wdk.TxStatusCompleted, wdk.TxStatusFailed, wdk.TxStatusSending, wdk.TxStatusUnproven:
		return fmt.Errorf("%w: action with status %s cannot be aborted", wdk.ErrNotAbortableAction, txStatus)
	}

	return fmt.Errorf("%w: unexpected transaction status %s", wdk.ErrNotAbortableAction, txStatus)
}
```
i think is easier to see what can be aborted ... and what cannot be ... but that could be just me :)

ac4ch (approved):

pkg/storage/internal/actions/abort.go
chris-4chain: I totally don't get above suggestion.

ac4ch: sure but i do...

pkg/storage/internal/actions/abort.go
jakubmkowalski: I applied your suggestion

pkg/storage/internal/actions/abort.go
jakubmkowalski: changed

pkg/internal/storage/repo/transactions.go
jakubmkowalski: I followed your suggestion with minor changes

pkg/storage/internal/actions/abort.go
chris-4chain: I don't understand why you mixed the `map` and `switch` approach.
I think, the switch is the most efficient and straightforward way to do so.
And thanks to lint-exhaustive we are sure that we cover all possible statuses.

jakubmkowalski: Looking at your implementation, I would allocate two maps, one for statuses that can be aborted and another for statuses that cannot be aborted.
I prefer the implementation with maps, but I will follow Chris suggestion.

ac4ch: because in my opinion is 
clear whitelist of what can be aborted.... 
proper error for unexpected statuses....
separation of known-non-abortables vs. unknown statuses....

as your suggestion you accepting unknown 

anyway up to Jakub what he prefer ... its just suggestion and i do not insist on implementing it

chris-4chain: My original point and the main suggestion is:
- "do not employ maps when you can use switch-case"
- maps have its performance overhead 
- and in this situation I don't see any reason to do so.

chris-4chain: Only thing, that could be adjusted (comparing to my initial suggestion) is to return error on "unknown" status, like:

```golang
func validateTxStatusForAbort(txStatus wdk.TxStatus) error {
	switch txStatus {
	case wdk.TxStatusCompleted, wdk.TxStatusFailed, wdk.TxStatusSending, wdk.TxStatusUnproven:
		return fmt.Errorf("txNotAbortable: action with status %s cannot be aborted", txStatus)
	case wdk.TxStatusUnprocessed, wdk.TxStatusUnsigned, wdk.TxStatusNoSend, wdk.TxStatusNonFinal, wdk.TxStatusUnfail:
		return nil
	default:
		return fmt.Errorf("unknown status %s", txStatus)
	}
}
```

jakubmkowalski: Yes, I added error for "unknown" status earlier. So the implementation now looks like in your last message

pkg/internal/storage/history/history_note.go
```diff
@@ -26,12 +27,16 @@ const (
 const (
 	statusNowAttr   = "status_now"
 	serviceNameAttr = "name"
+	abortActionAttr = "abort_action"
+	actionAttr      = "action"
```
dorzepowski: you forgot to remove them

pkg/storage/provider_abort_action_test.go
```diff
@@ -0,0 +1,296 @@
...
+	require.True(t, result.Aborted)
+
+	thenDBState := testabilities.ThenDBState(t, activeStorage)
+	thenDBState.HasUserTransactionByReference(testusers.Alice, createResult.Reference).
+		WithStatus(wdk.TxStatusUnsigned)
+}
+
+func TestAbortActionErrorCases(t *testing.T) {
+	tests := map[string]struct {
+		setupTransaction func(t *testing.T, given testabilities.StorageFixture) (string, wdk.AuthID)
```
dorzepowski: t *testing.T is not needed in this function, it's not used anywhere

pkg/storage/provider_abort_action_test.go
```diff
@@ -0,0 +1,296 @@
...
+		args             func(reference string) wdk.AbortActionArgs
+		expectedErrors   []string
+	}{
+		"transaction not found by reference": {
+			setupTransaction: func(t *testing.T, given testabilities.StorageFixture) (string, wdk.AuthID) {
+				return "non-existent-reference", testusers.Alice.AuthID()
+			},
+			args: func(reference string) wdk.AbortActionArgs {
```
dorzepowski: I'm almost sure that all those cases have the same implementation of args function

pkg/storage/provider_abort_action_test.go
```diff
@@ -0,0 +1,296 @@
...
+				require.Contains(t, err.Error(), expectedError)
+			}
+		})
+	}
+}
+
+func TestAbortActionAbortableStatuses(t *testing.T) {
+	tests := map[string]struct {
+		setupTransaction func(t *testing.T, given testabilities.StorageFixture) (string, wdk.AuthID)
+		args             func(reference string) wdk.AbortActionArgs
```
dorzepowski: args is not needed

pkg/storage/provider_abort_action_test.go
```diff
@@ -0,0 +1,296 @@
...
+			for _, expectedError := range test.expectedErrors {
+				require.Contains(t, err.Error(), expectedError)
+			}
+		})
+	}
+}
+
+func TestAbortActionAbortableStatuses(t *testing.T) {
+	tests := map[string]struct {
+		setupTransaction func(t *testing.T, given testabilities.StorageFixture) (string, wdk.AuthID)
```
dorzepowski: t *testing.T, is not needed

pkg/internal/storage/repo/transactions.go
dorzepowski: We talked about it, that it is possible, but we should introduce it in separate PR, because it's more change then I thought at the beginning.

pkg/internal/storage/repo/transactions.go
```diff
@@ -485,3 +554,55 @@ func (txs *Transactions) labelFilterScope(tx *gorm.DB, userID int, filter entity
...
+			{"release_outputs_reserved", func() error {
+				return txs.releaseOutputsReservedByTransaction(tx, transactionID)
+			}},
+			{"delete_outputs", func() error {
+				return txs.deleteOutputsByTransactionID(tx, transactionID)
+			}},
+			{"update_transaction_status", func() error {
+				return txs.updateTransactionStatusByID(tx, transactionID, wdk.TxStatusFailed)
+			}},
+		}
```
dorzepowski: ### Not blocking comment, just how I see it.


This should be done differently, to not push the abortion logic to db layer.

We should
- read the whole Transaction with all it's relations as an entity,
- call the method Abort on that entity
- entity should remove and update all the fields that needs update
- pass entity to Save method or something like that on the repo
- and repo based on those changes should update db state.

Although it might look not optimal, it would show clearly the place where the abort logic is applied and what state is changed. Beside that, it shouldn't be the method that is often call, so having it suboptimal shouldn't be a problem.

pkg/storage/provider_abort_action_test.go
jakubmkowalski: removed

pkg/storage/provider_abort_action_test.go
jakubmkowalski: removed

pkg/storage/provider_abort_action_test.go
jakubmkowalski: args removed

pkg/storage/provider_abort_action_test.go
jakubmkowalski: removed

pkg/internal/storage/history/history_note.go
jakubmkowalski: removed

pkg/storage/internal/actions/abort.go
```diff
@@ -0,0 +1,70 @@
...
+		return fmt.Errorf("%w: action with status %s cannot be aborted", wdk.ErrNotAbortableAction, txStatus)
+	case wdk.TxStatusUnprocessed, wdk.TxStatusUnsigned, wdk.TxStatusNoSend, wdk.TxStatusNonFinal, wdk.TxStatusUnfail:
+		return nil
+	default:
+		return fmt.Errorf("%w: unexpected transaction status %s", wdk.ErrNotAbortableAction, txStatus)
+	}
+}
+
+func (a *abortAction) isPotentiallyTxID(reference string) bool {
+	return len(reference) == 64
```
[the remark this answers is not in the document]
jakubmkowalski: changed

pkg/storage/internal/testabilities/assertions_db_state.go
jakubmkowalski: method removed

dorzepowski (dismissed):

pkg/internal/storage/repo/transactions.go
```diff
@@ -485,3 +554,55 @@ func (txs *Transactions) labelFilterScope(tx *gorm.DB, userID int, filter entity
 		return query.Where("id IN (?)", subQuery)
 	}
 }
+
+func (txs *Transactions) AbortTransactionAtomic(ctx context.Context, transactionID uint, txID *string, reference string) error {
```
dorzepowski, line 566: Please check this Issue from sonar: https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=395&id=bsv-blockchain_go-wallet-toolbox&open=AZg7TDpBiKNvHlIFi7PA

Maybe it's good to listen to it and make some refactoring to limit the cognition overhead.

pkg/internal/storage/repo/transactions.go
```diff
@@ -485,3 +554,55 @@ func (txs *Transactions) labelFilterScope(tx *gorm.DB, userID int, filter entity
...
+	if err := txs.db.WithContext(ctx).Transaction(func(tx *gorm.DB) error {
+		txNotes := history.NewBuilder().AbortAction(reference).Note().AsList()
+		historyBuilders := make([]history.Builder, len(txNotes))
+		for i, note := range txNotes {
+			historyBuilders[i] = history.NewBuilderFromNote(note)
+		}
```
dorzepowski: Ohh, thanks to sonar, I just noticed that:
1. you create a builder
2. asked builder to create a note and create a slice with this one note
3. iterate over this slice with one note
4. create a builder from this one note
5. add a builder to a new slice

So it could be simplified and optimised by:
```suggestion
		historyBuilders := []history.Builder{history.NewBuilder().AbortAction(reference)}
```

pkg/internal/storage/repo/transactions.go
```diff
@@ -485,3 +554,55 @@ func (txs *Transactions) labelFilterScope(tx *gorm.DB, userID int, filter entity
...
+			{"update_transaction_status", func() error {
+				return txs.updateTransactionStatusByID(tx, transactionID, wdk.TxStatusFailed)
+			}},
+		}
+
+		for _, s := range steps {
+			if err := s.fn(); err != nil {
+				return fmt.Errorf("AbortTransactionAtomic: step '%s' failed: %w", s.name, err)
+			}
+		}
```
dorzepowski: ### optional suggestion

It could be changed to, and then those steps could be created once as a private var or in a separate function
```suggestion
		type step struct {
			name string
			fn   func(tx *gorm.DB, transactionID uint) error
		}

		steps := []step{
			{"release_reserved_utxos", txs.releaseReservedUTXOs},
			{"check_if_any_output_is_spent", txs.checkIfAnyOutputIsSpent},
			{"release_outputs_reserved", txs.releaseOutputsReservedByTransaction},
			{"delete_outputs", txs.deleteOutputsByTransactionID},
			{"update_transaction_status", func(tx *gorm.DB, transactionID uint) error {
				return txs.updateTransactionStatusByID(tx, transactionID, wdk.TxStatusFailed)
			}},
		}

		for _, s := range steps {
			if err := s.fn(tx, transactionID); err != nil {
				return fmt.Errorf("AbortTransactionAtomic: step '%s' failed: %w", s.name, err)
			}
		}
```

pkg/internal/storage/repo/transactions.go
jakubmkowalski: simplified

pkg/internal/storage/repo/transactions.go
jakubmkowalski: refactored

pkg/internal/storage/repo/transactions.go
jakubmkowalski: refactored

dorzepowski (approved):

pkg/storage/internal/actions/abort.go
```diff
@@ -0,0 +1,78 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to find unique transaction by reference: %w", err)
+	}
+
+	if txEntity == nil && a.isPotentiallyTxID(args.Reference) {
+		txEntity, err = a.transactionsRepo.FindTransactionByUserIDAndTxID(ctx, userID, args.Reference)
+		if err != nil {
+			return nil, fmt.Errorf("failed to find transaction by txid: %w", err)
+		}
+	}
```
chris-4chain: I would make this error as:
`fmt.Errorf("failed to find transaction by txid %s: %w", args.Reference, err)`

And similarly, above, I would also print the reference in the error msg.

pkg/internal/storage/repo/transactions.go
```diff
@@ -206,23 +221,49 @@ func (txs *Transactions) FindTransactionByUserIDAndTxID(ctx context.Context, use
...
 		Preload("Labels").
-		First(&transaction).Error
+		Find(&transactions).
+		Error
 	if err != nil {
 		if errors.Is(err, gorm.ErrRecordNotFound) {
 			return nil, nil
 		}
+
 		return nil, fmt.Errorf("failed to find transaction by reference: %w", err)
```
chris-4chain: AI Generated comment:
The `FindTransactionByReference` function has been changed to fetch a list of transactions but then checks if the list contains exactly one item. This approach can be inefficient and may hide a potential design issue. If the `reference` is intended to be unique per user, it would be better to enforce this with a unique constraint in the database. If multiple transactions can share the same reference, the function should be named `FindTransactionsByReference` to more accurately reflect its behavior. This will make the code clearer and prevent potential misunderstandings.

From actual KT:
@jakubmkowalski FYI: Find don't return `ErrRecordNotFound` on not found.
- But I would suggest to try to use the First item (that get's only one element)

jakubmkowalski: changed to `First`, after changing `Reference` to unique index

pkg/storage/internal/actions/abort.go
jakubmkowalski: added context

pkg/internal/fixtures/default_abort_action.go
```diff
@@ -0,0 +1,17 @@
...
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/wdk"
+)
+
+const (
+	// DefaultAbortActionReference is a test reference for abort action
+	DefaultAbortActionReference = "test-reference-abort"
+)
+
+// DefaultValidAbortActionArgs returns default valid AbortActionArgs for testing
+func DefaultValidAbortActionArgs() wdk.AbortActionArgs {
```
chris-4chain: This function seems unused - and since it's an internal thing, I would suggest to remove it.

pkg/internal/storage/repo/transactions.go
```diff
@@ -207,26 +222,44 @@ func (txs *Transactions) FindTransactionByUserIDAndTxID(ctx context.Context, use
...
 	}
 
 	return txs.mapModelToTransactionEntity(&transaction), nil
 }
 
+func (txs *Transactions) FindTransactionByTxID(ctx context.Context, userID int, txID string) (*entity.Transaction, error) {
+	transaction := models.Transaction{}
+	err := txs.db.WithContext(ctx).
+		Scopes(scopes.UserID(userID)).
+		Where("tx_id = ?", txID).
```
chris-4chain, line 250: Please add `index` on DB column: `txID`

mgosek-4chain (approved):

chris-4chain (approved):

pkg/internal/fixtures/default_abort_action.go
jakubmkowalski: removed

pkg/internal/storage/repo/transactions.go
jakubmkowalski: added

PR #402, opened 2025-07-23
feat: validation no send

chris-4chain:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter


pkg/internal/storage/repo/outputs.go
```diff
@@ -171,6 +171,45 @@ func (o *Outputs) UnlinkOutputFromBasketByOutpoint(ctx context.Context, userID i
...
 }
 
+func (o *Outputs) FindOutputsByOutpoints(ctx context.Context, userID int, outpoints []wdk.OutPoint) ([]*entity.Output, error) {
+	if len(outpoints) == 0 {
+		return nil, nil
+	}
+
+	outpointsConditions := o.db
+	for _, outpoint := range outpoints {
+		outpointsConditions = outpointsConditions.Or("bsv_outputs.vout = ? AND tx.tx_id = ?", outpoint.Vout, outpoint.TxID)
```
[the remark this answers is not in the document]
chris-4chain: Good point - I've figured out another solution

pkg/internal/storage/repo/outputs.go
```diff
@@ -171,6 +172,48 @@ func (o *Outputs) UnlinkOutputFromBasketByOutpoint(ctx context.Context, userID i
...
+		return nil, nil
+	}
+
+	outpointStrings := slices.Map(outpoints, func(op wdk.OutPoint) string {
+		return op.String()
+	})
+
+	outputTableName := genquery.Output.TableName()
+	transactionTableName := genquery.Transaction.TableName()
+
```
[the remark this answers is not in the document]
chris-4chain: Is is just a getter from a field.

pkg/internal/storage/repo/outputs.go
```diff
@@ -171,6 +172,48 @@ func (o *Outputs) UnlinkOutputFromBasketByOutpoint(ctx context.Context, userID i
...
+	query := o.db.WithContext(ctx).Table(
+		"(?) as out",
+		o.db.Model(&models.Output{}).
+			Select(fmt.Sprintf("%s.*, tx.tx_id as tx_id, CONCAT(tx.tx_id, '.', %s.vout) as outpoint", outputTableName, outputTableName)).
+			Joins(fmt.Sprintf("INNER JOIN %s tx ON tx.id = %s.transaction_id", transactionTableName, outputTableName)).
+			Where(fmt.Sprintf("%s.user_id = ?", outputTableName), userID),
```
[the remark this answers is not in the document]
chris-4chain: The table names are from `gorm GEN` package - there is no risk of SQL injection

mgosek-4chain (approved):

pkg/storage/internal/actions/create.go
```diff
@@ -359,6 +367,47 @@ func (c *create) changeOutputVoutsResult(isNoSend bool, newOutputs ...*entity.Ne
...
+		}
+
+		if !output.Spendable {
+			return fmt.Errorf("failed to validate outputs: 'spendable' field value is false - output ID %d", output.ID)
+		}
+	}
+
+	return nil
+}
+
```
ac4ch: there is a lot of if cases .. maybe strategy pattern could be used here ? 

```go
type OutputValidator func(o *wdk.Output) error

func validateNotNil(o *wdk.Output) error {
	if o == nil {
		return fmt.Errorf("output is nil")
	}
	return nil
}

func validateProvidedBy(o *wdk.Output) error {
	if o.ProvidedBy != wdk.ProvidedByStorage.String() {
		return fmt.Errorf("invalid providedBy: got %s, want %s - output ID %d", o.ProvidedBy, wdk.ProvidedByStorage.String(), o.ID)
	}
	return nil
}


func (c *create) validateNoSendChange(ctx context.Context, userID int, params CreateActionParams) error {
	outpoints := params.NoSendChange

	outputs, err := c.outputRepo.FindOutputsByOutpoints(ctx, userID, outpoints)
	if err != nil {
		return fmt.Errorf("failed to find outputs by outpoints: %w", err)
	}

	if len(outpoints) != len(outputs) {
		return fmt.Errorf("output count mismatch: got %d, want %d", len(outputs), len(outpoints))
	}

	validators := []OutputValidator{
		validateNotNil,
		validateProvidedBy,
	       // ... add more cases 
	}

	for _, output := range outputs {
		for _, validate := range validators {
			if err := validate(output); err != nil {
				return err
			}
		}
	}

	return nil
}
```

ac4ch (approved):

pkg/storage/internal/actions/create.go
chris-4chain: Check it out 👀 :)

pkg/wdk/provided_by.go
```diff
@@ -3,6 +3,9 @@ package wdk
 // ProvidedBy indicates who provided the output.
 type ProvidedBy string
 
+// String returns the underlying string value.
+func (p ProvidedBy) String() string { return string(p) }
```
dorzepowski: ### just saying

It's not used.

But I have thought, that maybe we could make a task for checking those wdk types (and maybe even sdk types) and adding convenient methods like this, just to provide better developer experience.

pkg/internal/validate/valid_create_action_args.go
```diff
@@ -49,6 +49,23 @@ func ValidCreateActionArgs(args *wdk.ValidCreateActionArgs) error {
...
+func allOutpointsUnique(outpoints []wdk.OutPoint) error {
+	seen := make(map[wdk.OutPoint]struct{})
+	for i, outpoint := range outpoints {
+		if _, exists := seen[outpoint]; exists {
+			return fmt.Errorf("duplicate outpoint at index %d: %s.%d", i, outpoint.TxID, outpoint.Vout)
+		}
+		seen[outpoint] = struct{}{}
+	}
 	return nil
 }
```
dorzepowski: ### minor

Simplification, but at costs of details in error

```suggestion
	if args.IsNoSend && len(args.Options.NoSendChange) > 0 && !is.UniqueSlice(args.Options.NoSendChange) {
		return fmt.Errorf("duplicated outpoints in Options.NoSendChange")
	}

	return nil
}
```

pkg/internal/validate/validate_no_send_change_outputs_test.go
```diff
@@ -0,0 +1,106 @@
...
+				Change:     false,
+				ProvidedBy: "storage",
+				Purpose:    "default",
+				BasketName: to.Ptr("bad-basket-name"),
+			}},
+			expected: "validate no send change output error:",
+		},
+		"nil output": {
+			outputs:  []*entity.Output{nil},
+			expected: "validate no send change output error:",
```
dorzepowski, line 96: As we're here, where we're checking the result error, I would suggest to define public wdk Error, ex. wdk.ErrInvalidArgument -  and check if ErrorIs.

But we can also postpone it, to some other time/task, where we will revisit whole error handling

pkg/storage/internal/actions/create.go
```diff
@@ -125,6 +128,12 @@ func (c *create) Create(ctx context.Context, userID int, params CreateActionPara
 		return nil, fmt.Errorf("basket for change (%s) not found", wdk.BasketNameForChange)
 	}
 
+	if params.IsNoSend && len(params.NoSendChange) > 0 {
```
dorzepowski, line 135: Why we don't check it at the beginning of this method, if they are invalid - then we're burning resources for: creating a reference, for finding a basket, and even for logging.

pkg/internal/testabilities/fixture_faucet.go
```diff
@@ -35,7 +35,10 @@ type faucetFixture struct {
...
 func (f *faucetFixture) TopUp(satoshis satoshi.Value, opts ...TopUpOpts) (txtestabilities.TransactionSpec, *models.UserUTXO) {
 	f.t.Helper()
-	options := TopUpOptions{}
+	options := TopUpOptions{
+		Purpose: "test-faucet-purpose",
+	}
+
 	for _, opt := range opts {
 		opt(&options)
 	}
```
dorzepowski: ### minor
```suggestion
	options := to.OptionsWithDefault(TopUpOptions{
		Purpose: "test-faucet-purpose",
	}, opts...)
```

For that to work, TopUpOpts should be an alias type not a type standalone
```go
type TopUpOpts = func(*TopUpOptions)
```

But I'm not sure if this isn't even better to use alias instead of type for opts

pkg/internal/storage/repo/outputs.go
```diff
@@ -171,6 +172,48 @@ func (o *Outputs) UnlinkOutputFromBasketByOutpoint(ctx context.Context, userID i
...
+	).Where("outpoint IN (?)", outpointStrings)
```
dorzepowski: ### optional suggestion

```suggestion
	outpointStrings := slices.Map(outpoints, func(op wdk.OutPoint) []any {
		return []any{op.TxID, op.Vout}
	})

	outputTableName := genquery.Output.TableName()
	transactionTableName := genquery.Transaction.TableName()

	query := o.db.WithContext(ctx).Table(
		"(?) as out",
		o.db.Model(&models.Output{}).
			Select(fmt.Sprintf("%s.*, tx.tx_id as tx_id", outputTableName)).
			Joins(fmt.Sprintf("INNER JOIN %s tx ON tx.id = %s.transaction_id", transactionTableName, outputTableName)).
			Where(fmt.Sprintf("%s.user_id = ?", outputTableName), userID),
	).Where("(tx_id,vout) IN (?)", outpointStrings)
```

pkg/internal/validate/validate_no_send_change_outputs_test.go
chris-4chain: It sounds like a thing for a separate task.

pkg/wdk/provided_by.go
chris-4chain: For now I'll remove this. But also, I'll add such task.

pkg/storage/internal/actions/create.go
chris-4chain: I think you misunderstand this. Especially that the this not "simple" validation - which has been migrated to `validation` package. 

In this function, there could be "expensive" DB query made. 
So, let's assume we moved the function `before` the `FindBasketByName` and then,
let's reverse your statement:
"Why we don't check the basket at the beginning, if basked doesn't exist - then we're burning resources for: creating a reference, for **expensive** checking outputs, and even for logging.

pkg/internal/storage/repo/outputs.go
chris-4chain: 🆒 
The SQL will look like this:
```sql
SELECT * FROM (
SELECT bsv_outputs.*, tx.tx_id as tx_id FROM `bsv_outputs` 
INNER JOIN bsv_transactions tx ON tx.id = bsv_outputs.transaction_id 
WHERE bsv_outputs.user_id = 1 AND `bsv_outputs`.`deleted_at` IS NULL
) as out 
WHERE (tx_id,vout) IN (("b2bbb4d473226133640be86432be628a78400f4f19c7f507777cd3ec43c34d1b",0),("34f7ebb74d69fd436d8046b72dbb6b83182d1f195ac2127442695d971e5f5564",0)) 
AND `out`.`deleted_at` IS NULL
```

pkg/storage/internal/actions/create.go
chris-4chain: Return honor, you did understand it :) 
Let's postpone this decision, because I'm 99% sure I'll change that method.

pkg/wdk/provided_by.go
chris-4chain: https://github.com/bsv-blockchain/go-wallet-toolbox/issues/413

dorzepowski (approved):

PR #410, opened 2025-07-24
feat: SPV

chris-4chain:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter


pkg/storage/internal/actions/internalize.go
```diff
@@ -38,26 +40,38 @@ func newInternalizeAction(
...
+	if ok, err := beef.Verify(ctx, in.chaintracker, false); err != nil {
+		return nil, fmt.Errorf("failed to verify beef: %w", err)
+	} else if !ok {
+		return nil, fmt.Errorf("beef verification returned false")
+	}
+
+	tx := beef.FindAtomicTransactionByHash(txIDHash)
+	if tx == nil {
+		return nil, fmt.Errorf("problem with atomic beef: transaction with hash %s not found", txIDHash)
+	}
```
cain4chain: Some suggestion to make it slightly more readable and some ideas for error wording changes, what do you think?
```suggestion
	ok, err := beef.Verify(ctx, in.chaintracker, false)
        if err != nil {
	        return nil, fmt.Errorf("failed to verify beef: %w", err)
        }
        if !ok {
	        return nil, fmt.Errorf("beef verification failed")
        }

        tx := beef.FindAtomicTransactionByHash(txIDHash)
        if tx == nil {
	        return nil, fmt.Errorf("atomic beef error: transaction with hash %s not found")
        }
```

pkg/storage/internal/actions/create_process_inputs.go
```diff
@@ -122,7 +134,11 @@ func (proc *inputsProcessor) processInputs() (*processedInputsResult, error) {
 		return nil, fmt.Errorf("failed to get beef for inputs: %w", err)
 	}
 
-	// TODO: Make SPV
+	if ok, err := proc.beef.Verify(proc.ctx, proc.chaintracker, true); err != nil {
+		return nil, fmt.Errorf("failed to verify beef: %w", err)
+	} else if !ok {
+		return nil, fmt.Errorf("beef verification returned false")
+	}
```
mgosek-4chain:
```
ok, err := proc.beef.Verify(proc.ctx, proc.chaintracker, true)
if err != nil {
	return nil, fmt.Errorf("failed to verify beef: %w", err)
} 
if !ok {
	return nil, fmt.Errorf("beef verification returned false")
}
``` 

😸

pkg/storage/internal/actions/process.go
```diff
@@ -237,7 +237,11 @@ func (p *process) broadcastSingleTx(ctx context.Context, txID string) (*wdk.Proc
 		return nil, fmt.Errorf("failed to build valid BEEF: %w", err)
 	}
 
-	// TODO: SPV of the beef
+	if ok, err := beef.Verify(ctx, p.services, false); err != nil {
+		return nil, fmt.Errorf("failed to verify beef: %w", err)
+	} else if !ok {
+		return nil, fmt.Errorf("beef verification returned false")
+	}
```
mgosek-4chain: Same here :) but it's totally up to you if u want to apply the change :D

pkg/storage/internal/integrationtests/internalize_create_process_test.go
```diff
@@ -50,6 +50,13 @@ func TestInternalizeThenCreateThenProcess(t *testing.T) {
 			SeekPermission: nil,
 		}
 
+		// and:
+		given.Provider().BHS().OnMerkleRootVerifyResponse(
+			tsgenerated.BeefToInternalizeHeight,
+			tsgenerated.BeefToInternalizeMerkleRoot,
+			"CONFIRMED",
```
mgosek-4chain: Minor: I would suggest extracting the string literal into a const variable and placing it in the `testabilities` package for future use 😸

mgosek-4chain (changes requested):

pkg/storage/internal/actions/create_process_inputs.go
chris-4chain: It's interesting that, from my point of view, in this case the `if-scoped-variables` perfectly suit. 
But you (and @cain4chain) suggest me to unwrap it. 

Let's discuss this. Advantages of my implementation is:
- the `err` and `ok` are scope-limited to the considered part
    - especially `ok` is good to limit, because it loses context apart from the `Verify` call 
- it is a logical and "visual" block - easy to understand as a single thing

pkg/storage/internal/actions/internalize.go
chris-4chain: I rephrased the errors - including your suggestions.
But, regarding ifs and `err/ok` variables, see below comment 👇

pkg/storage/internal/actions/process.go
chris-4chain: ☝️

mgosek-4chain (approved):

pkg/storage/internal/actions/create_process_inputs.go
cain4chain: ok makes sense and sounds good to me :)

cain4chain (approved):

dorzepowski (approved):

PR #411, opened 2025-07-24
[FEATURE] WalletServices: create getScriptHashHistory method missing in providers

### LINKED ISSUE #383: [FEATURE] WalletServices: create getScriptHashHistory method missing in providers
## Desired Solution

## Suggested Implementation

## Alternatives Considered

## Additional Context

## Categorize Feature

ac4ch:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter


pkg/services/internal/bitails/helpers.go
```diff
@@ -217,3 +218,32 @@ func (b *Bitails) storeRootInCache(height uint32, root *chainhash.Hash) {
...
+
+	resp, ok := res.Result().(*dto.ScriptHistoryResponse)
+	if !ok {
+		return nil, "", fmt.Errorf("failed to parse script history response: %T", res.Result())
+	}
+	if resp.Error != "" {
+		return nil, "", fmt.Errorf("bitails API error: %s", resp.Error)
+	}
+	return resp.History, resp.PgKey, nil
+}
```
mgosek-4chain: Maybe we can simplify this a bit using the following code snippet:

```
func (b *Bitails) fetchScriptHistory(ctx context.Context, scriptHash, pgKey string, limit int) ([]dto.ScriptHistoryItem, string, error) {
	url, err := scriptHashHistoryURL(b.url, scriptHash)
	if err != nil {
		return nil, "", fmt.Errorf("build URL for script history: %w", err)
	}

	var dst dto.ScriptHistoryResponse
	req := b.httpClient.R().SetContext(ctx).
		SetQueryParam("pgkey", pgKey).
		SetQueryParam("limit", fmt.Sprintf("%d", limit)).
		SetResult(&dst)

	res, err := req.Get(url)
	if err != nil {
		return nil, "", fmt.Errorf("failed to get script history: %w", err)
	}
	if res.IsError() {
		return nil, "", fmt.Errorf("bitails API error: %w", err)
	}

	if !res.IsSuccess() {
		return nil, "", fmt.Errorf("unexpected status code %d fetching script history", res.StatusCode())
	}

	return dst.History, dst.PgKey, nil
}

```

pkg/services/internal/bitails/script_history_test.go
```diff
@@ -0,0 +1,204 @@
...
+import (
+	"fmt"
+	"testing"
+
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/services/internal/bitails/testabilities"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestBitails_GetScriptHistory_ConfirmedAndUnconfirmed(t *testing.T) {
```
mgosek-4chain: `TestBitails_GetScriptHistory_WithTransactionsOneConfirmedOneUnconfirmed`  🤔 However, this is just a suggestion to make the unit test name more descriptive. It's up to you and rest of the team to decide whether you want to apply this change.

pkg/services/internal/bitails/service.go
```diff
@@ -241,3 +241,69 @@ func (b *Bitails) RawTx(ctx context.Context, txID string) (*wdk.RawTxResult, err
...
+
+	return &wdk.ScriptHistoryResult{
+		Name:       ServiceName,
+		ScriptHash: scriptHash,
+		History:    items,
+	}, nil
+}
+
+// as i am not sure which one would be used .. as better to have
+// GetScriptHashHistory1 fetches paginated script hash history. Without limit to pull full history.
```
mgosek-4chain: Mind taking a quick look at this code snippet? 😸 @dorzepowski @chris-4chain 
I'm not sure either 😄

mgosek-4chain (changes requested):

pkg/services/internal/bitails/service.go
chris-4chain: @ac4ch Could you describe your point of view? What options do we have? What decision we need to make? And what are advantages of each option.

ac4ch: yes i thnk best is the 1st version of it ... as simulate same behaving as WoC ... this 5k is quite a lot :)

ac4ch: `GetScriptHashHistory` with page limit (100) - simulates behavior of WoC, fetches only the first page (default limit), fast and light... same as in WoC

`GetScriptHashHistoryNotSureIfWeWannaAllResults` - paginated pull with higher limit (5000), pulls the entire history by iterating over nextPageToken...

so we would need to decide wheter we want ... 
- fast .. consisten with WoC behavior .. (current GetSciptHashHistory implementation) 
- full data coverage regardless of limits 

i did suggest option which it simulates WoC behavior 

WoC doc:
image
image

Bitails doc:
image

pkg/services/internal/bitails/helpers.go
ac4ch: ok i took your example but changed it too 
```go
func (b *Bitails) fetchScriptHistory(ctx context.Context, scriptHash, pgKey string, limit int) ([]dto.ScriptHistoryItem, string, error) {
	url, err := scriptHashHistoryURL(b.url, scriptHash)
	if err != nil {
		return nil, "", fmt.Errorf("build URL for script history: %w", err)
	}

	var dst dto.ScriptHistoryResponse
	req := b.httpClient.R().SetContext(ctx).
		SetQueryParam("pgkey", pgKey).
		SetQueryParam("limit", fmt.Sprintf("%d", limit)).
		SetResult(&dst)

	res, err := req.Get(url)
	if err != nil {
		return nil, "", fmt.Errorf("failed to get script history: %w", err)
	}

		return nil, "", fmt.Errorf("unexpected status code %d fetching script history", res.StatusCode())
	}
	if dst.Error != "" {
		return nil, "", fmt.Errorf("error in script history response: %s", dst.Error)
	}

	return dst.History, dst.PgKey, nil
}
```

so is satisfies tests .. properly catches api error strings .. avoids misuse of `err` when `res.IsError` is true but `err == nil`

pkg/services/internal/bitails/service.go
chris-4chain: I would go for the simplest and fastest version to be consistent with what we already have for WoC.

jakubmkowalski (approved):

pkg/services/internal/bitails/utils.go
```diff
@@ -59,6 +60,20 @@ func ConvertHeader(raw []byte, height uint32) (*wdk.ChainBlockHeader, error) {
 	}, nil
 }
 
+func validateScriptHash(scriptHash string) error {
+	if scriptHash == "" {
+		return fmt.Errorf("scripthash cannot be empty")
+	}
+	if len(scriptHash)%2 != 0 {
+		return fmt.Errorf("invalid scripthash length: must be even")
+	}
```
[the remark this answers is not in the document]
ac4ch:
```go
case len(scriptHash) != 64:
		return fmt.Errorf("invalid scripthash length: must be 64 characters for a SHA256 hash")
```
check was added....

pkg/services/internal/bitails/script_history_test.go
```diff
@@ -0,0 +1,204 @@
...
+
+func TestBitails_GetScriptHistory_LargeHistory(t *testing.T) {
+	testCases := []struct {
+		name               string
+		confirmedCount     int
+		unconfirmedCount   int
+		startHeight        int
+		expectedTotalCount int
+	}{
+		{
```
chris-4chain, line 140: The test `TestBitails_GetScriptHistory_LargeHistory` is a bit misleading. It's intended to test a large history, but the mock is configured to return all transactions in a single response, regardless of the count. This means the test doesn't actually cover the pagination logic that would be required in a real-world scenario where the number of transactions exceeds the page limit.

I mean, the mock doesn't act as the real server which returns rows up to provided `limit`. 
- The mock ignores this limit - as far as I understand this.

pkg/services/internal/bitails/service.go
```diff
@@ -241,3 +241,29 @@ func (b *Bitails) RawTx(ctx context.Context, txID string) (*wdk.RawTxResult, err
...
 	}, nil
 }
+
+// GetScriptHashHistory fetches the script hash history for a given script hash.
+func (b *Bitails) GetScriptHashHistory(ctx context.Context, scriptHash string) (*wdk.ScriptHistoryResult, error) {
+	if err := validateScriptHash(scriptHash); err != nil {
+		return nil, fmt.Errorf("invalid script hash %s: %w", scriptHash, err)
+	}
+
+	history, _, err := b.fetchScriptHistory(ctx, scriptHash, "", defs.DefaultPageLimit)
```
chris-4chain: 1. I'm suggesting `scriptHistory` to not overlap the `"history"` name - which is already in this scope as package name. 
2. pgKey variable could be more descriptive - consider `paginationKey`. But anyway, as far as I read this correctly, there is no case where this argument is not-empty. So, consider removing this arg.

pkg/defs/services.go
```diff
@@ -23,6 +23,9 @@ const (
 	// DefaultRootForHeightRetries is the number of retries for fetching the root for height validation
 	DefaultRootForHeightRetries = 3
 
+	// DefaultPageLimit is the default limit for paginated requests for service get scripts history (100 is the maximum value for WoC)
+	DefaultPageLimit = 100
```
chris-4chain: Defining `DefaultPageLimit` on this level (quite global) can lead to misunderstanding and misusage of this constant. 

For your current purpose, it seems to me, you wanted to define a page size for bitails "how many rows to fetch from the 3rd party service".
- I would keep such a constant closer to the implementation and made it "private"
- Additionally, I would consider making the pageLimit configurable (but this is optional)

pkg/services/internal/bitails/utils.go
```diff
@@ -59,6 +60,26 @@ func ConvertHeader(raw []byte, height uint32) (*wdk.ChainBlockHeader, error) {
...
+		return fmt.Errorf("scripthash cannot be empty")
+	case len(scriptHash)%2 != 0:
+		return fmt.Errorf("invalid scripthash length: must be even")
+	case len(scriptHash) != 64:
+		return fmt.Errorf("invalid scripthash length: must be 64 characters for a SHA256 hash")
+	case !isHex(scriptHash):
+		return fmt.Errorf("invalid scripthash format: must be hex encoded")
+	default:
+		return nil
+	}
```
chris-4chain: The general-purpose HexString validation is already implemenetd in the "primitives" package. Consider creating specialized type in the primitives or use following implementation. 

```suggestion
	if scriptHash == "" {
		return fmt.Errorf("scripthash cannot be empty")
	}

	if len(scriptHash) != 64 {
		return fmt.Errorf("invalid scripthash length: must be 64 characters for a SHA256 hash")
	}

	if err := primitives.HexString(scriptHash).Validate(); err != nil {
		return fmt.Errorf("invalid scripthash format: %w", err)
	}

	return nil
```

pkg/internal/testabilities/testservices/fixture_bitails.go
```diff
@@ -314,3 +315,138 @@ func (b *bitailsFixture) WillReturnRawTxHttpError(txid string, status int) {
...
+		for i := 0; i < b.confirmedCount; i++ {
+			height := b.startHeight + i
+			response.History = append(response.History, ScriptHistoryItem{
+				TxID:   fmt.Sprintf("c%010de1b81dd2c9c0c6cd67f9bdf832e9c2bb12a1d57f30cb6ebbe78d9", i),
+				Height: &height,
+			})
+		}
+		for i := 0; i < b.unconfirmedCount; i++ {
+			response.History = append(response.History, ScriptHistoryItem{
+				TxID:   fmt.Sprintf("u%010de1b81dd2c9c0c6cd67f9bdf832e9c2bb12a1d57f30cb6ebbe78d9", i),
```
chris-4chain: `u` character is not a part of hex string.

pkg/services/internal/bitails/helpers.go
```diff
@@ -217,3 +218,31 @@ func (b *Bitails) storeRootInCache(height uint32, root *chainhash.Hash) {
...
+	url, err := scriptHashHistoryURL(b.url, scriptHash)
+	if err != nil {
+		return nil, "", fmt.Errorf("build URL for script history: %w", err)
+	}
+
+	var dst dto.ScriptHistoryResponse
+	req := b.httpClient.R().SetContext(ctx)
+	if pgKey != "" {
+		req = req.SetQueryParam("pgkey", pgKey)
+	}
```
chris-4chain: For current usage - this `pgKey != ""` never happens. 
Maybe we could remove this and (maybe) we should put here a `NOTE` comment what are our assumptions and what limitations. 

Something like: "NOTE: This fetches only the last history records - based on the `limit` argument. Pagination is not supported".

pkg/services/internal/bitails/helpers.go
```diff
@@ -217,3 +218,31 @@ func (b *Bitails) storeRootInCache(height uint32, root *chainhash.Hash) {
...
+	req = req.SetQueryParam("limit", fmt.Sprintf("%d", limit)).SetResult(&dst)
```
chris-4chain: 1. Consider `fmt.Sprint(limit)` - it would produce the same result with less code. 

2. Also, I would focus on the "SetQuery" thing at this line. I would move the `SetResult` above: `req := b.httpClient.R().SetContext(ctx).SetResult(&dst)`

pkg/services/internal/bitails/internal/dto/script_hash_history.go
```diff
@@ -0,0 +1,18 @@
+package dto
+
+// ScriptHistoryResponse represents the response from the script history endpoints
+type ScriptHistoryResponse struct {
+	ScriptHash string              `json:"scripthash,omitempty"`
```
chris-4chain: This field is not used.

pkg/services/internal/bitails/service.go
```diff
@@ -241,3 +241,29 @@ func (b *Bitails) RawTx(ctx context.Context, txID string) (*wdk.RawTxResult, err
...
+		return nil, fmt.Errorf("failed to fetch script hash history: %w", err)
+	}
+
+	items := make([]wdk.ScriptHistoryItem, 0, len(history))
+	for _, item := range history {
+		items = append(items, wdk.ScriptHistoryItem{
+			TxHash: item.TxID,
+			Height: item.Height,
+		})
+	}
```
chris-4chain:
```suggestion
	items := slices.Map(history, func(item dto.ScriptHistoryItem) wdk.ScriptHistoryItem {
		return wdk.ScriptHistoryItem{
			TxHash: item.TxID,
			Height: item.Height,
		}
	})
```

pkg/services/internal/bitails/internal/dto/script_hash_history.go
nonus25: yes but this is what we getting back from api 
```json
[
  {
    "scripthash": "c79e8d823c1ce9b80c9c340a389409f489989800044466c9d05bfef12c472232",
    "histories": [
    {
    "txid": "40ff70b97334699d71a552ab5ea97c996fa8d64a9e0dc0041ec1701b10ed6914",
    "inputSatoshis": 2000,
    "outputSatoshis": 0,
    "time": 1655049588,
    "blockheight": 815019,
    }
  ]
  }
]
```

pkg/services/internal/bitails/script_history_test.go
ac4ch: ok i removed pkKey fully to not confuse you per pagination .. and this test is more testing limits not pagination so i renamed it to `TestBitails_GetScriptHistory_ManyItems_NoPagination`

pkg/defs/bitails.go
```diff
@@ -1,12 +1,26 @@
...
+	scriptHashHistoryPageLimitMaximum = 5000
+	defaultScriptHashHistoryPageLimit = 100
+)
+
 // Bitails configures the Bitails service
 type Bitails struct {
 	// APIKey is the authentication key used to communicate with Bitails (if applicable)
-	APIKey string `mapstructure:"api_key"`
+	APIKey                     string `mapstructure:"api_key"`
+	ScriptHashHistoryPageLimit int    `mapstructure:"script_hash_history_page_limit"`
```
chris-4chain, line 14: 👍

pkg/services/internal/bitails/helpers.go
```diff
@@ -217,3 +218,28 @@ func (b *Bitails) storeRootInCache(height uint32, root *chainhash.Hash) {
...
 }
+
+func (b *Bitails) fetchScriptHistory(ctx context.Context, scriptHash string) ([]dto.ScriptHistoryItem, error) {
+	url, err := scriptHashHistoryURL(b.url, scriptHash)
+	if err != nil {
+		return nil, fmt.Errorf("build URL for script history: %w", err)
+	}
+
+	var dst dto.ScriptHistoryResponse
+	// NOTE: Although pagination is not supported, we still send a "limit" to match WoC behavior (max 100 items returned).
```
chris-4chain: SInce the limit is configurable, I think you should not mention the number "100"

pkg/services/internal/bitails/helpers.go
```diff
@@ -217,3 +218,28 @@ func (b *Bitails) storeRootInCache(height uint32, root *chainhash.Hash) {
...
+	req := b.httpClient.R().SetContext(ctx).SetResult(&dst).SetQueryParam("limit", fmt.Sprint(b.hashScriptHistoryPageLimit)).SetResult(&dst)
```
chris-4chain: Minor, just saying.
For me, it's more readable to have this multiline:
```golang
	req := b.httpClient.R().
		SetContext(ctx).
		SetResult(&dst).
		SetQueryParam("limit", fmt.Sprint(b.hashScriptHistoryPageLimit)).
		SetResult(&dst)
```

pkg/services/internal/bitails/internal/dto/script_hash_history.go
chris-4chain: And since we don't need this, we unnecessarily parse & store this field.

examples/internal/show/show.go
```diff
@@ -182,3 +182,29 @@ func RawTxOutput(result *wdk.RawTxResult) {
...
+
+	headers := []string{"TxHash", "Status", "Block Height"}
+	var rows [][]string
+
+	for _, item := range result.History {
+		status := "Unconfirmed"
+		height := "-"
+		if item.Height != nil && *item.Height > 0 {
+			status = "Confirmed"
+			height = fmt.Sprintf("%d", *item.Height)
```
chris-4chain:
```suggestion
			height = fmt.Sprint(*item.Height)
```

pkg/internal/testabilities/testservices/fixture_bitails.go
```diff
@@ -314,3 +315,139 @@ func (b *bitailsFixture) WillReturnRawTxHttpError(txid string, status int) {
...
+	} else {
+		for i := 0; i < b.confirmedCount; i++ {
+			height := b.startHeight + i
+			response.History = append(response.History, ScriptHistoryItem{
+				TxID:   fmt.Sprintf("%02x%062s", i, "e1b81dd2c9c0c6cd67f9bdf832e9c2bb12a1d57f30cb6ebbe78d9"),
+				Height: &height,
+			})
+		}
+		for i := 0; i < b.unconfirmedCount; i++ {
+			txid := fmt.Sprintf("%02x%062s", i, "e1b81dd2c9c0c6cd67f9bdf832e9c2bb12a1d57f30cb6ebbe78d9")
```
chris-4chain, line 435: Is it a problem that the same txIDs can be returned twice?
- for confirmed
- and for unconfirmed

ac4ch: no why that why before i had `u` and `c` there ...

chris-4chain: and `u` is not a part of hex characters. So instead of `u` you could have put there `'a', 'b', 'd', 'e', 'f' or 0...9`

ac4ch: changed to `e1b71dd2c9c0c6cd67f9bdf832e9c2bb12a1d57f30cb6ebbe78d9`

ac4ch: rigth since this was simple mock i could put there anything .. and u was for unconfirmed and c was for confirmed .. therefor you could easily distinguish to which test this tx belong .. anyway .. now is different

chris-4chain (approved):

mgosek-4chain (approved):

PR #419, opened 2025-07-25
feat(#166): add get block header for height to BHS. 

### LINKED ISSUE #166: [FEATURE] WalletServices: create getHeaderForHeight method that will use BHS

mgosek-4chain:
## Linked Issues / Tickets

Reference any related issues or tickets, e.g. "Closes #166 ".

## Testing Procedure

- [X] All tests pass locally

## Checklist:

- [X] I have run the linter


pkg/services/internal/bhs/internal/dto/dto.go
```diff
@@ -7,12 +7,36 @@ import (
 	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/wdk"
 )
 
+type BlockHeaderHeight struct {
```
chris-4chain: The name `BlockHeaderHeight` is misleading - one could think it is a struct to hold only the Height numeric value of a block.

pkg/wdk/services_chain_block_headers.go
```diff
@@ -46,3 +53,64 @@ type ChainBlockHeader struct {
...
+		writeLittleEndianOrder(&buff, c.Time),
+		writeLittleEndianOrder(&buff, c.Bits),
+		writeLittleEndianOrder(&buff, c.Nonce),
+	}
+
+	for _, err := range errs {
+		if err != nil {
+			return nil, err
+		}
+	}
```
chris-4chain: I know that this way it looks slightly more readable - but in case of error, the flow doesn't stop until all the parts are converted. 
This way we burn CPU for nothing. 

It would make sense if we joined the errors into one - but I'm not convinced it's necessary.

So consider "standard approach", like:
```golang
	var buff bytes.Buffer
	if err := writeLittleEndianOrder(&buff, c.Version); err != nil {
		return nil, err //of course with proper error wrapping :)
	}
	if err := writeReversedBytes(&buff, hash); err != nil {
		return nil, err
	}
	if err := writeReversedBytes(&buff, root); err != nil {
		return nil, err
	}
	if err := writeLittleEndianOrder(&buff, c.Time); err != nil {
		return nil, err
	}
	if err := writeLittleEndianOrder(&buff, c.Bits); err != nil {
		return nil, err
	}
	if err := writeLittleEndianOrder(&buff, c.Nonce); err != nil {
		return nil, err
	}
```

Or - slightly more alike to your code:
```golang
var buff bytes.Buffer
	components := []func() error {
		func() error { return writeLittleEndianOrder(&buff, c.Version) },
		func() error { return writeReversedBytes(&buff, hash) },
		func() error { return writeReversedBytes(&buff, root) },
		func() error { return writeLittleEndianOrder(&buff, c.Time) },
		func() error { return writeLittleEndianOrder(&buff, c.Bits) },
		func() error { return writeLittleEndianOrder(&buff, c.Nonce) },
	}
	
	for _, writeFunc := range components {
		if err := writeFunc(); err != nil {
			return nil, fmt.Errorf("failed to write block header component: %w", err)
		}
	}
```

pkg/wdk/services_chain_block_headers.go
```diff
@@ -46,3 +53,64 @@ type ChainBlockHeader struct {
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to convert 'previous hash' field into bytes slice: %w", err)
+	}
+
+	root, err := hex.DecodeString(c.MerkleRoot)
+	if err != nil {
+		return nil, fmt.Errorf("failed to convert 'merkle root' field into bytes slice: %w", err)
+	}
+
+	var buff bytes.Buffer
```
chris-4chain: Consider preallocating the buffer - it's trivial in this case, because it has well-known, constant length: `80`.

```suggestion
	buff := bytes.NewBuffer(make([]byte, 0, 80)) 
```

pkg/wdk/services_chain_block_headers.go
```diff
@@ -46,3 +53,64 @@ type ChainBlockHeader struct {
...
+	}
+	return hex.EncodeToString(bb), nil
+}
+
+// Bytes returns the serialized byte representation of the block header.
+// It includes the reversed previous block hash and Merkle root,
+// followed by time, bits, and nonce fields written in little-endian order.
+// Returns an error if any of the fields cannot be parsed or written.
+func (c *ChainBaseBlockHeader) Bytes() ([]byte, error) {
+	hash, err := hex.DecodeString(c.PreviousHash)
```
chris-4chain, line 73: I'm not sure if we should validate `c` field here. 
If so, we need to validate the length of the input strings - they both should be "32-byte hex"

pkg/services/internal/bhs/service.go
```diff
@@ -54,6 +54,38 @@ func New(httpClient *resty.Client, logger *slog.Logger, network defs.BSVNetwork,
...
+	url, err := headerByHeight(b.cfg.URL)
+	if err != nil {
+		return nil, fmt.Errorf("error building URL: %w", err)
+	}
+
+	var blocks []dto.BlockHeaderHeight
+	res, err := b.httpClient.R().
+		SetContext(ctx).
+		SetResult(&blocks).
+		SetQueryParam("height", fmt.Sprintf("%d", height)).
```
chris-4chain: I think `fmt.Sprint(height)` would do the same job.

pkg/services/internal/bhs/service.go
```diff
@@ -54,6 +54,38 @@ func New(httpClient *resty.Client, logger *slog.Logger, network defs.BSVNetwork,
...
+		SetQueryParam("count", "1").
+		Get(url)
+	if err != nil {
+		return nil, fmt.Errorf("unexpected response from API (URL: %s, code: %d): %w", res.Request.URL, res.StatusCode(), err)
```
chris-4chain: `res` variable can be nil at this point. 
Consider the situation when `resty` returns `return nil, err`.

pkg/internal/testabilities/testservices/fixture_bhs.go
```diff
@@ -13,16 +13,15 @@ import (
...
-	PreviousBlock string `json:"prevBlockHash"`
-}
+type endpoint string
+
+func (e endpoint) build() string { return defs.BHSTestURL + string(e) }
+
+const (
+	tipLongestPath       endpoint = "/api/v1/chain/tip/longest"
+	verifyMerkleRootPath endpoint = "/api/v1/chain/merkleroot/verify"
+	headerByHeightPath   endpoint = "/api/v1/chain/header/byHeight"
```
chris-4chain: Is the custom type with the `build` method necessary?
Just:
```golang
const (
	tipLongestPath        = defs.BHSTestURL + "/api/v1/chain/tip/longest"
	verifyMerkleRootPath  = defs.BHSTestURL + "/api/v1/chain/merkleroot/verify"
	headerByHeightPath    = defs.BHSTestURL + "/api/v1/chain/header/byHeight"
)
```

But maybe, it's beneficial to "don't repeat yourself" with the `defs.BHSTestURL` 🤷 - what do you think?

pkg/wdk/services_chain_block_headers_test.go
```diff
@@ -0,0 +1,108 @@
...
+			actualBytes, err := tc.block.Bytes()
+
+			// then:
+			assert.NotNil(t, err)
+			assert.Nil(t, actualBytes)
+		})
+	}
+}
+
+func TestChainBaseBlockHeader_Bytes_PositivePaths(t *testing.T) {
```
chris-4chain, line 11: I would prefer to put "PositivePaths" before the "negative" ones :)

pkg/wdk/services_chain_block_headers.go
```diff
@@ -46,3 +53,64 @@ type ChainBlockHeader struct {
...
+		return "", fmt.Errorf("failed to marshal chain block header height: %w", err)
+	}
+	return hex.EncodeToString(bb), nil
+}
+
+// Bytes returns the serialized byte representation of the block header.
+// It includes the reversed previous block hash and Merkle root,
+// followed by time, bits, and nonce fields written in little-endian order.
+// Returns an error if any of the fields cannot be parsed or written.
+func (c *ChainBaseBlockHeader) Bytes() ([]byte, error) {
```
chris-4chain, line 72: I'm wondering if this shouldn't return `[80]byte` as the block header "fingerprint" has constant length.

pkg/wdk/services_chain_block_headers_test.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
+			// when:
+			actualBytes, err := tc.block.Bytes()
+
+			// then:
+			assert.NoError(t, err)
+			assert.Equal(t, tc.expectedBytes, actualBytes)
```
chris-4chain, line 80: Maybe it would be beneficial to add the check for the "expected block hash" as it's well known and easy to validate in, for example WoC.

NOTE: Expected hash can be got from WoC for a given height. 

```golang
func matchesDoubleHash(t testing.TB, fingerprint []byte, expectedHash string) {
	first := sha256.Sum256(fingerprint)
	doubleHashed := sha256.Sum256(first[:])

	blockHash, err := chainhash.NewHash(doubleHashed[:])
	require.NoError(t, err, "Failed to create chainhash from double-hashed bytes")

	assert.Equal(t, expectedHash, blockHash.String(), "Double hash does not match expected value")
}

```

pkg/wdk/services_chain_block_headers.go
```diff
@@ -46,3 +53,76 @@ type ChainBlockHeader struct {
...
+		return nil, fmt.Errorf("failed to write the 'version' field bytes in little-endian order: %w", err)
+	}
+	if err := writeReversedBytes(buff, hash); err != nil {
+		return nil, fmt.Errorf("failed to write the 'merkle root' field bytes in little-endian order: %w", err)
+	}
+	if err := writeReversedBytes(buff, root); err != nil {
+		return nil, fmt.Errorf("failed to write the 'merkle root' field bytes in little-endian order: %w", err)
+	}
+	if err := writeLittleEndianOrder(buff, c.Time); err != nil {
+		return nil, fmt.Errorf("failed to write the 'merkle root' field bytes in little-endian order: %w", err)
```
chris-4chain: The text: "failed to write the 'merkle root' field bytes in little-endian order" seems copypasted. Please create different error messages for different cases.

pkg/wdk/services_chain_block_headers.go
chris-4chain: You didn't reply me to this comment.
What do you think about my suggestion?

pkg/wdk/services_chain_block_headers_test.go
chris-4chain: 💯

pkg/internal/testabilities/testservices/fixture_bhs.go
```diff
@@ -46,123 +43,131 @@ type BHSFixture interface {
...
-	transport       *httpmock.MockTransport
-	longestChainTip *longestChainTipResponse
+	transport                  *httpmock.MockTransport
+	longestChainTip            *longestChainTipResponse
+	headerByHeightResponse     []headerByHeightResponse
+	bhsAnyEndpointRegexFixture *regexp.Regexp
+}
+
+func (b *bhsFixture) DefaultHeaderByHeightResponse() []headerByHeightResponse {
+	b.Helper()
```
ac4ch, line 53: if we choose now to go like this .. WDYT about updating all of them at lease in this file ? to remove `TB` from `b.Helper`

pkg/services/internal/bhs/internal/dto/dto.go
```diff
@@ -7,12 +7,36 @@ import (
...
+	}
+}
+
 type TipResponse struct {
 	Hash             string   `json:"hash"`
 	Version          uint32   `json:"version"`
 	PreviousBlock    string   `json:"prevBlockHash"`
 	MerkleRoot       string   `json:"merkleRoot"`
-	Timestamp        uint64   `json:"creationTimestamp"`
+	Timestamp        uint32   `json:"creationTimestamp"`
```
ac4ch, line 39: i understand that this timestamp would only store seconds .. as if milliseconds then uint32 would overflows quite fast ?

pkg/services/internal/bhs/service.go
```diff
@@ -54,6 +54,39 @@ func New(httpClient *resty.Client, logger *slog.Logger, network defs.BSVNetwork,
...
+		SetContext(ctx).
+		SetResult(&blocks).
+		SetQueryParam("height", fmt.Sprint(height)).
+		SetQueryParam("count", "1")
+
+	res, err := req.Get(url)
+	if err != nil {
+		return nil, fmt.Errorf("unexpected response from API (URL: %s): %w", req.URL, err)
+	}
+	if !res.IsSuccess() {
```
ac4ch: since this only checks success like this i think you also should use IsError method to check statuses over 400... like 503 :) 

> IsSuccess method returns true if HTTP status `code >= 200 and <= 299` otherwise false.
> IsError method returns true if HTTP status `code >= 400` otherwise false.

not sure how u wanna handle redirections 302 .. or if there is need to handle this ...

pkg/wdk/services_chain_block_headers.go
```diff
@@ -46,3 +53,76 @@ type ChainBlockHeader struct {
...
+		return "", fmt.Errorf("failed to marshal chain block header height: %w", err)
+	}
+	return hex.EncodeToString(bb), nil
+}
+
+// Bytes returns the serialized byte representation of the block header.
+// It includes the reversed previous block hash and Merkle root,
+// followed by time, bits, and nonce fields written in little-endian order.
+// Returns an error if any of the fields cannot be parsed or written.
+func (c *ChainBaseBlockHeader) Bytes() ([]byte, error) {
```
ac4ch, line 72: this method has so many if statements that i am wondering if it is not better to use here strategy pattern ?

pkg/wdk/services_chain_block_headers.go
mgosek-4chain: I would like to hear @dorzepowski  opinion about this optimization :D

pkg/services/internal/bhs/internal/dto/dto.go
mgosek-4chain: Good catch, thanks.

EDIT: After discuss with @chris-4chain, we agreed that we can leave it as `uint32`.

pkg/wdk/services_chain_block_headers.go
mgosek-4chain: What do you mean by the strategy pattern?  By definition, the strategy patten alters runtime behavior - we don't have such a scenario here.

pkg/internal/testabilities/testservices/fixture_bhs.go
mgosek-4chain: @chris-4chain @dorzepowski

pkg/wdk/services_chain_block_headers.go
ac4ch: https://www.ottorinobruni.com/simplify-your-csharp-code-replace-if-else-with-the-strategy-pattern-in-dotnet/#:~:text=The%20Strategy%20Pattern%20is%20particularly,more%20adaptable%20to%20future%20changes.

or look here 
https://github.com/bsv-blockchain/go-wallet-toolbox/blob/c70f47215057a4ee925c9ed3f7c7ede0adef1e04/pkg/internal/validate/validate_no_send_change_outputs.go

mgosek-4chain: @ac4ch  this is not the strategy pattern.

The Strategy pattern can be applied in scenarios where you need to change behavior at runtime. In my codebase, we have a static workflow built around `if` statements. 

Let's' say you're designing a parking lot application and want to apply different policies for parking spot reservations ( reservation is a behavior) based on customer types. For example, for regular customers you might define a policy (strategy) that allows reservations only within a specific time range. 

However, for premium customers, you might use a completely different strategy that allows them to reserve parking spots without such restrictions. 

Instead of writing large blocks of `if` statements like:

```
if customer.Type == "PREMIUM" {
   // ...
} else if customer.Type == "REGULAR" {
  // ...
}
``` 

You can simplify the behavior by defining a common interface and implementing specific behaviors (policies) for each customer type. This behavior is then selected at runtime, typically in a domain services based on the customer's role or type.

mgosek-4chain: @chris-4chain @dorzepowski please take a look at @ac4ch  suggested approach. Maybe it's a good idea to go in this direction to keep the codebase more consistent?

ac4ch: sure u know better .. https://rednafi.com/go/strategy_pattern/ 
not sure why all of them use this name ... or i do use it must be mine misunderstanding ...
also this conversation was in slack before and i think @chris-4chain already implemented something like this ...
and i provided you the example  
image

ac4ch: here: https://github.com/bsv-blockchain/go-wallet-toolbox/blob/c70f47215057a4ee925c9ed3f7c7ede0adef1e04/pkg/internal/validate/validate_no_send_change_outputs.go

pkg/wdk/services_chain_block_headers.go
dorzepowski: I would postpone this change,
the array is less convenient in use then slices,
most of the common functions like: hex.EncodeToString or hashing are accepting the slice, and compiler will complain when you would pass and array, so you would need always to do `bytes[:]` 

What's more, honestly, I don't see any reason for getting those bytes except to calculate the hash from them.

pkg/wdk/services_chain_block_headers.go
mgosek-4chain: > sure u know better .. https://rednafi.com/go/strategy_pattern/
> not sure why all of them use this name ... or i do use it must be mine misunderstanding ...

It might be the latter. The article shows that the strategy in question executes specific logic. Typically, the type of strategy is selected at runtime and strategy means 'behavior'. 


> Then, we initialize these strategy classes at runtime and explicitly pass the instances to the display function. The function knows how to use the strategy instances and executes a specific strategy to print a message in a particular format based on a certain condition.

We don’t initialize any policies (or strategies) at runtime - we currently have a single flow that includes validation and returns the serialized byte representation of the block header. This is our default behavior (or strategy/policy, if you will).

If, in the future, we need to support an alternative serialization method based on query parameters or other conditions, then it would make sense to introduce a strategy selector type. That would allow us to return a specific serialization (output bytes) based on the input.

@chris-4chain implementation is a validator type that encapsulates the entire logic; it's not a strategy - we're not altering any business logic behavior here. That said, we can use the same approach - no problem. I'd just like to wait for Damian's (@dorzepowski ) opinion.

dorzepowski: IMHO current solution looks good enough.

Introducing some additional patterns here smells like overengineering.
This is sad but true: all those ifs are just a standard way of error passing that cannot be skipped, because some language designers decide that there won't be exceptions to throw.

chris-4chain (approved):

pkg/services/internal/bhs/internal/dto/dto.go
chris-4chain: We need to stick to the 32 bit variable as it is aligned with our blockchain.

ac4ch (approved):

PR #422, opened 2025-07-26
feat: processAction: 'sendWith' and 'noSend' 

chris-4chain:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter


pkg/storage/internal/actions/process.go
```diff
@@ -45,22 +45,39 @@ func newProcessAction(logger *slog.Logger, txRepo TransactionsRepo, commissionCf
...
 	}
 
-	if args.IsDelayed {
-		panic("not implemented yet")
+	return p.broadcastTxs(ctx, p.txIDsToBroadcast(args))
+}
+
+func (p *process) txIDsToBroadcast(args *wdk.ProcessActionArgs) []string {
+	count := len(args.SendWith)
+	if args.TxID != nil {
```
dorzepowski, line 68: AFAIR this should be included only if NoSend is false

pkg/storage/internal/actions/process.go
```diff
@@ -45,22 +45,39 @@ func newProcessAction(logger *slog.Logger, txRepo TransactionsRepo, commissionCf
...
+		count++
 	}
 
-	return p.broadcastSingleTx(ctx, string(*args.TxID))
+	result := make([]string, 0, count)
+	for _, txID := range args.SendWith {
+		result = append(result, string(txID))
+	}
+	if args.TxID != nil {
```
dorzepowski, line 76: AFAIR this should be included only if NoSend is false

pkg/storage/internal/actions/process.go
```diff
@@ -215,24 +232,33 @@ func (p *process) newStatuses(args *wdk.ProcessActionArgs) (txStatus wdk.TxStatu
 	return
 }
 
-func (p *process) broadcastSingleTx(ctx context.Context, txID string) (*wdk.ProcessActionResult, error) {
-	sendStatus, err := p.getSendStatus(ctx, txID)
+func (p *process) broadcastTxs(ctx context.Context, txIDs []string) (*wdk.ProcessActionResult, error) {
+	if len(txIDs) == 0 {
```
dorzepowski: this shouldn't ever happen - we can skip this check here

pkg/internal/storage/repo/known_tx.go
```diff
@@ -135,17 +143,25 @@ func (p *KnownTx) FindKnownTxRawTxs(ctx context.Context, txIDs []string) (map[st
...
-	return model.Status, nil
+
+	return seq.Reduce(
+		seq.FromSlice(rows),
+		func(acc map[string]wdk.ProvenTxReqStatus, row *models.KnownTx) map[string]wdk.ProvenTxReqStatus {
+			acc[row.TxID] = row.Status
+			return acc
+		},
+		make(map[string]wdk.ProvenTxReqStatus, len(rows)),
+	), nil
```
dorzepowski:
```suggestion
	txIDStatuses := seq.MapTo(seq.FromSlice(rows), func(row *models.KnownTx) (string, wdk.ProvenTxReqStatus) {
		return row.TxID, row.Status
	})
	
	return seq2.CollectToMap(txIDStatuses), nil
```

pkg/internal/storage/repo/known_tx.go
```diff
@@ -261,7 +277,7 @@ func (p *KnownTx) recursiveBuildValidBEEF(ctx context.Context, depth int, mergeT
 	return nil
 }
 
-func (p *KnownTx) GetBEEFForTxIDs(ctx context.Context, txids iter.Seq[string], knownTxIDs []string, statusesToFilterOut []wdk.ProvenTxReqStatus) ([]byte, error) {
+func (p *KnownTx) BuildValidBEEFForTxIDs(ctx context.Context, txids iter.Seq[string], knownTxIDs []string, statusesToFilterOut []wdk.ProvenTxReqStatus) (*transaction.Beef, error) {
```
dorzepowski: ### Suggestion

What does it mean `Valid`?
Why it need a name `Build`.

In other words, IMHO previous name was better when looking from domain perspective.
Maybe in other repository we won't need to build it, we would just select it with query (imagine we have neo4j attached)

pkg/storage/internal/actions/process.go
```diff
@@ -215,24 +232,33 @@ func (p *process) newStatuses(args *wdk.ProcessActionArgs) (txStatus wdk.TxStatu
...
+	}
+
+	sendStatusesLookup, err := p.getSendStatuses(ctx, txIDs...)
 	if err != nil {
 		return nil, err
 	}
 
-	if sendStatus != wdk.SendWithResultStatusSending {
+	sendStatuses := maps.Values(sendStatusesLookup)
+	if seq.ContainsAll(sendStatuses, wdk.SendWithResultStatusUnproven) {
```
dorzepowski: Are you sure this is what you want to use?
Shouldn't It be rather:
```suggestion
	if seq.Every(sendStatuses, is.EqualTo(wdk.SendWithResultStatusUnproven)) {
```

Otherwise we won't broadcast any transaction if at least one would have status "unproven"

pkg/storage/internal/actions/process.go
```diff
@@ -215,24 +232,33 @@ func (p *process) newStatuses(args *wdk.ProcessActionArgs) (txStatus wdk.TxStatu
...
-			SendWithResults: []wdk.SendWithResult{
-				{
+			SendWithResults: seq.Collect(seq2.MapTo(maps.All(sendStatusesLookup), func(txID string, status wdk.SendWithResultStatus) wdk.SendWithResult {
+				return wdk.SendWithResult{
 					TxID:   primitives.TXIDHexString(txID),
-					Status: sendStatus,
-				},
-			},
+					Status: status,
+				}
```
dorzepowski: shouldn't we also add "NotDelayedResults"?

pkg/storage/internal/actions/process.go
```diff
@@ -243,30 +269,54 @@ func (p *process) broadcastSingleTx(ctx context.Context, txID string) (*wdk.Proc
...
-	if !ok {
-		return nil, fmt.Errorf("failed to find aggregated result for txID %s", txID)
-	}
+	sendWithResults := make([]wdk.SendWithResult, 0, len(txIDs))
+	notDelayedResults := make([]wdk.ReviewActionResult, 0, len(txIDs))
+
+	aggregated := results.Aggregated(txIDs)
+	for _, txID := range txIDs {
+		aggBroadcastResult, ok := aggregated[txID]
+		if !ok {
```
dorzepowski, line 296: (this comment is valid only if the comment about seq.ContainsAll(sendStatuses, wdk.SendWithResultStatusUnproven) is valid )


maybe I'm wrong, but if one txID would be "unproven" then we would return here error, for that transaction

pkg/storage/internal/actions/process.go
```diff
@@ -45,22 +45,39 @@ func newProcessAction(logger *slog.Logger, txRepo TransactionsRepo, commissionCf
 }
 
 func (p *process) Process(ctx context.Context, userID int, args *wdk.ProcessActionArgs) (*wdk.ProcessActionResult, error) {
```
dorzepowski, line 47: It would be good to add some tests (especially for corner cases) for ProcessAction on storage with NoSend and SendWith

pkg/storage/internal/actions/process.go
```diff
@@ -215,24 +232,33 @@ func (p *process) newStatuses(args *wdk.ProcessActionArgs) (txStatus wdk.TxStatu
...
 		return nil, err
 	}
 
-	if sendStatus != wdk.SendWithResultStatusSending {
+	sendStatuses := maps.Values(sendStatusesLookup)
+	if seq.ContainsAll(sendStatuses, wdk.SendWithResultStatusUnproven) {
 		return &wdk.ProcessActionResult{
-			SendWithResults: []wdk.SendWithResult{
-				{
+			SendWithResults: seq.Collect(seq2.MapTo(maps.All(sendStatusesLookup), func(txID string, status wdk.SendWithResultStatus) wdk.SendWithResult {
```
dorzepowski: ### minor thing

there are multiple places where `maps.All(sendStatusesLookup)` is called
you could create a variable just after p.getSendStatuses(ctx, txIDs...) as it will be used in all places.

Or you could use `seq2.FromMap` as this is the same - and would keep the code in the same namespace

pkg/storage/internal/actions/process.go
```diff
@@ -243,30 +269,54 @@ func (p *process) broadcastSingleTx(ctx context.Context, txID string) (*wdk.Proc
...
+		}
+
+		err = p.knownTxRepo.UpdateKnownTxStatus(ctx, txID, newReqStatus, wdk.ProvenTxReqBeyondBroadcastStageStatuses, notes)
+		if err != nil {
+			return nil, fmt.Errorf("failed to update transaction status after broadcast: %w", err)
+		}
+
+		sendWithResults = append(sendWithResults, sendWithResult)
+		notDelayedResults = append(notDelayedResults, reviewActionResult)
 	}
```
dorzepowski: It would be nice to have method for creation of result from post beef result

pkg/storage/internal/actions/process.go
chris-4chain: According to the TS version: `no`.
If `args.IsNoSend && len(args.SendWith) == 0`
then we don't reach this point

Otherwise, if `SendWith` is not empty, the `IsNoSend` is ignored and all the `SendWith` txid and the "new transaction" as well are sent.
See the "note" comment, I made:
> // NOTE: SendWith overrides IsNoSend, so if SendWith is NOT empty, we will broadcast txs anyway

pkg/storage/internal/actions/process.go
chris-4chain: ☝️

pkg/storage/internal/actions/process.go
chris-4chain: I'm not so sure. And if the implementator is not sure - I think, some "defensive programming" is acceptable. 
Besides, now maybe it won't happen, but it can be small change in validation when this starts to be necessary.

pkg/internal/storage/repo/known_tx.go
chris-4chain: This is legacy from TS naming. But you have valid point.
I'll rename `BuildValidBEEFForTxIDs` -> `GetBEEFForTxIDs` and `BuildValidBEEFForTxID` -> `GetBEEFForTxID`

pkg/storage/internal/actions/process.go
chris-4chain: Good point. I misunderstood the `ContainsAll` function purpose.

pkg/storage/internal/actions/process.go
chris-4chain: According to `TS` version: `no`.
In there, the `ndr` (Not Delayed Result) field is added **after** the check:
```js
if (readyToSendReqIds.length < 1) {
    return createResults()
  }
```

pkg/storage/internal/actions/process.go
chris-4chain: I've checked `ContainsAll` -> `Every`
- So I guess, this comment in not applicable.

pkg/storage/internal/actions/process.go
chris-4chain: It's not so easy - that's why I decided to test it from "wallet" side. 
This way, it covers most of cases. 

But sure, we could discuss it, because it would require more tweaks.

pkg/storage/internal/actions/process.go
chris-4chain: I did it, but for me the variable name was less readable than explicit `maps.All(sendStatusesLookup)`.
And since this doesn't have any influence on the performance, I gave up.

pkg/storage/internal/actions/process.go
chris-4chain: I refactored this part a little bit - check it out 👍

mgosek-4chain (approved):

pkg/storage/internal/actions/process.go
dorzepowski: my mistake

pkg/storage/internal/actions/process.go
dorzepowski: it's a job for Superman .... no, Tests

pkg/storage/internal/actions/process.go
dorzepowski: actually it is if ContainsAll -> Every

pkg/storage/internal/actions/process.go
dorzepowski: As we talked, it will be in another PR

pkg/storage/internal/actions/process.go
chris-4chain: ok, ok 😅 - removed. More tests will be added when we create possibility to "sign" transactions on storage level for tests

pkg/storage/internal/actions/process.go
dorzepowski: maybe to be consistent, let's return NotDelayedResults always

dorzepowski (approved):
LGETM

PR #433, opened 2025-07-30
fix: not used sender identity key returned by storage for unlocking input.

dorzepowski:


mgosek-4chain (approved):

pkg/wallet/internal/actions/wallet_create_action_transaction_assembler.go
```diff
@@ -131,7 +131,7 @@ func (a *createActionTransactionAssembler) toTxInputFromManagedInput(it *wdk.Sto
 			DerivationPrefix: to.Value(it.DerivationPrefix),
 			DerivationSuffix: to.Value(it.DerivationSuffix),
 		},
-		a.keyDeriver,
+		to.ValueOr(it.SenderIdentityKey, a.keyDeriver.IdentityKeyHex()),
```
chris-4chain: Since the `IdentityKeyHex` method is not trivial, I would calculate it only when necessary:

```golang
	var recipientPrivateKeySource string
	if it.SenderIdentityKey != nil {
		recipientPrivateKeySource = *it.SenderIdentityKey
	} else {
		recipientPrivateKeySource = a.keyDeriver.IdentityKeyHex()
	}

	// TODO: in TS they don't create unlocking script template at this stage, but rather create it later
	// 	BUT they need then store the KeyID for each input in the wallet instance
	//  or need to query for those inputs the storage
	//  for now (until SignAction is implemented) we create unlocking script template here.
	input.UnlockingScriptTemplate, err = brc29.Unlock(
		a.keyDeriver,
		brc29.KeyID{
			DerivationPrefix: to.Value(it.DerivationPrefix),
			DerivationSuffix: to.Value(it.DerivationSuffix),
		},
		recipientPrivateKeySource,
	)
```

chris-4chain (changes requested):

chris-4chain (approved):

PR #441, opened 2025-08-04
fix: not deterministic process-action result order

chris-4chain:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter

### DIFF
--- a/pkg/storage/internal/actions/process.go
+++ b/pkg/storage/internal/actions/process.go
@@ -257,9 +257,13 @@ func (p *process) broadcastTxs(ctx context.Context, txIDs []string) (*wdk.Proces
 		}, nil
 	}
 
-	readyToSendTxIDs := seq.Collect(seq2.Keys(seq2.Filter(maps.All(sendStatusesLookup), func(txID string, status wdk.SendWithResultStatus) bool {
-		return status == wdk.SendWithResultStatusSending || status == wdk.SendWithResultStatusFailed
-	})))
+	var readyToSendTxIDs []string
+	for _, txID := range txIDs {
+		status := sendStatusesLookup[txID]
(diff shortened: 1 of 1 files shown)


pkg/storage/internal/actions/process.go
```diff
@@ -257,9 +257,13 @@ func (p *process) broadcastTxs(ctx context.Context, txIDs []string) (*wdk.Proces
...
-	readyToSendTxIDs := seq.Collect(seq2.Keys(seq2.Filter(maps.All(sendStatusesLookup), func(txID string, status wdk.SendWithResultStatus) bool {
-		return status == wdk.SendWithResultStatusSending || status == wdk.SendWithResultStatusFailed
-	})))
+	var readyToSendTxIDs []string
+	for _, txID := range txIDs {
+		status := sendStatusesLookup[txID]
+		if status == wdk.SendWithResultStatusSending || status == wdk.SendWithResultStatusFailed {
+			readyToSendTxIDs = append(readyToSendTxIDs, txID)
+		}
+	}
```
ac4ch, line 266: would not be better to move this to some helper function ?

ac4ch (approved):

pkg/storage/internal/actions/process.go
chris-4chain: In this case it isn't so beneficial because:
1. This code assumes that all the `txIDs` are in `sendStatusesLookup` - which is true only in this function flow
2. The `process` action has already complex structure and many helper functions - I would like to limit them.

ac4ch: sure ofc ...

mgosek-4chain (approved):

PR #462, opened 2025-08-11
[FEATURE] WalletServices: create getStatusForTxids method

### LINKED ISSUE #170: [FEATURE] WalletServices: create getStatusForTxids method

ac4ch:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter


examples/services_examples/services_get_status_for_txids/get_status_for_txids.go
```diff
@@ -0,0 +1,70 @@
...
+		"866c535d32935d97e021c06e81d824418cf257af1604fcaeb153afb5fe82cbe2",
+		"ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff",
+	}
+
+	show.Step("Wallet-Services", "query depth/status for multiple txids")
+
+	res, err := srv.GetStatusForTxids(context.Background(), txids)
+	if err != nil {
+		show.WalletError("GetStatusForTxids", map[string]any{"txids": txids}, err)
+		return
```
mgosek-4chain: As far as I remember we decided to throw a panic.

mgosek-4chain (approved):

pkg/wdk/services_tx_status.go
```diff
@@ -0,0 +1,42 @@
...
+
+// TxStatusDetail holds the status of a single txid
+type TxStatusDetail struct {
+	TxID   string
+	Depth  *int
+	Status string
+}
+
+// GetStatusForTxidsResult represents result of a GetStatusForTxids query
+type GetStatusForTxidsResult struct {
```
chris-4chain:
```suggestion
type GetStatusForTxIDsResult struct {
```

pkg/wdk/services_tx_status.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+// GetStatusResult represents the status of a GetStatusForTxids query
+type GetStatusResult string
+
+const (
+	// GetStatusSuccess indicates the query was successful
+	GetStatusSuccess GetStatusResult = "success"
+	// GetStatusError indicates the query encountered an error
+	GetStatusError GetStatusResult = "error"
```
chris-4chain: The `GetStatusError` seems unused.

pkg/services/services.go
```diff
@@ -329,3 +338,18 @@ func (s *WalletServices) IsUtxo(ctx context.Context, scriptHash string, outpoint
 
 	return result, nil
 }
+
+// GetStatusForTxids returns depth/status info for a list of txids.
+func (s *WalletServices) GetStatusForTxids(ctx context.Context, txids []string) (*wdk.GetStatusForTxidsResult, error) {
```
chris-4chain: Only suggestion, apply if you aggree:
`txIDs []string`

pkg/services/services.go
```diff
@@ -329,3 +338,18 @@ func (s *WalletServices) IsUtxo(ctx context.Context, scriptHash string, outpoint
...
+	if len(txids) == 0 {
+		return nil, fmt.Errorf("no txids provided")
+	}
```
chris-4chain, line 346: Please add empty line after "return-error" block.

pkg/services/internal/whatsonchain/get_status_for_txids.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/wdk"
+)
+
+func (woc *WhatsOnChain) getStatusForTxids(ctx context.Context, url string, txids []string) (*wdk.GetStatusForTxidsResult, error) {
+	response, err := woc.doStatusRequest(ctx, url, txids)
+	if err != nil {
+		return nil, err
+	}
+
+	results := mapWocStatusResponse(response)
```
chris-4chain: Let's try to stick to a single method of mapping slices.
```suggestion
	results := slices.Map(response, mapSingleTxStatus)
```

pkg/services/internal/bitails/helpers.go
```diff
@@ -12,6 +12,30 @@ import (
...
+//
+//	found=false on 404
+//	found=true  with mined=true/false and height (if mined) on 200
+func (b *Bitails) getTxStatus(ctx context.Context, txid string) (found bool, mined bool, height int64, err error) {
+	url, err := txStatusURL(b.url, txid)
+	if err != nil {
+		return false, false, 0, fmt.Errorf("build tx status URL: %w", err)
+	}
+
+	var info dto.FetchInfoResponse
```
chris-4chain, line 23: Shouldn't `BlockHeight` in the `dto.FetchInfoResponse` be `uint64` or even `uint32`?

pkg/services/internal/bitails/helpers.go
```diff
@@ -12,6 +12,30 @@ import (
...
+	if err != nil {
+		return false, false, 0, fmt.Errorf("fetch tx status: %w", err)
+	}
+	if !found {
+		return false, false, 0, nil
+	}
+	if info.BlockHeight > 0 {
+		return true, true, info.BlockHeight, nil
+	}
+	return true, false, 0, nil
```
chris-4chain: Wouldn't it be more readable?

```suggestion

	mined = info.BlockHeight > 0
	height = info.BlockHeight
	return
```

pkg/services/internal/bitails/service.go
```diff
@@ -275,3 +276,63 @@ func (b *Bitails) GetScriptHashHistory(ctx context.Context, scriptHash string) (
...
+			} else {
+				dd := int(d)
+				item.Depth = &dd
+			}
+
+		default:
+			anyFound = true
+			item.Status = wdk.TxStatusUnconfirmed.String()
+			zero := 0
+			item.Depth = &zero
```
chris-4chain:
```suggestion
			item.Depth = to.Ptr(0)
```

pkg/services/internal/bitails/service.go
```diff
@@ -275,3 +276,63 @@ func (b *Bitails) GetScriptHashHistory(ctx context.Context, scriptHash string) (
...
+			item.Status = wdk.TxStatusMined.String()
+
+			d := int64(tip) - height + 1
+			if d < 0 {
+				zero := 0
+				item.Depth = &zero
+			} else {
+				dd := int(d)
+				item.Depth = &dd
+			}
```
chris-4chain: Please move this logic into some helper method.

pkg/services/internal/bitails/service.go
chris-4chain: If the status is unconfirmed, should it have Depth? I wonder, if this can be nil, maybe this is the case. Please cross check it with ts-wallet-toolbox behaviour.

pkg/services/internal/whatsonchain/testabilities/tx_status_fixtures.go
```diff
@@ -0,0 +1,12 @@
...
+
+type TxStatusExpectation struct {
+	TxID                string
+	ExpectBlockHash     string
+	ExpectBlockHeight   int64
+	ExpectConfirmations int
+}
+
+const (
+	DefaultConfirmations = 10
```
chris-4chain: This is not used.

pkg/services/internal/bitails/service.go
ac4ch: if unconfirmed depth should be 0 .. if not found depth should be nil

pkg/services/internal/bitails/helpers.go
ac4ch: image
that why i used int64 will try to correct ur code too ..

ac4ch: image

pkg/services/internal/whatsonchain/testabilities/tx_status_fixtures.go
ac4ch: ok removed it

pkg/services/services.go
ac4ch: sure i changed .. as i agree but .. before why i did this way as you were doing same before 
image
so was trying to match it ...

pkg/services/services.go
ac4ch: added empty line

pkg/services/internal/bitails/helpers.go
chris-4chain: We're still improving our code. Nothing is "set in stone". It's hard to tell who introduced this first and who only "was only trying to match it". And there is no point to investigate or blame.

pkg/services/services.go
chris-4chain: I can find many examples where I (and you and others) used `txIDs` name. 
You can always adjust this code (per `Aggregatd` method) if you found inconsistency. 

image

pkg/services/services.go
```diff
@@ -330,6 +339,22 @@ func (s *WalletServices) IsUtxo(ctx context.Context, scriptHash string, outpoint
...
+// GetStatusForTxIDs returns depth/status info for a list of txIDs.
+func (s *WalletServices) GetStatusForTxIDs(ctx context.Context, txIDs []string) (*wdk.GetStatusForTxIDsResult, error) {
+	if len(txIDs) == 0 {
+		return nil, fmt.Errorf("no txIDs provided")
+	}
+
+	res, err := s.getStatusForTxIDsServices.OneByOne(ctx, txIDs)
+	if err != nil {
+		if errors.Is(err, servicequeue.ErrEmptyResult) {
+			return nil, fmt.Errorf("no status found for provided txIDs")
```
chris-4chain: It's sometimes necessary to check if there was "not found" situation.
```suggestion
			return nil, fmt.Errorf("no status found for provided txIDs: %w", wdk.NotFoundError)
```

For this, general-purpose `wdk.NotFoundError` was created.

This is also applicable to other functions. Could you do this as your next task?

pkg/services/internal/bitails/helpers.go
```diff
@@ -12,6 +12,30 @@ import (
 	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/wdk"
 )
 
+// getTxStatus returns:
+//
+//	found=false on 404
+//	found=true  with mined=true/false and height (if mined) on 200
```
chris-4chain: Please improve readability of this doc comment.

pkg/services/internal/bitails/helpers.go
```diff
@@ -12,6 +12,30 @@ import (
...
+func (b *Bitails) getTxStatus(ctx context.Context, txid string) (found bool, mined bool, height uint32, err error) {
+	url, err := txStatusURL(b.url, txid)
+	if err != nil {
+		return false, false, 0, fmt.Errorf("build tx status URL: %w", err)
```
chris-4chain: Since you're using named return values.
```suggestion
		err = fmt.Errorf("build tx status URL: %w", err)
		return
```

pkg/services/internal/bitails/helpers.go
```diff
@@ -12,6 +12,30 @@ import (
...
+	}
+
+	var info dto.FetchInfoResponse
+	found, err = b.handleJSON(ctx, url, &info, http.StatusOK, true /* allow 404 */)
+	if err != nil {
+		return false, false, 0, fmt.Errorf("fetch tx status: %w", err)
```
chris-4chain: And here:
```suggestion
		err = fmt.Errorf("fetch tx status: %w", err)
		return
```

pkg/services/internal/bitails/helpers.go
```diff
@@ -12,6 +12,30 @@ import (
...
+	}
+	if !found {
+		return false, false, 0, nil
```
chris-4chain: Without this `if !found` the result is the same. Or I'm missing something?

pkg/wdk/services_tx_status.go
```diff
@@ -0,0 +1,40 @@
...
+// GetStatusResult represents the status of a GetStatusForTxids query
+type GetStatusResult string
+
+const (
+	// GetStatusSuccess indicates the query was successful
+	GetStatusSuccess GetStatusResult = "success"
+)
+
+// Status represents the status of a transaction
+type Status string
```
chris-4chain: The wdk package is so wide and this `Status` is only applicable to the result of `getStatusForTxids` method in services.
1. I think, the name is too broad and can be misleading
2. The comment: "Status represents the status of a transaction" is true but in the wdk we have already "wdk.TxStatus" that refers to `wallet-toolbox's` transaction.
3. You mapped status "TxStatusUnconfirmed" -> "known" - I would suggest: "TxStatusKnown" -> "known"
4. I would name this type similar to TS: `ResultStatusForTxID`
5. and `ResultStatusForTxIDMined`, `...Known`, `...NotFound`

pkg/services/internal/whatsonchain/service.go
```diff
@@ -358,3 +358,23 @@ func (woc *WhatsOnChain) IsUtxo(ctx context.Context, scriptHash string, outpoint
...
+	url, err := txsStatusURL(woc.url)
+	if err != nil {
+		return nil, fmt.Errorf("failed to build URL: %w", err)
+	}
+
+	results, err := woc.getStatusForTxIDs(ctx, url, txIDs)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get status for txIDs: %w", err)
+	}
+	if results == nil {
```
chris-4chain, line 377:
```suggestion
	if len(results) == 0 {
```

If `getStatusForTxIDs` returns a pointer but with an empty Results array (but not nil), this would fall through and appear as success, even if no data was returned. Meanwhile, other implementations (see Bitails) explicitly return an error if no results are found. This may cause inconsistent handling.

pkg/services/internal/bitails/utils.go
```diff
@@ -76,6 +77,14 @@ func validateScriptHash(scriptHash string) error {
 	return nil
 }
 
+func calcDepth(tip, height uint32) *int {
+	d := int(tip) - int(height) + 1
+	if d < 0 {
+		return to.Ptr(0)
```
chris-4chain: If `tip < height`, you clamp to zero. However, in blockchain systems, it should not normally be possible for a block to be ahead of tip, except if there are data inconsistencies or bugs elsewhere. Returning zero in such an event may mask data issues and generate misleading "0 confirmation" states.
Suggestion:
Consider logging or explicitly erroring when `height > tip`

pkg/services/internal/whatsonchain/utils.go
```diff
@@ -93,6 +94,31 @@ func containsI(subject string, contains ...string) bool {
...
+	var (
+		depth  *int
+		status wdk.Status
+	)
+
+	switch {
+	case tx.Confirmations > 0:
+		depth = to.Ptr(tx.Confirmations)
+		status = wdk.TxStatusMined
+	case tx.Confirmations == 0 && tx.BlockHash != "":
```
chris-4chain: This condition, I think, is not ok.
1. I guess, `tx.Confirmations` should be nillable to distinguish `0` and "undefined/null" confirmation.
2. `Unconfirmed` status should be when "confirmations" are zero
3. If a tx has BlockHash it should have also at leat `one` confirmation. This is how blockchain works. 
4. If `tx.Confirmations == 0 && tx.BlockHash != ""` it means this is some inconsistency error.

Please recheck your logic with the source TS code:
https://github.com/4chain-ag/wallet-toolbox/blob/f8f74ca8625e5b8ea850aee38de18be671e44042/src/services/providers/WhatsOnChain.ts#L58-L71

chris-4chain (changes requested):

pkg/services/internal/whatsonchain/service.go
ac4ch: invalid argument: results (variable of type *wdk.GetStatusForTxIDsResult) for built-in len
```go
if results == nil {
		return nil, fmt.Errorf("no status found for provided txIDs")
	}
	if len(results.Results) == 0 {
		return nil, fmt.Errorf("no results found for provided txIDs")
	}
```

jakubmkowalski (approved):

pkg/services/internal/whatsonchain/service.go
chris-4chain: I'm wondering should we return somehow information that it is "not found" situation. 🤔 
For example like this:
```golang
return nil, fmt.Errorf("no status found for provided txIDs: %w", wdk.NotFoundError)
```

pkg/services/internal/whatsonchain/utils.go
chris-4chain: I like the way you solved it 👍

chris-4chain (approved):

pkg/services/internal/whatsonchain/service.go
ac4ch: ok i added check for status

PR #470, opened 2025-08-11
feat(#463): add wallet IsAuthenticated

### LINKED ISSUE #463: [FEATURE] Wallet: IsAuthenticated

jakubmkowalski:
Closes #463
This pull request implements the `IsAuthenticated` method in the `Wallet` struct and adds comprehensive tests to validate its behavior, especially focusing on the validation of the `originator` parameter. The main changes are grouped as follows:

Implementation:

* Implemented the `IsAuthenticated` method in `wallet.go` to validate the `originator` argument and return an authenticated result if valid. The method now checks for invalid originators and returns an error if validation fails.

Testing:

* Added a new test file `wallet_is_authenticated_test.go` with tests for `IsAuthenticated`, covering both error cases (invalid originator formats) and success cases (various valid arguments and originators). These tests ensure that the method properly validates input and returns the expected results.463


pkg/wallet/wallet.go
```diff
@@ -355,8 +355,13 @@ func (w *Wallet) DiscoverByAttributes(ctx context.Context, args sdk.DiscoverByAt
 
 // IsAuthenticated checks the authentication status of the user.
 func (w *Wallet) IsAuthenticated(ctx context.Context, args any, originator string) (*sdk.AuthenticatedResult, error) {
```
chris-4chain:
```suggestion
func (w *Wallet) IsAuthenticated(_ context.Context, _ any, originator string) (*sdk.AuthenticatedResult, error) {
```

chris-4chain (approved):

pkg/wallet/wallet_is_authenticated_test.go
```diff
@@ -0,0 +1,95 @@
...
+	"strings"
+	"testing"
+
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/internal/fixtures"
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/wallet/internal/testabilities"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestWalletIsAuthenticatedArgsValidation(t *testing.T) {
```
ac4ch, line 13: consider to add more test cases if you think my suggestion make sense to you :) 
```go
		{
			name:       "max single label length (63) across multiple parts",
			originator: label63 + "." + label63 + "." + label63,
			args:       map[string]any{"k": "v"},
		},
		{
			name:       "max total length (250 chars) should pass",
			originator: valid250,
			args:       struct{ Foo string }{Foo: "bar"},
		},
		{
			name:       "default originator still passes",
			originator: fixtures.DefaultOriginator,
			args:       nil,
		},
		{
			name:       "case-insensitive originator (if normalization is supported)",
			originator: strings.ToUpper(fixtures.DefaultOriginator),
			args:       nil,
		},
		{
			name:       "originator with 2 labels and simple map args",
			originator: "example.com",
			args:       map[string]string{"key": "value"},
		},
``` 
WDYT ?

ac4ch (approved):

pkg/wallet/wallet_is_authenticated_test.go
jakubmkowalski: added more test cases

PR #478, opened 2025-08-13
feat: background broadcaster

chris-4chain:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter


pkg/storage/internal/service/background_broadcaster.go
```diff
@@ -0,0 +1,104 @@
...
+	for i := 0; i < BackgroundBroadcasterWorkerCount; i++ {
+		bb.wg.Add(1)
+		go bb.worker()
+	}
+}
+
+func (bb *BackgroundBroadcaster) Stop() {
+	bb.cancel()
+	bb.wg.Wait()
+	close(bb.broadcastChannel)
```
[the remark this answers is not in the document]
chris-4chain: I stop method follows this sequence:

1. `bb.cancel()`: Signals all workers that they should begin shutting down.
2. `bb.wg.Wait()`: Waits for all worker goroutines to finish completely and call `wg.Done()`.
3. `close(bb.broadcastChannel)`: Closes the channel.

This order is crucial because it prevents a panic. The golden rule of Go channels is: never write to a closed channel.

If I were to close the channel first (`close -> cancel -> wait`), I would create a race condition. Another part of the app could call the Add method after the channel has been closed but before the entire service is fully stopped. This would cause the program to panic with a send on closed channel error.

By waiting for the workers to stop first (wg.Wait()), I guarantee that the only goroutines that could possibly interact with the channel are shut down. Closing the channel becomes the final, safe cleanup step.

pkg/internal/fixtures/walletargs/wallet_create_action_args_opts.go
```diff
@@ -32,3 +32,18 @@ func WithSendWith(sendWith ...chainhash.Hash) func(args *wallet.CreateActionArgs
...
 	}
 }
+
+func WithDelayedBroadcast() func(args *wallet.CreateActionArgs) {
+	return func(args *wallet.CreateActionArgs) {
+		args.Options.AcceptDelayedBroadcast = to.Ptr(true)
+	}
+}
+
+func WithSatoshisAsFirstOutput(satoshis uint64) func(args *wallet.CreateActionArgs) {
```
mgosek-4chain, line 42: Minor thing: maybe it's worth creating a non-exported type for `func(args *wallet.CreateActionArgs)` to simplify future function option definition?  🤔

pkg/internal/storage/repo/syncrepo/sync_output.go
```diff
@@ -234,3 +247,18 @@ func (s *SyncOutput) mapModelToTableOutput(model *OutputReadModel) *wdk.TableOut
...
+		return wdk.UTXOStatusMined
+	case wdk.TxStatusSending:
+		return wdk.UTXOStatusSending
+	case wdk.TxStatusUnproven:
+		return wdk.UTXOStatusUnproven
+	case wdk.TxStatusFailed, wdk.TxStatusUnprocessed, wdk.TxStatusUnsigned, wdk.TxStatusNoSend, wdk.TxStatusNonFinal, wdk.TxStatusUnfail:
+		fallthrough
+	default:
+		return wdk.UTXOStatusUnknown
+	}
```
mgosek-4chain, line 263: If I understand it correctly, this falls through `default`.  If so, maybe simplify it to a single `default` case is a good idea? 🤔

pkg/storage/internal/actions/process.go
```diff
@@ -42,13 +54,13 @@ func newProcessAction(logger *slog.Logger, txRepo TransactionsRepo, commissionCf
 		commissionRepo: commissionRepo,
 		services:       services,
 	}
+
+	p.backgroundBroadcaster = service.NewBackgroundBroadcaster(ctx, logger, p)
```
mgosek-4chain, line 58: Could you please clarify if there's any reason to avoid inline initialization of the background broadcaster? 🤔

pkg/storage/internal/actions/process.go
```diff
@@ -503,3 +555,40 @@ func (p *process) singleTxBroadcastResult(aggBroadcastResult *wdk.AggregatedPost
...
+			ctx,
+			broadcastedTxID,
+			aggBroadcastResult,
+			results.ServiceErrors(),
+			beef,
+			txIDs,
+		)
+		if err != nil {
+			return fmt.Errorf("failed to update single tx after background broadcast: %w", err)
+		}
```
mgosek-4chain, line 588: Question about `updateSingleTx` implementation: is the update operation done in transactional manner? 🤔

mgosek-4chain (changes requested):

mgosek-4chain (approved):

pkg/internal/fixtures/walletargs/wallet_create_action_args_opts.go
chris-4chain: This wouldn't simplify things much. And because I haven't introduced this concept here, I would like not to adjust it in this PR. But I'll keep this in mind for future.

pkg/storage/internal/actions/process.go
chris-4chain: because `NewBackgroundBroadcaster` requires `p` as argument. 
`p` as `process` implements `BackgroundBroadcast` method that is required by the BackgroundBroadcaster

jakubmkowalski (approved):

PR #495, opened 2025-08-20
feat: delayed broadcast task

chris-4chain:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter

### DIFF
--- a/pkg/monitor/internal/tasks/send_waiting_test.go
+++ b/pkg/monitor/internal/tasks/send_waiting_test.go
@@ -9,7 +9,7 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestSynchronizeTransactionStatuses(t *testing.T) {
+func TestSendWaitingMonitorTask(t *testing.T) {
 	// given:
 	given, then := testabilities.New(t)
 
@@ -20,7 +20,7 @@ func TestSynchronizeTransactionStatuses(t *testing.T) {
 
 	// when:
 	err := daemon.Start(map[defs.MonitorTask]defs.TaskConfig{
-		defs.CheckForProofsMonitorTask: {
+		defs.SendWaitingMonitorTask: {
 			Enabled:          true,
 			IntervalSeconds:  seconds,
 			StartImmediately: false,
@@ -29,8 +29,32 @@ func TestSynchronizeTransactionStatuses(t *testing.T) {
 	require.NoError(t, err)
 
 	// then:
-	then.SynchronizeTransactionStatuses().
+	then.SendWaitingTransactions().
 		WaitForTaskExecution(taskInterval).
 		ExecutedInTime().
 		Called()
 }
+
+func TestSendWaitingMonitorTask_StartedImmediately(t *testing.T) {
+	// given:
+	given, then := testabilities.New(t)
+
+	daemon := given.Daemon()
+
+	// when:
+	err := daemon.Start(map[defs.MonitorTask]defs.TaskConfig{
+		defs.SendWaitingMonitorTask: {
+			Enabled:          true,
+			IntervalSeconds:  1,
+			StartImmediately: true,
+		},
+	})
+	require.NoError(t, err)
+
+	// then:
+	then.SendWaitingTransactions().
+		WaitForTaskExecution(100 * time.Millisecond).
+		ExecutedInTime().
+		Called()
+}
+
--- a/pkg/monitor/internal/tasks/send_waiting_test.go
+++ b/pkg/monitor/internal/tasks/send_waiting_test.go
@@ -11,6 +11,7 @@ import (
 )
 
 func TestSendWaitingMonitorTask(t *testing.T) {
+	t.Parallel()
 	// given:
 	given, then := testabilities.New(t)
 
@@ -37,6 +38,7 @@ func TestSendWaitingMonitorTask(t *testing.T) {
 }
 
 func TestSendWaitingMonitorTask_StartedImmediately(t *testing.T) {
+	t.Parallel()
 	// given:
 	given, then := testabilities.New(t)
 
@@ -60,6 +62,7 @@ func TestSendWaitingMonitorTask_StartedImmediately(t *testing.T) {
 }
 
 func TestSendWaitingMonitorTask_FirstRunWithZeroAgedLimit(t *testing.T) {
+	t.Parallel()
 	// given:
 	mockStorage := &testabilities.MockStorage{}
 	task := tasks.NewSendWaitingTask(mockStorage)
--- a/pkg/monitor/internal/tasks/send_waiting_test.go
+++ b/pkg/monitor/internal/tasks/send_waiting_test.go
@@ -57,4 +57,3 @@ func TestSendWaitingMonitorTask_StartedImmediately(t *testing.T) {
 		ExecutedInTime().
 		Called()
 }
-
--- a/pkg/monitor/internal/tasks/send_waiting_test.go
+++ b/pkg/monitor/internal/tasks/send_waiting_test.go
@@ -5,6 +5,7 @@ import (
 	"time"
 
 	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/defs"
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/monitor/internal/tasks"
 	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/monitor/internal/testabilities"
 	"github.com/stretchr/testify/require"
 )
@@ -56,4 +57,26 @@ func TestSendWaitingMonitorTask_StartedImmediately(t *testing.T) {
 		WaitForTaskExecution(100 * time.Millisecond).
 		ExecutedInTime().
 		Called()
+}
+
+func TestSendWaitingMonitorTask_FirstRunWithZeroAgedLimit(t *testing.T) {
+	// given:
+	mockStorage := &testabilities.MockStorage{}
+	task := tasks.NewSendWaitingTask(mockStorage)
+
+	// when:
+	err := task.Run(t.Context())
+	require.NoError(t, err, "task should run without error")
+
+	// then:
+	require.Equal(t, 1, mockStorage.SendWaitingTransactionsCalled)
+	require.Equal(t, time.Duration(0), mockStorage.SendWaitingLastAgedLimit)
+
+	// when:
+	err = task.Run(t.Context())
+
+	// then:
+	require.NoError(t, err, "task should run without error on subsequent call")
+	require.Equal(t, 2, mockStorage.SendWaitingTransactionsCalled)
+	require.NotZero(t, mockStorage.SendWaitingLastAgedLimit)
 }
--- a/pkg/monitor/internal/tasks/send_waiting_test.go
+++ b/pkg/monitor/internal/tasks/send_waiting_test.go
@@ -61,7 +61,7 @@ func TestSendWaitingMonitorTask_StartedImmediately(t *testing.T) {
 		Called()
 }
 
-func TestSendWaitingMonitorTask_FirstRunWithZeroAgedLimit(t *testing.T) {
+func TestSendWaitingMonitorTask_FirstRunWithZeroMinTransactionAge(t *testing.T) {
 	t.Parallel()
 	// given:
 	mockStorage := &testabilities.MockStorage{}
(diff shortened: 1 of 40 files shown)


mgosek-4chain (approved):

pkg/monitor/internal/tasks/send_waiting_test.go
```diff
@@ -0,0 +1,82 @@
...
+	"time"
+
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/defs"
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/monitor/internal/tasks"
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/monitor/internal/testabilities"
+	"github.com/stretchr/testify/require"
+)
+
+func TestSendWaitingMonitorTask(t *testing.T) {
+	// given:
```
dorzepowski, line 15: As all those tests in tasks_tests are pretty long,
consider using t.Parallel in all of them 🙈

pkg/storage/internal/actions/synchronize_tx_statuses.go
```diff
@@ -11,13 +11,15 @@ import (
...
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/internal/storage/queryopts"
 	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/wdk"
 	"github.com/go-softwarelab/common/pkg/slices"
 )
 
 const (
-	syncTxStatusLimit  = 100
-	lastBlockHeightKey = "synchronize_tx_statuses_last_block_height"
+	syncTxStatusMaxPages  = 100
+	syncTxStatusesPerPage = 100
```
dorzepowski: why only 100?
IMHO it would be better to have bigger page (like 1000) and less pages

pkg/storage/internal/actions/synchronize_tx_statuses.go
```diff
@@ -91,11 +93,22 @@ func (s *synchronizeTxStatuses) SynchronizeTxStatuses(ctx context.Context) (resu
...
-		return fmt.Errorf("knownTxRepo.FindTxIDsByStatuses failed: %w", err)
+	var txsToSync []*entity.KnownTxForStatusSync
+	paging := queryopts.Paging{Limit: syncTxStatusesPerPage, Sort: "asc"}
+	for range syncTxStatusMaxPages {
+		txsPage, err := s.provenTxRepo.FindKnownTxIDsByStatuses(ctx, statusesReadyToSync, queryopts.WithPage(paging))
+		if err != nil {
+			return fmt.Errorf("provenTxRepo.FindKnownTxIDsByStatuses failed: %w", err)
+		}
+
+		if len(txsPage) == 0 {
```
dorzepowski: ## minor optimisation
if len(txsPage) < syncTxStatusesPerPage 
then we can break (but after appending)
this way we would have one less call to repo

pkg/storage/internal/actions/send_waiting.go
```diff
@@ -0,0 +1,124 @@
...
+
+var (
+	statusesOfWaitingTxs = []wdk.ProvenTxReqStatus{
+		wdk.ProvenTxStatusUnsent,
+		wdk.ProvenTxStatusSending,
+	}
+)
+
+func (p *process) SendWaitingTransactions(ctx context.Context, agedLimit time.Duration) error {
+	log := p.logger.With("action", "send waiting transactions").With("agedLimit", agedLimit.String())
```
dorzepowski: 1. use camel case or snake case as value as it is easier to select such a value to search for logs with it.
2. Maybe use single With, and dedicated slog functions? like slog.String
3. Did you try to use slog.Duration for agedLimit?

pkg/storage/internal/actions/send_waiting.go
```diff
@@ -0,0 +1,124 @@
...
+	lockAcquired := p.sendWaitingLock.TryLock()
+	if !lockAcquired {
+		log.Warn("SendWaitingTransactions is already running, skipping this run")
+		return nil
+	}
+	defer p.sendWaitingLock.Unlock()
+
+	paging := queryopts.Paging{Limit: sendWaitingItemsPerPage, Sort: "asc"}
+	since := queryopts.Since{
+		Time: time.Now().Add(agedLimit),
```
dorzepowski: hmmm... I think that I don't get the idea.

So if for some reason the transaction would be older then 5 minutes (for example when the server is down for 5 minutes) then none of those transactions would be ever broadcasted - am I right?

pkg/storage/internal/actions/send_waiting.go
```diff
@@ -0,0 +1,124 @@
...
+	"log/slog"
+	"time"
+
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/internal/storage/queryopts"
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/wdk"
+)
+
+const (
+	sendWaitingMaxPages     = 100
+	sendWaitingItemsPerPage = 100
```
dorzepowski: again, IMHO 1000 would be better value for page

pkg/storage/internal/actions/send_waiting.go
```diff
@@ -0,0 +1,124 @@
...
+)
+
+var (
+	statusesOfWaitingTxs = []wdk.ProvenTxReqStatus{
+		wdk.ProvenTxStatusUnsent,
+		wdk.ProvenTxStatusSending,
+	}
+)
+
+func (p *process) SendWaitingTransactions(ctx context.Context, agedLimit time.Duration) error {
```
dorzepowski: why this is method on process?
maybe whole broadcasting should be put in a separate component?

pkg/storage/internal/actions/send_waiting.go
```diff
@@ -0,0 +1,124 @@
...
+
+	// TODO: Keep in mind that the transactions above max attempts will be reviewed in another "reviewStatus" periodic task.
+
+	return nil
+}
+
+func (p *process) delayedBroadcastTransaction(ctx context.Context, log *slog.Logger, txIDs []string) {
+	log.InfoContext(ctx, "Attempting to broadcast transactions", "txIDs", txIDs)
+
+	if err := p.knownTxRepo.IncreaseKnownTxAttemptsForTxIDs(ctx, txIDs); err != nil {
```
dorzepowski: why you do it here? not in a background broadcaster?

pkg/storage/internal/actions/send_waiting.go
chris-4chain: Nice catch 👍 - I used wrong scope. What I wanted to do is to use "until".
Now, I introduced `Until` scope that creates a filter: `"created_at" <= time.Now().Add(-agedLimit)`

image

pkg/storage/internal/actions/send_waiting.go
chris-4chain: The `process` component is responsible for "broadcasting". 
- It contains all of the necessary functions and subcomponents, like: background broadcaster component or a method `broadcastTxs` that is the core of this action. 
- It has methods that update the state `after` broadcasting. 

`SendWaitingTransactions` effectively does the same thing as `processAction` but "delayed". 
- It uses private method `broadcastTxs` for each obtained transaction(s), which uses "standard" processAction flow

pkg/storage/internal/actions/send_waiting.go
chris-4chain: I moved this into `broadcastTxs` method. This way for each "attempt", regardless if it's delayed or not, the counter will be increased.

pkg/storage/internal/actions/send_waiting.go
```diff
@@ -0,0 +1,124 @@
...
+
+		p.delayedBroadcastTransaction(ctx, log, txIDs)
+	}
+
+	// TODO: Keep in mind that the transactions above max attempts will be reviewed in another "reviewStatus" periodic task.
+
+	return nil
+}
+
+func (p *process) delayedBroadcastTransaction(ctx context.Context, log *slog.Logger, txIDs []string) {
```
dorzepowski: ### minor thing
shouldn't it be `broadcastDelayedTransaction` ?

pkg/storage/internal/actions/send_waiting.go
```diff
@@ -0,0 +1,119 @@
...
+)
+
+var (
+	statusesOfWaitingTxs = []wdk.ProvenTxReqStatus{
+		wdk.ProvenTxStatusUnsent,
+		wdk.ProvenTxStatusSending,
+	}
+)
+
+func (p *process) SendWaitingTransactions(ctx context.Context, agedLimit time.Duration) error {
```
dorzepowski: maybe 

```suggestion
func (p *process) SendWaitingTransactions(ctx context.Context, minTransactionAge time.Duration) error {
```

or

```suggestion
func (p *process) SendWaitingTransactions(ctx context.Context, minTransactionWaitingTime time.Duration) error {
```

pkg/storage/internal/actions/send_waiting.go
```diff
@@ -0,0 +1,119 @@
...
+		p.delayedBroadcastTransaction(ctx, log, []string{txID})
+	}
+
+	for batchName, txIDs := range batchesToBroadcast {
+		log.InfoContext(ctx, "Processing batch", "batchName", batchName, "txIDs", txIDs)
+
+		if len(txIDs) == 0 {
+			log.WarnContext(ctx, "No transactions found in batch", "batchName", batchName)
+			continue
+		}
```
dorzepowski: I'm sure that this couldn't ever happen, because you're always appending something

pkg/storage/internal/actions/send_waiting.go
```diff
@@ -0,0 +1,119 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to find known txs by statuses: %w", err)
+		}
+
+		for _, item := range txIDsPage {
+			if item.Batch != nil {
+				batchesToBroadcast[*item.Batch] = append(batchesToBroadcast[*item.Batch], item.TxID)
+			} else {
+				txIDsToBroadcast = append(txIDsToBroadcast, item.TxID)
+			}
```
dorzepowski: ### this is minor thing

because there is no big difference between broadcasting single transaction and transactions in batches.
Maybe batchesToBroadcast map would be enough?
```suggestion
			if item.Batch != nil {
				batchesToBroadcast[*item.Batch] = append(batchesToBroadcast[*item.Batch], item.TxID)
			} else {
				batchesToBroadcast[item.TxID] = append(batchesToBroadcast[item.TxID], item.TxID)
			}
```

pkg/storage/internal/actions/send_waiting.go
```diff
@@ -0,0 +1,119 @@
...
+
+		paging.Next()
+	}
+
+	if len(txIDsToBroadcast) == 0 && len(batchesToBroadcast) == 0 {
+		log.InfoContext(ctx, "No transactions found to send")
+		return nil
+	}
+
+	log.InfoContext(ctx, "Found transactions to send", "transactions count", len(txIDsToBroadcast), "batches count", len(batchesToBroadcast))
```
dorzepowski: prefere camelCase for arg key, as it's easier to search for it later in logs

pkg/storage/internal/actions/send_waiting.go
```diff
@@ -0,0 +1,119 @@
...
+	success := true
+	for _, res := range result.NotDelayedResults {
+		if res.Status != wdk.ReviewActionResultStatusSuccess {
+			success = false
+			log.WarnContext(ctx, "Problematic broadcast result", "txID", res.TxID, "status", res.Status)
+		}
+	}
+
+	if !success {
+		log.WarnContext(ctx, "Broadcasting transactions failed", "txIDs", txIDs)
```
dorzepowski, line 105: ### thinking out loud
Maybe we should return errors, and then collect them with errors.Join and log all of them also in monitor?

pkg/storage/internal/actions/send_waiting.go
chris-4chain: 🤷 👍

pkg/storage/internal/actions/send_waiting.go
chris-4chain: Clever 💯

pkg/storage/internal/actions/send_waiting.go
chris-4chain: That would make it more complicated with very similar log output. 
- I would stick to the current way.

dorzepowski (approved):

PR #506, opened 2025-08-22
feat: fail abandoned txs task & fix abortAction to restore UTXOs & improve GORMProvider initialization 

chris-4chain:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter


infra-config.example.yaml
```diff
@@ -26,6 +26,8 @@ db:
         user: postgres
     sqlite:
         connection_string: ./storage.sqlite
+fail_abandoned:
+    min_transaction_age_seconds: 300
```
dorzepowski, line 30: ### minor

Maybe this could be a part of task configuration instead?

pkg/infra/gorm_provider_options.go
```diff
@@ -0,0 +1,13 @@
+package infra
+
+import "github.com/bsv-blockchain/go-wallet-toolbox/pkg/storage"
+
+// GORMProviderOptionsFromConfig returns a slice of ProviderOption based on the given Config values for GORM storage setup.
+func GORMProviderOptionsFromConfig(cfg *Config) []storage.ProviderOption {
+	return []storage.ProviderOption{
```
dorzepowski, line 7: Going back to what we had in spv-wallet, and what we considered hard to use?

pkg/storage/internal/actions/abort_abandoned.go
```diff
@@ -0,0 +1,84 @@
...
+			log.ErrorContext(ctx, "Cannot abort transaction because some outputs are already spent", "transactionID", id, "error", err.Error())
+			continue
+		}
+
+		if err := a.abortTx(ctx, id); err != nil {
+			log.ErrorContext(ctx, "Failed to abort transaction", "transactionID", id, "error", err.Error())
+		} else {
+			log.InfoContext(ctx, "Successfully aborted transaction", "transactionID", id)
+		}
+	}
```
dorzepowski: I would prefer to use attrs when logging
```suggestion
	for _, id := range idsToAbort {
		if err := a.outputsRepo.ShouldTxOutputsBeUnspent(ctx, id); err != nil {
			log.ErrorContext(ctx, "Cannot abort transaction because some outputs are already spent", logging.Number("transactionID", id), logging.Error(err))
			continue
		}

		if err := a.abortTx(ctx, id); err != nil {
			log.ErrorContext(ctx, "Failed to abort transaction", logging.Number("transactionID", id), logging.Error(err))
		} else {
			log.InfoContext(ctx, "Successfully aborted transaction", logging.Number("transactionID", id))
		}
	}
```

pkg/storage/internal/actions/abort_abandoned.go
```diff
@@ -0,0 +1,84 @@
...
+	if len(idsToAbort) == 0 {
+		log.InfoContext(ctx, "No abandoned transactions found to fail")
+		return nil
+	}
+
+	log.InfoContext(ctx, "Found abandoned transactions to fail", slog.Int("count", len(idsToAbort)))
+
+	for _, id := range idsToAbort {
+		if err := a.outputsRepo.ShouldTxOutputsBeUnspent(ctx, id); err != nil {
+			log.ErrorContext(ctx, "Cannot abort transaction because some outputs are already spent", "transactionID", id, "error", err.Error())
```
dorzepowski: Maybe I'm wrong, and don't uderstand this case,

But reading the code it seems like a sign of a huge problem in application state. Maybe the error message should be more frightening and descriptive?

pkg/storage/provider.go
```diff
@@ -35,31 +34,20 @@ type Provider struct {
...
-	Commission            defs.Commission
-	Services              wdk.Services
-	SynchronizeTxStatuses defs.SynchronizeTxStatuses
-}
-
 // NewGORMProvider creates a new storage provider with GORM repository.
-func NewGORMProvider(ctx context.Context, logger *slog.Logger, config GORMProviderConfig, opts ...ProviderOption) (*Provider, error) {
-	if err := config.FeeModel.Validate(); err != nil {
-		return nil, fmt.Errorf("invalid fee model: %w", err)
+func NewGORMProvider(ctx context.Context, logger *slog.Logger, chain defs.BSVNetwork, dbConfig defs.Database, services wdk.Services, opts ...ProviderOption) (*Provider, error) {
```
dorzepowski: After some rethinking IMHO we should try to stick to rule like this:
1. to inject optional (that we can create defaults for them) dependencies (services/components)  in opts
2. pass required dependencies (services/components) as arguments
3. if there are more then, let's say, three configuration (not services/components) arguments (except context) and at least one is optional - use config structure for all of them

WDYT about this?

That would mean that here:
1. all services (gormDB, funder, randomizer, beefVerifier, **logger**) should be passed as options
2. whole config should be passed as config structure as it was previously

pkg/internal/storage/repo/utxos.go
```diff
@@ -66,6 +66,17 @@ func (u *UTXOs) CountUTXOs(ctx context.Context, userID int, basketName string) (
...
 
+func (u *UTXOs) UnreserveUTXOsByTransactionID(ctx context.Context, transactionID uint) error {
+	err := u.db.WithContext(ctx).Model(&models.UserUTXO{}).
+		Where("reserved_by_id = ?", transactionID).
+		Update("reserved_by_id", nil).Error
+	if err != nil {
+		return fmt.Errorf("failed to unreserve UTXOs by transaction ID %d: %w", transactionID, err)
+	}
+
+	return nil
```
dorzepowski: maybe use gen query?
```suggestion
	table := u.query.UserUTXO
	_, err := table.WithContext(ctx).
		Where(table.ReservedByID.Eq(transactionID)).
		Update(table.ReservedByID, nil)
	if err != nil {
		return fmt.Errorf("failed to unreserve UTXOs by transaction ID %d: %w", transactionID, err)
	}

	return nil
```

pkg/internal/storage/repo/transactions.go
```diff
@@ -331,50 +317,14 @@ func (txs *Transactions) UpdateTransactionStatusByTxID(ctx context.Context, txID
...
-}
-
-func (txs *Transactions) releaseOutputsReservedByTransaction(tx *gorm.DB, transactionID uint) error {
-	return tx.Model(&models.Output{}).
-		Where("spent_by = ?", transactionID).
-		Updates(map[string]any{
-			"spent_by":  nil,
-			"spendable": true,
+			"status": txStatus,
 		}).Error
```
dorzepowski: Maybe use genquery
```go
	table := txs.query.Transaction

	_, err := table.WithContext(ctx).
		Where(table.ID.Eq(transactionID)).
		Update(table.Status, txStatus)

	if err != nil {
		return fmt.Errorf("update query for transaction status failed: %w", err)
	}
	return nil
```

pkg/internal/storage/repo/transactions.go
```diff
@@ -520,29 +470,21 @@ func (txs *Transactions) labelFilterScope(tx *gorm.DB, userID int, filter entity
...
+	err := txs.db.WithContext(ctx).
+		Model(&models.Transaction{}).
+		Select(
+			txs.query.Transaction.ID.ColumnName().String(),
+		).
+		Scopes(scopes.FromQueryOpts(opts)...).
+		Where("status IN ? ", txStatus).
+		Find(&rows).Error
+	if err != nil {
+		return nil, fmt.Errorf("failed to find transactions by statuses: %w", err)
```
dorzepowski: maybe use genquery
```suggestion
	table := &txs.query.Transaction
	rows, err := table.WithContext(ctx).
		Select(table.ID).
		Scopes(scopes.FromQueryOptsForGen(table, opts)...).
		Where(table.Status.In(slices.Map(txStatus, func(txStatus wdk.TxStatus) string {return string(txStatus)})...)).
		Find()
	if err != nil {
		return nil, fmt.Errorf("query for finding transaction ids by statuses failed: %w", err)
```

pkg/internal/storage/repo/outputs.go
```diff
@@ -359,68 +360,110 @@ func (o *Outputs) SaveOutputs(ctx context.Context, outputs []*entity.Output) err
 	return nil
 }
 
-func (o *Outputs) MakeOutputsSpendable(ctx context.Context, txID string, utxoStatus wdk.UTXOStatus) error {
+func (o *Outputs) MakeOutputsSpendableForTxID(ctx context.Context, txID string) error {
```
dorzepowski, line 364: As you already recreating all of this, maybe use genquery instead of plain gorm with getting some parts from genquery?

pkg/infra/gorm_provider_options.go
mgosek-4chain: @dorzepowski Could you please remind us (i.e. me and reviewers ^^) why it was hard to use? 

Minor: I'm wondering, is it really worth creating a separate function? It looks like this can be done more straightforwardly by passing the slice variable to `storage.NewGORMProvider`.

pkg/storage/internal/actions/abort_abandoned.go
mgosek-4chain: In the context of logging, I'd like to suggest creating a common logger and then, based on its structure, creating derived types as needed with extended attributes list. This approach will help avoid setting shared fields, such as 'transaction ID' etc. for each logging message.

mgosek-4chain (approved):

pkg/infra/gorm_provider_options.go
chris-4chain: In SPV-wallet it works differently, because the main issue was we didn't have access to the config struct after the "client" had been created. 

Here there is a convenient method that transforms `infra.Config` to provider options - to streamline the process of instantiating the GORM Storage Provider. 

This is used in several places that's why I decided to make a dedicated function to not repeat myself.

pkg/storage/internal/actions/abort_abandoned.go
chris-4chain: @dorzepowski  Ok, I changed to `attrs`. 
@mgosek-4chain I do use common logger with common fields. Here are fields that are not relevant for every log though. So they need to be defined explicitly.

pkg/storage/internal/actions/abort_abandoned.go
chris-4chain: I made this log message "more frightening"

pkg/storage/provider.go
chris-4chain: I adjusted the code according to our conversation. 
All but two (network & services) have been moved to options because we can create default values for them.

pkg/internal/storage/repo/outputs.go
chris-4chain: It wasn't easy, but I managed to used this approach. The code looks simpler now so, I guess, I was worth it.

pkg/infra/gorm_provider_options.go
dorzepowski: Yes, after some talks we agreed that main issue was that, the engine was created with options, which affects creation of some subservices/components, but there wasn't any config structure inside engine that would reflect effect of opts applying.

I don't remember exactly the problem but after some thoughts the approach was different then what is here.

dorzepowski (approved):

PR #507, opened 2025-08-25
Feat: complex examples create faucet server

### LINKED ISSUE #493: [FEATURE] Complex Wallet Example : Create Faucet Server
## Desired Solution

## Suggested Implementation

## Alternatives Considered

## Additional Context

## Categorize Feature

cain4chain:
Closes #493 

This pull request provides a complex example using the wallet-toolbox by creating a faucet server. This includes the following endpoints and values:

##API Endpoints
GET /info- Returns the faucet's address and current balance information including status, address, balance in satoshis, and network.
POST /faucet - Creates and broadcasts a faucet payment. Accepts address and amount parameters as a list, returns status, message, transaction ID, and BEEF hex data.
POST /topup - Internalizes a transaction output to the faucet wallet using outpoint format. Accepts outpoint parameter in txid:outputIndex format, returns status and message.

##Test Clients
test_address - Retrieves faucet address and balance with formatted output
test_faucet - Makes single faucet funding requests with detailed response logging
test_topup - Internalizes transactions using outpoint format with validation
test_faucet_periodically - Runs continuous faucet calls at configurable intervals for load testing
test_faucet_max_outputs - Runs a loop on the number of outputs provided in a faucet request to a specified address and amount
test_multi_output - Setup with multiple output template for faucet endpoint

##Configuration
The server requires the following environment variables:
FAUCET_PRIVATE_KEY - Private key for the faucet wallet funding operations
NETWORK - BSV network configuration (test/main)
PORT - Server port for HTTP API (default: 8080)
SERVER_URL - For API access

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter


examples/complex_wallet_examples/create_faucet_server/internal/handlers/address.go
```diff
@@ -0,0 +1,61 @@
...
+
+// NewGetAddressHandler returns faucet address and its current balance.
+func NewGetAddressHandler(deps AddressDeps) fiber.Handler {
+	return func(c *fiber.Ctx) error {
+		addr, err := methods.DeriveAddress(deps.FaucetKeyHex, deps.Network)
+		if err != nil {
+			return c.Status(http.StatusBadRequest).JSON(AddressResponse{Status: "error", Message: err.Error()})
+		}
+
+		ctx := context.Background()
```
dorzepowski: Use c.Context() instead of Background

examples/complex_wallet_examples/create_faucet_server/cmd/main.go
```diff
@@ -0,0 +1,52 @@
...
+			context.Background(),
+			infra.WithEnvPrefix(""),
+		)
+		if err != nil {
+			log.Fatalf("infra init failed: %v", err)
+		}
+		if err := srv.ListenAndServe(); err != nil {
+			log.Fatalf("infra server exited: %v", err)
+		}
+	}()
```
dorzepowski: use storage Gorm provider instead.
There is no need to start another server.

examples/complex_wallet_examples/create_faucet_server/internal/config/config.go
```diff
@@ -0,0 +1,85 @@
...
+)
+
+type Config struct {
+	Port int `mapstructure:"port"`
+
+	// Faucet configuration
+	ServerPrivateKey string          `mapstructure:"server_private_key"`
+	FaucetPrivateKey string          `mapstructure:"faucet_private_key"`
+	Network          defs.BSVNetwork `mapstructure:"network"`
+	ServerURL        string          `mapstructure:"server_url"` // URL of wallet toolbox storage server
```
dorzepowski: I would start without this part (just use local version of storage)

examples/complex_wallet_examples/create_faucet_server/internal/config/config.go
```diff
@@ -0,0 +1,85 @@
...
+
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/defs"
+)
+
+type Config struct {
+	Port int `mapstructure:"port"`
+
+	// Faucet configuration
+	ServerPrivateKey string          `mapstructure:"server_private_key"`
+	FaucetPrivateKey string          `mapstructure:"faucet_private_key"`
```
dorzepowski: What is the difference between those keys?
One key should be enough, we need only private key for faucet.

examples/complex_wallet_examples/create_faucet_server/internal/config/config.go
```diff
@@ -0,0 +1,85 @@
...
+		cfg.FaucetPrivateKey = faucetKey
+	}
+
+	if network := os.Getenv("NETWORK"); network != "" {
+		cfg.Network = defs.BSVNetwork(network)
+	}
+
+	if serverURL := os.Getenv("SERVER_URL"); serverURL != "" {
+		cfg.ServerURL = serverURL
+	}
```
dorzepowski: use constants for env names

examples/complex_wallet_examples/create_faucet_server/internal/config/config.go
```diff
@@ -0,0 +1,85 @@
...
+		return fmt.Errorf("port must be between 1 and 65535")
+	}
+
+	if c.ServerURL == "" {
+		return fmt.Errorf("server_url is required")
+	}
+
+	if c.Network, err = defs.ParseBSVNetworkStr(string(c.Network)); err != nil {
+		return fmt.Errorf("invalid network: %w", err)
+	}
```
dorzepowski: Add info in error how to set those param (what env variable to use), so it's easier from deployer perspective.

examples/complex_wallet_examples/create_faucet_server/internal/handlers/faucet.go
```diff
@@ -0,0 +1,61 @@
...
+				return c.Status(http.StatusBadRequest).JSON(FaucetResponse{Status: "error", Message: "address is required for all outputs"})
+			}
+			if output.Amount == 0 {
+				return c.Status(http.StatusBadRequest).JSON(FaucetResponse{Status: "error", Message: "amount must be greater than 0 for all outputs"})
+			}
+			totalAmount += output.Amount
+		}
+
+		// Check total amount limit
+		if totalAmount > constants.MaxFaucetTotalAmount {
```
dorzepowski: As I said in our discussion, Max faucet top up amount should be configurable and optional (so it should be unlimited by default)

examples/complex_wallet_examples/create_faucet_server/internal/handlers/topup.go
```diff
@@ -0,0 +1,73 @@
...
+
+		// Parse outpoint into txid and output index
+		parts := strings.Split(req.Outpoint, ":")
+		if len(parts) != 2 {
+			return c.Status(http.StatusBadRequest).JSON(TopUpResponse{Status: "error", Message: "invalid outpoint format, expected txid:outputIndex"})
+		}
+
+		txid := parts[0]
+		outputIndexStr := parts[1]
+		outputIndex, err := strconv.ParseUint(outputIndexStr, 10, 32)
```
dorzepowski: wouldn't it be simpler to use object with txid and index fields (or simply transaction.Outpoint) as a field in request? 


If not, then maybe use `transaction.OutpointFromString` as this code is already in go-sdk

examples/complex_wallet_examples/create_faucet_server/internal/handlers/address.go
```diff
@@ -0,0 +1,61 @@
...
+		storageClient, cleanup, err := storage.NewClient(deps.ServerURL)
+		if err != nil {
+			return c.Status(http.StatusInternalServerError).JSON(AddressResponse{Status: "error", Message: err.Error()})
+		}
+		defer cleanup()
+
+		w, err := wallet.New(deps.Network, priv, storageClient)
+		if err != nil {
+			return c.Status(http.StatusInternalServerError).JSON(AddressResponse{Status: "error", Message: err.Error()})
+		}
```
dorzepowski: The wallet (with storage and private key) should be created once at the startup of the faucet, and injected as dependency.

This comment applies to all handlers.

examples/complex_wallet_examples/create_faucet_server/internal/methods/address.go
```diff
@@ -0,0 +1,85 @@
...
+
+const (
+	balanceListLimit = uint32(100)
+)
+
+// DeriveAddress returns a faucet BRC-29 address
+func DeriveAddress(faucetKeyHex string, network defs.BSVNetwork) (string, error) {
+	priv, err := ec.PrivateKeyFromHex(faucetKeyHex)
+	if err != nil {
+		return "", err
```
dorzepowski: please wrap the errors, as it simplifies checking where was the source of the error

examples/complex_wallet_examples/create_faucet_server/internal/methods/address.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/go-softwarelab/common/pkg/to"
+)
+
+const (
+	balanceListLimit = uint32(100)
+)
+
+// DeriveAddress returns a faucet BRC-29 address
+func DeriveAddress(faucetKeyHex string, network defs.BSVNetwork) (string, error) {
+	priv, err := ec.PrivateKeyFromHex(faucetKeyHex)
```
dorzepowski: you often build that key.
It should be then created at the start of faucet and passed as a dependency to handlers 
(this will make the faucet fail at start if the key is invalid, and you won't need to make a conversion here and handle error)

examples/complex_wallet_examples/create_faucet_server/internal/methods/address.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+	derivationSuffixBytes, err := utils.BytesFromBase64(constants.DefaultBase64Suffix)
+	if err != nil {
+		return "", err
+	}
+
+	keyID := brc29.KeyID{
+		DerivationPrefix: base64.StdEncoding.EncodeToString(derivationPrefixBytes),
+		DerivationSuffix: base64.StdEncoding.EncodeToString(derivationSuffixBytes),
+	}
```
dorzepowski: this is confusing. As prefix and suffix aren't configurable - becuase they are in constants.
I would simply assign those constants to KeyID fields.

examples/complex_wallet_examples/create_faucet_server/internal/constants/constants.go
```diff
@@ -0,0 +1,8 @@
+package constants
+
+const (
+	DefaultBase64Prefix = "SfKxPIJNgdI="
+	DefaultBase64Suffix = "NaGLC6fMH50="
```
dorzepowski:
```suggestion
	FaucetAddressKeyIDPrefix = "SfKxPIJNgdI="
	FaucetAddressKeyIDSuffix = "NaGLC6fMH50="
```

examples/complex_wallet_examples/create_faucet_server/internal/methods/faucet.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/bsv-blockchain/go-sdk/script"
+	"github.com/bsv-blockchain/go-sdk/transaction/template/p2pkh"
+	sdk "github.com/bsv-blockchain/go-sdk/wallet"
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/defs"
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/storage"
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/wallet"
+	"github.com/go-softwarelab/common/pkg/to"
+)
+
+type FaucetDeps struct {
```
dorzepowski: this seems to be used in all handlers.
Maybe move it to separate file

examples/complex_wallet_examples/create_faucet_server/internal/methods/faucet.go
```diff
@@ -0,0 +1,95 @@
...
+	Address string `json:"address"`
+	Amount  uint64 `json:"amount"`
+}
+
+// FundAddress creates and broadcasts a faucet payment with one or more outputs.
+// Returns the txid string and full Atomic BEEF hex on success.
+func FundAddress(ctx context.Context, deps FaucetDeps, outputs ...FaucetOutput) (string, string, error) {
+	if deps.FaucetKeyHex == "" {
+		return "", "", fmt.Errorf("faucet key not configured")
+	}
```
dorzepowski: we can assume it's configured correctly, as it should be checked at the start of application. And we want that faucet to be fairly simple (and less code == simpler)

examples/complex_wallet_examples/create_faucet_server/internal/methods/faucet.go
```diff
@@ -0,0 +1,95 @@
...
+	if err != nil {
+		return "", "", fmt.Errorf("create wallet: %w", err)
+	}
+
+	// Create outputs for each address and amount
+	createOutputs := make([]sdk.CreateActionOutput, len(outputs))
+	for i, output := range outputs {
+		addr, err := script.NewAddressFromString(output.Address)
+		if err != nil {
+			return "", "", fmt.Errorf("invalid address %s: %w", output.Address, err)
```
dorzepowski: add information about index to error message

examples/complex_wallet_examples/create_faucet_server/internal/methods/topup.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/bsv-blockchain/go-sdk/transaction/template/p2pkh"
+	sdk "github.com/bsv-blockchain/go-sdk/wallet"
+	"github.com/bsv-blockchain/go-wallet-toolbox-faucet-server/internal/constants"
+	"github.com/bsv-blockchain/go-wallet-toolbox-faucet-server/internal/utils"
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/defs"
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/services"
+)
+
+// TopUpInternalize validates tx belongs to faucet and internalizes it.
+func TopUpInternalize(ctx context.Context, deps FaucetDeps, _ *ec.PublicKey, w sdk.Interface, txid string, outputIndex uint32) error {
```
dorzepowski: remove unused argument `_ *ec.PublicKey`

examples/complex_wallet_examples/create_faucet_server/internal/methods/topup.go
```diff
@@ -0,0 +1,97 @@
...
+
+	derivationSuffixBytes, err := utils.BytesFromBase64(constants.DefaultBase64Suffix)
+	if err != nil {
+		return fmt.Errorf("failed to convert derivation suffix from base64: %w", err)
+	}
+
+	faucetPriv, err := ec.PrivateKeyFromHex(deps.FaucetKeyHex)
+	if err != nil {
+		return fmt.Errorf("failed to parse faucet private key: %w", err)
+	}
```
dorzepowski: here again you need to create a key, so it would be better to have it in dependencies.

examples/complex_wallet_examples/create_faucet_server/internal/methods/address.go
```diff
@@ -0,0 +1,85 @@
...
+
+	var addr *script.Address
+	if network == defs.NetworkMainnet {
+		addr, err = brc29.Address(priv, keyID, identityKey, brc29.WithMainNet())
+	} else {
+		addr, err = brc29.Address(priv, keyID, identityKey, brc29.WithTestNet())
```
dorzepowski: This is minor thing, but to be aligned with BA practices we should rather use AnyoneKey as sender

examples/complex_wallet_examples/create_faucet_server/internal/server/server.go
```diff
@@ -0,0 +1,30 @@
...
+func New(cfg config.Config) *Server {
+	app := fiber.New()
+
+	deps := handlers.FaucetDeps{
+		FaucetKeyHex: cfg.FaucetPrivateKey,
+		Network:      cfg.Network,
+		ServerURL:    cfg.ServerURL,
+	}
+	app.Post("/faucet", handlers.NewFaucetHandler(deps))
+	app.Get("/address", handlers.NewGetAddressHandler(deps))
```
dorzepowski: ### just thinking
maybe `/info` would be better name, and we could simply extend this endpoint with more status information in the future if needed.

examples/complex_wallet_examples/create_faucet_server/internal/utils/utils.go
```diff
@@ -0,0 +1,14 @@
+package utils
+
+import (
+	"encoding/base64"
+	"fmt"
+)
+
+func BytesFromBase64(s string) ([]byte, error) {
```
dorzepowski: looks like this doesn't give a big benefit as you still should wrap the error at the caller side. I would rather inline it 

(especially as it should be call in single place - where the internalize is done

examples/complex_wallet_examples/create_faucet_server/test_examples/test_faucet_multi_output/main.go
```diff
@@ -0,0 +1,86 @@
...
+
+type faucetReq struct {
+	Outputs []methods.FaucetOutput `json:"outputs"`
+}
+
+type faucetResp struct {
+	Status  string `json:"status"`
+	Message string `json:"message"`
+	Txid    string `json:"txid"`
+	BEEFHex string `json:"beef_hex"`
```
dorzepowski, line 21: ### just thinking
shouldn't we also get an outpoint index in response?

examples/complex_wallet_examples/create_faucet_server/cmd/main.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+	if err := cfg.Validate(); err != nil {
+		log.Fatalf("invalid config: %v", err)
+	}
+
+	privKey, err := ec.NewPrivateKey()
+	if err != nil {
+		log.Fatalf("failed to generate ephemeral server key: %v", err)
+	}
+	_ = hex.EncodeToString(privKey.Serialize())
```
dorzepowski: you don't use it so please remove this code

examples/complex_wallet_examples/create_faucet_server/internal/handlers/faucet.go
```diff
@@ -0,0 +1,59 @@
...
+		totalAmount := uint64(0)
+		for _, output := range req.Outputs {
+			if output.Address == "" {
+				return c.Status(http.StatusBadRequest).JSON(FaucetResponse{Status: "error", Message: "address is required for all outputs"})
+			}
+			if output.Amount == 0 {
+				return c.Status(http.StatusBadRequest).JSON(FaucetResponse{Status: "error", Message: "amount must be greater than 0 for all outputs"})
+			}
+			totalAmount += output.Amount
+		}
```
dorzepowski: ### minor

```suggestion
		for i, output := range req.Outputs {
			if output.Address == "" {
				return c.Status(http.StatusBadRequest).JSON(FaucetResponse{Status: "error", Message: "address is missing in output " + to.StringFromInteger(i)})
			}
			if output.Amount == 0 {
				return c.Status(http.StatusBadRequest).JSON(FaucetResponse{Status: "error", Message: "amount equal to zero is not allowed in output " + to.StringFromInteger(i)})
			}
			totalAmount += output.Amount
		}
```

examples/complex_wallet_examples/create_faucet_server/internal/methods/faucet.go
dorzepowski: Still I'm suggesting to remove this check

examples/complex_wallet_examples/create_faucet_server/internal/methods/faucet.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+	if len(outputs) == 0 {
+		return "", "", fmt.Errorf("at least one output is required")
+	}
+
+	if deps.Storage == nil {
+		return "", "", fmt.Errorf("storage provider not configured")
+	}
+
+	faucetPriv, err := ec.PrivateKeyFromHex(deps.FaucetKeyHex)
```
dorzepowski: I'm suggesting (again) to add parsed private key to deps (especially as you're already parsing it in server.go)

examples/complex_wallet_examples/create_faucet_server/internal/methods/faucet.go
```diff
@@ -0,0 +1,98 @@
...
+	if err != nil {
+		return "", "", fmt.Errorf("invalid faucet key: %w", err)
+	}
+
+	faucetWallet := deps.Wallet
```
dorzepowski: I would believe our deps, and not check for null values of them. It's already checked in server when deps are build, and this example should have as little code as possible (but still in good quality and fairly reliable)

examples/complex_wallet_examples/create_faucet_server/internal/utils/utils.go
dorzepowski: Please remove unused function

examples/complex_wallet_examples/create_faucet_server/internal/methods/address.go
dorzepowski: 🆙

examples/complex_wallet_examples/create_faucet_server/internal/methods/topup.go
```diff
@@ -0,0 +1,93 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to decode derivation prefix: %w", err)
+	}
+	derivationSuffixBytes, err := base64.StdEncoding.DecodeString(constants.FaucetAddressKeyIDSuffix)
+	if err != nil {
+		return fmt.Errorf("failed to decode derivation suffix: %w", err)
+	}
+
+	// Sender: AnyoneKey to align with BA practices
+	_, senderPub := sdk.AnyoneKey()
```
dorzepowski: You changed it here and didn't change it in DeriveAddress

examples/complex_wallet_examples/create_faucet_server/internal/handlers/topup.go
```diff
@@ -0,0 +1,68 @@
...
+
+		// Parse outpoint using transaction helper
+		op, err := transaction.OutpointFromString(req.Outpoint)
+		if err != nil {
+			return c.Status(http.StatusBadRequest).JSON(TopUpResponse{Status: "error", Message: fmt.Sprintf("invalid outpoint: %v", err)})
+		}
+
+		// Use injected wallet if available
+		w := deps.Wallet
+		if w == nil {
```
dorzepowski: Again, I would trust the deps, that it contains non nil wallet

examples/complex_wallet_examples/create_faucet_server/internal/handlers/topup.go
```diff
@@ -0,0 +1,68 @@
...
+			if err != nil {
+				return c.Status(http.StatusBadRequest).JSON(TopUpResponse{Status: "error", Message: err.Error()})
+			}
+			w, err = wallet.New(deps.Network, priv, deps.Storage)
+			if err != nil {
+				return c.Status(http.StatusInternalServerError).JSON(TopUpResponse{Status: "error", Message: err.Error()})
+			}
+		}
+
+		if err := methods.TopUpInternalize(context.Background(), deps, w, op.Txid.String(), uint32(op.Index)); err != nil {
```
dorzepowski: no need for casting op.Index to uint32 as it is already in this type

dorzepowski: use c.Context() instead of context.Background()

examples/complex_wallet_examples/create_faucet_server/internal/handlers/faucet.go
```diff
@@ -0,0 +1,59 @@
...
+
+		if deps.MaxFaucetTotalAmount > 0 && totalAmount > deps.MaxFaucetTotalAmount {
+			return c.Status(http.StatusBadRequest).JSON(FaucetResponse{Status: "error", Message: fmt.Sprintf("total amount must be <= %d satoshis", deps.MaxFaucetTotalAmount)})
+		}
+
+		txid, beefHex, err := methods.FundAddress(context.Background(), deps, req.Outputs...)
```
dorzepowski: use c.Context instead of context.Background

examples/complex_wallet_examples/create_faucet_server/internal/handlers/address.go
```diff
@@ -0,0 +1,54 @@
...
+
+// NewGetAddressHandler returns faucet address and its current balance.
+func NewGetAddressHandler(deps AddressDeps) fiber.Handler {
+	return func(c *fiber.Ctx) error {
+		addr, err := methods.DeriveAddress(deps.FaucetKeyHex, deps.Network)
+		if err != nil {
+			return c.Status(http.StatusBadRequest).JSON(AddressResponse{Status: "error", Message: err.Error()})
+		}
+
+		w := deps.Wallet
```
dorzepowski: and again, trust the deps to contain the wallet

examples/complex_wallet_examples/create_faucet_server/test_examples/test_faucet_multi_output/main.go
cain4chain: We can add this, however not sure how we would have the response when its multiple outputs requested.
Plus thinking if we are providing with multiple outputs having the output index may not be enough, we may also require the amount and address as well so that each one is identified if wanting to process it "blindly" (being the assumption is the order is as per the request provided).

My thought is we can provide the outputIndex, address and amount as list format - catch is depending on the amount of outputs requested this could be a large amount of data 🤔

chris-4chain (approved):

dorzepowski (approved):

PR #518, opened 2025-08-27
[FEATURE] CRUD for Transactions entity

### LINKED ISSUE #504: [FEATURE] CRUD for Transaction entity
This task should focus on Read capability.

ac4ch:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter


mgosek-4chain (approved):

pkg/internal/storage/entity/transaction.go
```diff
@@ -3,9 +3,11 @@ package entity
 import (
 	"time"
 
+	coreentity "github.com/bsv-blockchain/go-wallet-toolbox/pkg/entity"
 	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/wdk"
 )
 
+// Transaction represents a blockchain transaction recorded in the system.
```
chris-4chain: This is internal package and it doesn't need a doc-comment.
Add doc-comment only when it is actually reasonable. IN this case I would remove this other comments you've added.

pkg/internal/storage/entity/transaction.go
```diff
@@ -23,3 +25,26 @@ type Transaction struct {
 	InputBEEF   []byte
 	Labels      []string
 }
+
+// TransactionReadSpecification defines filter criteria for querying transactions.
+type TransactionReadSpecification struct {
```
chris-4chain: This specification should be rather in `pkg/entity` (non-internal) entity package.

I should be consistent to how we made similar specifications for other entities

pkg/internal/storage/entity/transaction.go
```diff
@@ -23,3 +25,26 @@ type Transaction struct {
...
+	ID                  *uint
+	UserID              *int
+	Status              *wdk.TxStatus
```
chris-4chain: For knownTx you used `Comparable` generic type for "enum" types. Here is just a status. 

Please make it consistent. 
To me, it should the Comparable should be used here also.

pkg/internal/storage/entity/transaction.go
```diff
@@ -23,3 +25,26 @@ type Transaction struct {
...
+	Reference           *string
+	IsOutgoing          *bool
```
chris-4chain: And for other types... Why you haven't used Comparable type as you did for KnownTx?

pkg/storage/crud/transaction.go
```diff
@@ -0,0 +1,166 @@
...
+	t.spec.ID = to.Ptr(id)
+	return t
+}
+
+func (t *transaction) UserID(userID int) TransactionReader {
+	t.spec.UserID = to.Ptr(userID)
+	return t
+}
+
+func (t *transaction) Status(status wdk.TxStatus) TransactionReader {
```
chris-4chain: You completely skipped the way we defined "comparable" fields for other entities.

pkg/storage/crud/transaction.go
```diff
@@ -0,0 +1,166 @@
...
+	return &transaction{
+		repo: repo,
+	}
+}
+
+func (t *transaction) Read() TransactionReader {
+	return t
+}
+
+func (t *transaction) Create(ctx context.Context, tx *entity.Transaction) error {
```
chris-4chain, line 67: The Transaction entity needs to be moved to the "public" `entity` package: `pkg/entity`. 

Now, it is not usable by the adopters, because they cannot access the Transaction struct (because it is internal. )

chris-4chain (changes requested):

pkg/storage/crud/transaction.go
ac4ch: it was moved

pkg/storage/crud/transaction.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+	return count, nil
+}
+
+func (t *transaction) ID(id uint) TransactionReader {
+	t.spec.ID = to.Ptr(id)
+	return t
+}
+
+func (t *transaction) UserID(userID int) NumericCondition[TransactionReader, int] {
```
chris-4chain, line 110: You forgot to remove the unnecessary arguments. After the change you need to adjust unit tests. 
image

chris-4chain (changes requested):

pkg/storage/provider_transaction_enity_test.go
```diff
@@ -0,0 +1,206 @@
...
+	err := activeStorage.TransactionEntity().Update(t.Context(), &pkgentity.TransactionUpdateSpecification{
+		ID:     1,
+		Status: &newStatus,
+	})
+
+	// then:
+	require.NoError(t, err)
+
+	// when:
+	count, err := activeStorage.TransactionEntity().Read().ID(1).Status().Equals(newStatus).Count(t.Context())
```
chris-4chain: When you call `ID` method the rest of the filters will be ignored. See how you implemented `repos/transaction/conditionsBySpec` (you made it correctly - consistent to other entities)

But this reveals another issue. The interface design should be made in a way that when you call `ID` other filters should not be accessible. 
- See how `knownTx` or `commission` crud is made. 

So, here, I would search only by status. You can always check the count before and after the update. Or... you can get by ID and check the value of the field from the returned entity.

pkg/storage/crud/transaction.go
```diff
@@ -0,0 +1,189 @@
...
+type TransactionReadOperations interface {
+	Find(ctx context.Context) ([]*entity.Transaction, error)
+	Count(ctx context.Context) (int64, error)
+}
+
+// TransactionReader enables fluent query building.
+type TransactionReader interface {
+	TransactionReadOperations
+
+	ID(id uint) TransactionReader
```
chris-4chain: Here:
```suggestion
	ID(id uint) TransactionReadOperations
```

chris-4chain (changes requested):

chris-4chain (approved):

PR #520, opened 2025-08-27
feat: provide the functions in brc29 package to generate Address for recipient from sender side and or himself

dorzepowski:
This is proposition to have 2 functions for address creation.
One `AddressForSelf` that recipient can use to prepare an address and pass it to sender.
Second `AddressForCounterparty` to be used by sender to create an address to send funds to recipient.

### DIFF
--- a/examples/internal/example_setup/faucet_address.go
+++ b/examples/internal/example_setup/faucet_address.go
@@ -18,10 +18,10 @@ func FaucetAddress(wallet *Setup) {
 		DerivationSuffix: base64.StdEncoding.EncodeToString(parts.DerivationSuffix),
 	}
 
-	address, err := brc29.Address(
-		wallet.PrivateKey,
+	address, err := brc29.MyAddress(
+		parts.SenderIdentityKey,
 		keyID,
-		wallet.IdentityKey,
+		wallet.PrivateKey,
 		brc29.WithTestNet(),
 	)
 	if err != nil {
--- a/examples/internal/example_setup/faucet_address.go
+++ b/examples/internal/example_setup/faucet_address.go
@@ -18,7 +18,7 @@ func FaucetAddress(wallet *Setup) {
 		DerivationSuffix: base64.StdEncoding.EncodeToString(parts.DerivationSuffix),
 	}
 
-	address, err := brc29.MyAddress(
+	address, err := brc29.AddressForSelf(
 		parts.SenderIdentityKey,
 		keyID,
 		wallet.PrivateKey,
--- a/manual_tests/internal/tui/internalize_form.go
+++ b/manual_tests/internal/tui/internalize_form.go
@@ -240,7 +240,7 @@ func calculateAddressForInternalize(derivationPrefix, derivationSuffix string, u
 	}
 
 	networkOption := to.IfThen(bsvNetwork == defs.NetworkMainnet, brc29.WithMainNet()).ElseThen(brc29.WithTestNet())
-	address, err := brc29.YourAddress(anyonePriv, keyID, user.PublicKey(), networkOption)
+	address, err := brc29.AddressForCounterparty(anyonePriv, keyID, user.PublicKey(), networkOption)
 	if err != nil {
 		return "", fmt.Errorf("failed to calculate address: %w", err)
 	}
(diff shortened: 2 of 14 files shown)


manual_tests/internal/tui/internalize_form.go
```diff
@@ -240,7 +240,7 @@ func calculateAddressForInternalize(derivationPrefix, derivationSuffix string, u
 	}
 
 	networkOption := to.IfThen(bsvNetwork == defs.NetworkMainnet, brc29.WithMainNet()).ElseThen(brc29.WithTestNet())
-	address, err := brc29.Address(anyonePriv, keyID, user.PublicKey(), networkOption)
+	address, err := brc29.AddressForRecipient(anyonePriv, keyID, user.PublicKey(), networkOption)
```
chris-4chain: In this case the user wants to generate an address to obtain funds from faucet. So, I suppose, the user is Recipient. But, According to the description of `AddressForRecipient` this function is meant to be used by "sender". So, I'm confused 🤔😅

dorzepowski: My thought about the name for this function was:
I am a sender and I need to calculate an address for recipient.

I don't have currently better idea for this name. Any suggestions are welcome.

dorzepowski: hmmm... maybe:
`RecipientAddress` would be better name?

cain4chain: Maybe we should change the name of the `Address `method to `AddressForSelf`
Then keep the name for the other as `AddressForRecipient`

Maybe it's important for the key word `Address ` to be the first word of the method naming in this scenario.
WDYT?

examples/internal/example_setup/faucet_address.go
```diff
@@ -18,7 +18,7 @@ func FaucetAddress(wallet *Setup) {
 		DerivationSuffix: base64.StdEncoding.EncodeToString(parts.DerivationSuffix),
 	}
 
-	address, err := brc29.Address(
+	address, err := brc29.AddressForRecipient(
```
cain4chain: In this example we are able to use the anyoneKey for the private key value. The update is part of these changes here.
https://github.com/bsv-blockchain/go-wallet-toolbox/blob/f6eab2d446071ecfee2594a7ac655fe22568e1d4/examples/internal/example_setup/faucet_address.go#L24

So in this case we can still do the same but I will use the AddressForRecipient method and the anyonePub key as third param (instead of anyonePriv) and pass in the walletPrivate key as first param.

No changes required in this PR, I will update my PR once this one is merged :D

UPDATE : sorry, I thought the new method was for self generation, but since this example is generating an address for self then I might update it later to use the `Address` method instead for clarity.

cain4chain (approved):

pkg/brc29/brc29_template.go
```diff
@@ -17,7 +16,7 @@ import (
 //   - recipient: the recipient key. This is the public key from private key that will be able to unlock it later. Can be a public key hex or a key deriver or ec.PublicKey.
 //   - opts: additional options.
 func Lock[SenderKey CounterpartyPrivateKey, RecipientKey CounterpartyPublicKey](senderPrivateKeySource SenderKey, keyID KeyID, recipientPublicKeySource RecipientKey, opts ...func(*lockOptions)) (*script.Script, error) {
```
cain4chain, line 40: Should we consider a method `LockForSelf` in this case? 
Use case could be related to a requirement to generate an identical locking script based on some previous transaction output. WDYT?
Can always use the single additional `p2pkh.Lock` method to get the locking script so it might be overkill?

dorzepowski: I would postpone creating it, util we would really have situation where it is needed.
I don't really know, what is the case when the recipient would need to an identical locking script. and if he have previous transaction, then it's simpler to copy the script then regenerating it.

examples/internal/example_setup/faucet_address.go
dorzepowski: Another version is here, and I hope it will dispel all doubts

manual_tests/internal/tui/internalize_form.go
dorzepowski: Another version is here, and I hope it will dispel all doubts

pkg/brc29/brc29_template.go
cain4chain: yes understandable. It's more an edge case method.
I guess one use case is wanted to generate the locking script using some predefined payment remittance values. Maybe in the case of validating a locking script derivation matches a specified transaction output. 

Theoretically, thinking if there was a dispute on address derivation or payment then the user wallet side could generate to check that it is actually their own address. This is more a tx mitigation helper method than anything.

chris-4chain (approved):

PR #522, opened 2025-08-28
fix: UTXO management for noSend scenario 

chris-4chain:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter


pkg/storage/provider_no_send_test.go
```diff
@@ -3,162 +3,378 @@ package storage_test
...
+			ContainsTxsWithStatus(t, wdk.SendWithResultStatusUnproven, givenNoSend.NoSendTxs()...)
+
+		testabilities.
+			ThenDBState(t, activeStorage).
+			HasUserTransactionsByTxIDsWithStatus(testusers.Alice, wdk.TxStatusUnproven, givenNoSend.NoSendTxs()...)
+
+		testabilities.ThenFunds(t, testusers.Alice, activeStorage).
+			ShouldBeAbleToReserveSatoshis(
+				inputSatoshis +
+					-4 + // two no-send txs (2 sats each)
```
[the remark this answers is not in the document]
chris-4chain: I provided comments. It should be enough to understand what the numbers mean.

mgosek-4chain (approved):

cain4chain (approved):

jakubmkowalski (approved):

pkg/storage/provider_no_send_test.go
```diff
@@ -3,162 +3,378 @@ package storage_test
...
+		// when:
+		var noSendChangeOutpoints []wdk.OutPoint
+		for i := range noSendChainCount {
+			t.Logf("Creating NoSend tx #%d", i+1)
+			noSendChangeOutpoints = givenNoSend.CreateAndProcessNoSendAction(noSendChangeOutpoints)
+		}
+
+		// and:
+		// Call processAction using sendWith and IsNewTx set to false, including the two previous transactions in SendWithSlice.
+		sendWithProcessAction := givenNoSend.ProcessAction(wdk.ProcessActionArgs{
```
dorzepowski: I know that it add a lot of work, but in case of those tests, readability is important to not loose the clue what we're testing.

Therefore I would suggest to:
1. create additional testability package (it could have other name to make it easier to find, like `nosendtest`)
2. make `nosendtest.New` to return `given, when, then := nosendtest.New(t)`
3. Given should have only methods for setup, like: Provider, FundWallet, Faucet
4. When should have those actions that you added to givenNoSend, so for example CreateAndProcessNoSendAction and ProcessAction

pkg/storage/provider_no_send_test.go
```diff
@@ -3,162 +3,378 @@ package storage_test
...
-		IsNewTx:    false,
-		IsNoSend:   false,
-		SendWith:   []primitives.HexString{primitives.HexString(result.FirstTxID), primitives.HexString(result.SecondTxID)},
-		IsSendWith: true,
+		// when:
+		var noSendChangeOutpoints []wdk.OutPoint
+		for i := range noSendChainCount {
+			t.Logf("Creating NoSend tx #%d", i+1)
+			noSendChangeOutpoints = givenNoSend.CreateAndProcessNoSendAction(noSendChangeOutpoints)
+		}
```
dorzepowski: I would skip the "automatisation here" and simply duplicate this line to explicitly show that we're making 2 no send transactions.

⚠️ The same suggestion applies to all other tests with such a loop

pkg/storage/provider_no_send_test.go
```diff
@@ -3,162 +3,378 @@ package storage_test
...
+		defer cleanup()
 
-	fixture := testabilities.NewBuildNoSendTransactionFixture(t, inputSatoshis)
-	defer fixture.Cleanup()
+		activeStorage := given.Provider().GORM()
+		givenNoSend := testabilities.GivenNoSend(t, given, activeStorage, testusers.Alice)
 
-	result := fixture.BuildNoSendTransaction()
+		// and:
+		givenNoSend.FundWallet(inputSatoshis) // This makes the wallet not-empty with several UTXOs
```
dorzepowski: Maybe
```suggestion
		givenNoSend.UserHasSeveralUTXOToSpend(inputSatoshis)
```

pkg/storage/provider_no_send_test.go
```diff
@@ -3,162 +3,378 @@ package storage_test
...
-		HasUserTransactionsByTxIDsWithStatus(testusers.Alice, wdk.TxStatusUnproven, result.FirstTxID, result.SecondTxID)
-}
+		activeStorage := given.Provider().GORM()
+		givenNoSend := testabilities.GivenNoSend(t, given, activeStorage, testusers.Alice)
 
-func TestNoSendSendWithScenario_SendWithNewTx(t *testing.T) {
-	// given:
-	const inputSatoshis = 99904
+		// and:
+		given.Faucet(activeStorage, testusers.Alice).TopUp(inputSatoshis) // Intentionally, create only one UTXO
```
dorzepowski: maybe new function to be explicit about intention?
```suggestion
		givenNoSend.UserHasSingleUTXOToSpend(inputSatoshis)
```

or:
```suggestion
		givenNoSend.UserReceivedSingleUTXO(inputSatoshis)
```

pkg/storage/provider_no_send_test.go
```diff
@@ -3,162 +3,378 @@ package storage_test
...
+		const initialUTXOSats = 10000
+		const noSendChainCount = 2
+
+		given, cleanup := testabilities.Given(t)
+		defer cleanup()
+
+		activeStorage := given.Provider().GORM()
+		givenNoSend := testabilities.GivenNoSend(t, given, activeStorage, testusers.Alice)
+
+		// and:
```
dorzepowski, line 107: maybe add more details to those comments, prepare dedicated methods in fixture (Given)


Here for example could be
```suggestion
		// and: user has 2 UTXO each one higher then the amount spend in all no send transactions
```

pkg/storage/provider_no_send_test.go
```diff
@@ -3,162 +3,378 @@ package storage_test
...
-	"github.com/go-softwarelab/common/pkg/slices"
-	"github.com/go-softwarelab/common/pkg/to"
 	"github.com/stretchr/testify/require"
 )
 
-func TestNoSendSendWithScenario_SendWithoutNewTx(t *testing.T) {
-	// given:
-	const inputSatoshis = 99904
+func TestNoSendPlusSendWithScenario(t *testing.T) {
+	t.Run("two no-send txs, many initial UTXOs, all noSendChange is used, no newTx when sendWith", func(t *testing.T) {
```
dorzepowski, line 14: Check this out, it looks a little bit weird, but maybe it's reading better 🤷 
Maybe this approach is worth to discuss on our slack channel.


```suggestion
	t.Run(`Alice has many UTXOs.
				She created two subsequent no-send transactions,
				and then send them in SendWith with not new transaction`,
		func(t *testing.T) {
```

And the result in console looks like that, IMO it's still quite readable:
```sh
=== RUN   TestNoSendPlusSendWithScenario/Alice_has_many_UTXOs._____She_created_two_subsequent_no-send_transactions,_____and_then_send_them_in_SendWith_with_not_new_transaction
```

pkg/storage/provider_no_send_test.go
dorzepowski, line 1: It's probably out of the scope of this PR,
BUT I think that we found another pretty big issue thanks to them
....
Every test in this file (or most of them) is taking 3-6 seconds to run.
Those are tests with mocked dependencies (WoC and ARC) 
So this mean that making let's say 3 transactions is taking 3 seconds without the impact of those services
This is pretty bad result 🙈 

Can you confirm my observations? And if so, can you create an issue to measure, find bottleneck and optimise this process.

chris-4chain: Yes - I even wanted to fix it by the way of this task but later I decided not to because it is not so easy to resolve.

pkg/storage/provider_no_send_test.go
chris-4chain: I would prefer describing details in a comment rather than in such complex test name

dorzepowski: you still have complex test name. But current name, doesn't mean much to me.

pkg/storage/provider_no_send_test.go
dorzepowski: It occurs that the problem was because of retries when calling the WoC and Bitails - so it should be fixed when we will prepare some more option to configure wallet services and we could disable broadcasting via WoC and Bitails in tests.

pkg/storage/provider_no_send_test.go
chris-4chain: Done 👍 Check it out now :)

pkg/storage/provider_no_send_test.go
chris-4chain: After refactor the code should be self explanatory enough 🤞

pkg/storage/provider_no_send_test.go
chris-4chain: The description is in code (self-explanatory) and additional comments. 
The test name shows which "case" is considered, like:
- how many transactions
- what are predefined UTXOs
- etc...

We can keep polishing the names still, but this won't be very productive and won't have real impact if code-reader "understand" the test or not.

pkg/storage/internal/testabilities/nosendtest/act_nosend.go
```diff
@@ -0,0 +1,166 @@
...
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/internal/fixtures/testusers"
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/storage"
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/wdk"
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/wdk/primitives"
+	"github.com/go-softwarelab/common/pkg/slices"
+	"github.com/go-softwarelab/common/pkg/to"
+	"github.com/stretchr/testify/require"
+)
+
+type NosendAct interface {
```
dorzepowski: ### very minor 

```suggestion
type NoSendAct interface {
```

pkg/storage/internal/testabilities/nosendtest/act_nosend.go
```diff
@@ -0,0 +1,166 @@
...
+
+	CreateAction(args wdk.ValidCreateActionArgs) (*wdk.StorageCreateActionResult, *transaction.Transaction)
+	ProcessAction(args wdk.ProcessActionArgs) *wdk.ProcessActionResult
+	CreateAndProcessNoSendAction(prevNoSendOutpoints []wdk.OutPoint) (createdNoSendChange, allocatedNoSendChangeAsInputs []wdk.OutPoint)
+	CreateActionNoSendArgsModifier(prevNoSendOutpoints []wdk.OutPoint, isNoSend bool) func(args *wdk.ValidCreateActionArgs)
+	CreateActionSendWithArgsModifier(sendWithHexStrings ...primitives.HexString) func(args *wdk.ValidCreateActionArgs)
+	CreateAndProcessSendWithAction(sendWithHexStrings []primitives.HexString, opts ...func(*wdk.ValidCreateActionArgs)) (*wdk.ProcessActionResult, string)
+}
+
+type nosendAct struct {
```
dorzepowski: ### very minor again

```suggestion
type noSendAct struct {
```

pkg/storage/internal/testabilities/nosendtest/fixture_nosend.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"testing"
+
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/internal/fixtures/testusers"
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/internal/satoshi"
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/storage"
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/storage/internal/testabilities"
+)
+
+type NosendFixture interface {
```
dorzepowski: ### very minor

```suggestion
type NoSendFixture interface {
```

pkg/storage/internal/testabilities/nosendtest/fixture_nosend.go
```diff
@@ -0,0 +1,39 @@
...
+	UserOwnsMultipleUTXOsToSpend(satoshis uint64)
+	UserOwnsGivenUTXOsToSpend(satoshis ...uint64)
+}
+
+type nosendFixture struct {
```
dorzepowski: ### very minor

```suggestion
type noSendFixture struct {
```

pkg/storage/provider_no_send_test.go
```diff
@@ -3,162 +3,337 @@ package storage_test
...
+		// given:
+		const inputSatoshis = 99904
+
+		givenStorage, cleanup := testabilities.Given(t)
+		defer cleanup()
+
+		activeStorage := givenStorage.Provider().WithRandomizer(randomizer.NewTestRandomizer()).GORM()
+		given, when, then := nosendtest.New(t, testusers.Alice, givenStorage, activeStorage)
```
dorzepowski: It's not big deal, so I won't push further
But just want to leave this thought/comment to you.
nosendtest.New, should wrap the functionality of testabilities.Given 
(should create it inside, and expose method Provider to use)

this way, there would be single point for setup the tests, and could even have some additional defaults.

Resolve this comment when you read it.
Unfortunatelly it cannot self-destruct

dorzepowski (approved):
There are no blocking comments.

PR #537, opened 2025-09-01
example: pushdrop token on with no-send & sendWith

chris-4chain:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter


mgosek-4chain (approved):

cain4chain (approved):
Looks good to me. Just missing the README for the example. Thanks

examples/wallet_examples/no_send_send_with/no_send_send_with.go
```diff
@@ -0,0 +1,123 @@
...
+)
+
+const (
+	tokensCount = 3
+	dataPrefix  = "exampletoken"
+)
+
+var rand = randomizer.NewTestRandomizer()
+
+func main() {
```
dorzepowski, line 26: maybe a comment (and/xor readme) what is this example about

examples/wallet_examples/no_send_send_with/token/mint.go
```diff
@@ -0,0 +1,75 @@
...
+	label := mintPushDropTokenLabel
+	satoshis := uint64(mintPushDropTokenSatoshis)
+
+	createActionResult, err := aliceWallet.CreateAction(ctx, wallet.CreateActionArgs{
+		Outputs: []wallet.CreateActionOutput{
+			{
+				LockingScript:      lockingScript.Bytes(),
+				Satoshis:           satoshis,
+				OutputDescription:  label,
+				Tags:               []string{"relinquish"},
```
dorzepowski: maybe it's good to know, why we're setting the tag value to relinquish 
if we don't know, maybe it's better to add some more meaningful tag, like "minted" or some name of this token.

examples/wallet_examples/no_send_send_with/token/mint.go
```diff
@@ -0,0 +1,75 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	return Token{
+		TxID:            createActionResult.Txid,
+		KeyID:           keyID,
+		FromIdentityKey: aliceIdentityKey,
+		Satoshis:        satoshis,
+	}, createActionResult.NoSendChange
```
dorzepowski, line 74: ### question

What is in NoSendChange in such a scenario (i will skip the amounts because I don't want to spend time to calculate proper values to make the example working):
1. we have single UTXO for X sats
2. we're making one no send tx -> we get couple of NoSendChange in result
3. we're making next no send tx 
4. only part of no send change from previous tx are used
5. a new no send change is created for the change from those used no send change
6. what happens with those no send change items that weren't used in point 4?

Image

In other words (based on the image)
Would the yellow stickies be also in place of white stickies, or there wouldn't be anything more, just orange sticky?

dorzepowski: Got answer:
Not used No Send Change won't be included in result.NoSendChange

In other words:
image

chris-4chain: CreateAction doesn't return unused noSendChange outputs provided in params. 
The adopter needs to take care of it, and check what noSendChange outputs are already used and what can be passed to the next create action. 
image

examples/wallet_examples/no_send_send_with/no_send_send_with.go
chris-4chain: Added a comment 
- ... and the documentation `.md`

examples/wallet_examples/no_send_send_with/token/mint.go
chris-4chain: I don't know - it seems like a standard in TS-examples. I applied your suggestion: `"mint"`.

PR #541, opened 2025-09-02
Task (#531) add validation to internalize action

### LINKED ISSUE #531: [Task] Wallet : provide additional validation to internalize action method
In this PR we will provide additional validations for `protocol` of  `wallet payment` type internalize action.
Additionally, transaction beef will be validated for any `protocol` type internalize action.

- add validation method for `protocol wallet payment` to check locking script match
- add condition for `protocol basket insertion` as per no locking script check requirement
- Add transaction beef validation to general internalize action flow

cain4chain:
## Description of Changes

In this PR we will provide additional validations for protocol of wallet payment type internalize action.
Additionally, transaction beef will be validated for any protocol type internalize action.

- Add validation method for protocol wallet payment to check locking script match in wallet interface 
- Add condition for protocol basket insertion as per no locking script check requirement in wallet interface 
- Add transaction beef validation to general internalize action flow in provider interface
- Add test cases for locking script match/mismatch
- Update existing test cases involving internalize action concerning locking script validation

Reference [TS-wallet-toolbox](https://github.com/bsv-blockchain/wallet-toolbox/blob/1812c31ab6b0943a18edd8c755c945f193905f19/src/signer/methods/internalizeAction.ts#L71) implementation 

## Linked Issues / Tickets

Closes #531

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter


ac4ch (approved):

pkg/internal/validate/validate_internalize_action_args.go
```diff
@@ -27,5 +34,34 @@ func ValidInternalizeActionArgs(args *wdk.InternalizeActionArgs) error {
...
+}
+
+// validateBeef validates the BEEF transaction using the provided services
+func validateBeef(ctx context.Context, txBytes []byte, services wdk.Services) error {
+	beef, txIDHash, err := transaction.NewBeefFromAtomicBytes(txBytes)
+	if err != nil {
+		return fmt.Errorf("failed to create atomic beef from bytes: %w", err)
+	}
+
+	ok, err := beef.Verify(ctx, services, false)
```
chris-4chain: Beef Verify is already done in the action. Why do we need to do it twice? 

https://github.com/bsv-blockchain/go-wallet-toolbox/blob/main/pkg/storage/internal/actions/internalize.go#L85

Besides - recently, we started to use an interface BeefVerifier that abstracts the algorithm.

chris-4chain (changes requested):

pkg/wallet/internal/actions/wallet_internalize_action.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// validateWalletPaymentOutput validates a wallet payment output using BRC-29
+func validateWalletPaymentOutput(keyDeriver *sdk.KeyDeriver, output wdk.InternalizeOutput, txOutput *transaction.TransactionOutput) error {
+
+	payment := output.PaymentRemittance
+
+	keyID := brc29.KeyID{
+		DerivationPrefix: string(payment.DerivationPrefix),
+		DerivationSuffix: string(payment.DerivationSuffix),
```
dorzepowski: I'm pretty sure that the bytes from payment prefix and suffix, should be encoded with base64

pkg/wallet/internal/actions/wallet_internalize_action.go
```diff
@@ -0,0 +1,88 @@
...
+
+	expectedAddress, err := brc29.AddressForSelf(senderPubKey, keyID, keyDeriver)
+	if err != nil {
+		return fmt.Errorf("failed to generate expected BRC-29 address: %w", err)
+	}
+
+	expectedLockScript, err := p2pkh.Lock(expectedAddress)
+	if err != nil {
+		return fmt.Errorf("failed to create expected P2PKH locking script: %w", err)
+	}
```
dorzepowski: You should use brc29.Lock, instead of creating address and p2pkh.Lock, like:

```suggestion
	expectedLockScript, err := brc29.Lock(brc29.PubHex(payment.SenderIdentityKey), keyID, keyDeriver)
	if err != nil {
		return fmt.Errorf("failed to create expected locking script: %w", err)
	}
```

pkg/wallet/internal/actions/wallet_internalize_action.go
```diff
@@ -0,0 +1,88 @@
...
+
+	"github.com/bsv-blockchain/go-sdk/transaction"
+	"github.com/bsv-blockchain/go-sdk/transaction/template/p2pkh"
+	sdk "github.com/bsv-blockchain/go-sdk/wallet"
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/brc29"
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/wdk"
+)
+
+// ValidateWalletInternalizeAction performs wallet-specific validation for internalize actions
+func ValidateWalletInternalizeAction(ctx context.Context, keyDeriver *sdk.KeyDeriver, args *wdk.InternalizeActionArgs) error {
```
dorzepowski: this whole code should be in `validate` package

pkg/internal/validate/validate_internalize_action_args.go
```diff
@@ -1,12 +1,19 @@
...
+	"github.com/bsv-blockchain/go-sdk/transaction"
 	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/wdk"
 )
 
 func ValidInternalizeActionArgs(args *wdk.InternalizeActionArgs) error {
+	return ValidInternalizeActionArgsWithServices(context.Background(), args, nil)
+}
+
+// ValidInternalizeActionArgsWithServices validates internalize action arguments including BEEF validation when services are provided
+func ValidInternalizeActionArgsWithServices(ctx context.Context, args *wdk.InternalizeActionArgs, services wdk.Services) error {
```
dorzepowski: whole `ValidateWalletInternalizeAction` should be moved here as `WalletInternalizeAction` and the validateBEEF should be called directly from it.

While the rest should be moved back to `ValidInternalizeActionArgs` (and this should also be called by `WalletInternalizeAction`)

pkg/internal/validate/validate_internalize_action_args.go
cain4chain: ok no problem thanks! I will remove this from the validation

pkg/internal/validate/validate_internalize_action_args.go
cain4chain: Thanks, updating now.

There was mention the beef validation is already handled -> https://github.com/bsv-blockchain/go-wallet-toolbox/pull/541#discussion_r2314960301

we can then skip this duplicate validation in our logic ?

pkg/wallet/internal/actions/wallet_internalize_action.go
cain4chain: Actually, the brc29.Lock Method only performs the method as per counter party private key perspective currently.
I can consolidate the key deriver validation and the senderPubKey variable though.

We could consider creating - LockForSelf method, but might be too much for a single use case. WDYT?

pkg/wallet/internal/actions/wallet_internalize_action.go
cain4chain: Since we are getting from storage I think this is already encoded base 64 format meaning we do not need to convert in that case??
To confirm I have double checked with manual test for internalize action example is working with valid transaction locking script. 

Let me know though if I might have still missed something there, thanks

pkg/internal/validate/validate_wallet_internalize_action.go
```diff
@@ -0,0 +1,82 @@
...
+
+	"github.com/bsv-blockchain/go-sdk/transaction"
+	"github.com/bsv-blockchain/go-sdk/transaction/template/p2pkh"
+	sdk "github.com/bsv-blockchain/go-sdk/wallet"
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/brc29"
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/wdk"
+)
+
+// ValidateWalletInternalizeAction performs wallet-specific validation for internalize actions
+func ValidateWalletInternalizeAction(ctx context.Context, keyDeriver *sdk.KeyDeriver, args *wdk.InternalizeActionArgs) error {
```
dorzepowski: the package name is `validate` not need to have this function prefixed with `Validate` because package name is part of this function name when calling it.

```suggestion
func WalletInternalizeAction(ctx context.Context, keyDeriver *sdk.KeyDeriver, args *wdk.InternalizeActionArgs) error {
```

pkg/wallet/internal/actions/wallet_internalize_action.go
dorzepowski: my bad, I didn't check the type underneath (I thought it's a payment from sdk, not wdk.

pkg/wallet/internal/actions/wallet_internalize_action.go
dorzepowski: actually, I see that we're using it in 2 places, so maybe it would be useful to have function LockForSelf.

Finally, we have real use case for adding it.

pkg/internal/validate/validate_wallet_internalize_action.go
dorzepowski, line 1: please move the code from this file to `validate_internalize_action_args.go` because this is how we did for other validation functions.

pkg/wallet/wallet.go
```diff
@@ -259,6 +259,10 @@ func (w *Wallet) InternalizeAction(ctx context.Context, args sdk.InternalizeActi
 		return nil, fmt.Errorf("invalid internalize action args: %w", err)
 	}
 
+	if err := validate.ValidateWalletInternalizeAction(ctx, w.keyDeriver, &wdkArgs); err != nil {
```
dorzepowski: please call the `validate.ValidInternalizeActionArgs` inside `validate.ValidateWalletInternalizeAction` so here we would have single validation function called.

pkg/internal/validate/validate_internalize_action_args.go
```diff
@@ -29,3 +34,77 @@ func ValidInternalizeActionArgs(args *wdk.InternalizeActionArgs) error {
 
 	return nil
 }
+
+// WalletInternalizeAction performs wallet-specific validation for internalize actions
+func WalletInternalizeAction(ctx context.Context, keyDeriver *sdk.KeyDeriver, args *wdk.InternalizeActionArgs) error {
```
chris-4chain: `Ctx` seems unused

pkg/internal/validate/validate_internalize_action_args.go
```diff
@@ -29,3 +34,77 @@ func ValidInternalizeActionArgs(args *wdk.InternalizeActionArgs) error {
...
+		}
+	}
+
+	return nil
+}
+
+// validateOutput validates a single output based on its protocol
+func validateOutput(keyDeriver *sdk.KeyDeriver, output wdk.InternalizeOutput, tx *transaction.Transaction) error {
+
+	txOutput := tx.Outputs[output.OutputIndex]
```
chris-4chain, line 64: Unnecessary empty line above

chris-4chain (approved):

ac4ch (approved):

PR #581, opened 2025-09-08
feat: sign action

chris-4chain:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter


pkg/internal/assembler/create_action_tx_assembler.go
```diff
@@ -53,6 +53,10 @@ func (a *CreateActionTransactionAssembler) Assemble() (*AssembledTransaction, er
 	return &AssembledTransaction{Transaction: a.tx, inputBEEF: a.inputBEEF}, nil
 }
 
+func (a *CreateActionTransactionAssembler) InputBEEF() *transaction.Beef {
```
dorzepowski: I't is not used anymore, after changes from my PR

pkg/wallet/internal/actions/wallet_create_action.go
```diff
@@ -64,70 +60,98 @@ func (a *CreateAction) handleNotNewTX(ctx context.Context) (*wallet.CreateAction
...
+	}
 
-	result, err := mapping.SignableTransactionResult(tx, a.wdkArgs, createActionResult)
+	result, err := mapping.SignableTransactionResult(tx.TxID(), txAtomic, a.wdkArgs, storageCreateActionResult)
 	if err != nil {
 		return nil, fmt.Errorf("failed to build signable transaction: %w", err)
 	}
+
+	err = a.PendingSignActionsCache.Set(storageCreateActionResult.Reference, &wdk.PendingSignAction{
+		Tx:               *tx.Transaction,
```
dorzepowski, line 119: Why we're not storing simpy the AssembledTransaction?

pkg/storage/internal/actions/create_process_inputs.go
```diff
@@ -134,7 +137,7 @@ func (proc *inputsProcessor) processInputs() (*processedInputsResult, error) {
 		return nil, fmt.Errorf("failed to get beef for inputs: %w", err)
 	}
 
-	if ok, err := proc.beef.Verify(proc.ctx, proc.chaintracker, true); err != nil {
+	if ok, err := proc.beefVerifier.VerifyBeef(proc.ctx, proc.beef, proc.chaintracker, true); err != nil {
```
dorzepowski, line 141: actually, why chaintracker is not a part of BEEFVerifier, and need to be passed by action?

mgosek-4chain (approved):

pkg/wallet/local_pending_sign_actions_cache.go
dorzepowski, line 1: this should be placed in internal package

pkg/wdk/pending_sign_action.go
```diff
@@ -0,0 +1,27 @@
...
+	CreateActionArgs ValidCreateActionArgs
+}
+
+// PendingSignActionsCache defines an interface for managing cached pending sign actions.
+// It allows setting, getting, and deleting actions based on a string reference.
+type PendingSignActionsCache interface {
+	Set(reference string, action *PendingSignAction) error
+	Get(reference string) (*PendingSignAction, error)
+	Delete(reference string) error
+}
```
dorzepowski: IMHO, this is not good moment to provide this as exposed API.
1. TS doesn't have something like that
2. We don't have any alternative (except the mocked Cache - which could be removed - and the tests with it also)
3. this force us to rewrite AssembledTransaction into smaller pieces and read it back again into it.

pkg/wallet/wallet_sign_action_test.go
```diff
@@ -0,0 +1,551 @@
...
+			require.NoError(t, err)
+			require.NotNil(t, signActionResult)
+
+			// and:
+			test.then(t, signActionResult)
+		})
+	}
+}
+
+func (s *WalletTestSuite) TestWalletSignAction_PendingSignActions_CacheErrors() {
```
dorzepowski, line 461: it would be good to test real world error cases also (or instead)
like:
- call signAction without createAction and passed some not existing reference
- call signAction without unlocking script for all custom inputs

chris-4chain: Added new test: `attempt to sign an action that doesn't exist`
For "call signAction without unlocking script" there is already a test: `attempt to sign action of tx with provided unlocking script length only, without client-side sign`

I ran tests with coverage to double check if such error cases are actually covered.

pkg/storage/internal/actions/create_process_inputs.go
chris-4chain: I don't get your question.

pkg/wallet/internal/actions/wallet_create_action.go
chris-4chain: It's not enough - because we need also the create action args that are not a part of AssembledTransaction.
Besides, your `AssembledTransaction` is an internal concept, on the other hand, in current implementation, the `wdk.PendingSignAction` is public - that's why.

pkg/wallet/local_pending_sign_actions_cache.go
chris-4chain: In current implementation, this needs to be public to allow users to provide it with custom TTL value if default value is not appropriate.

pkg/internal/assembler/create_action_tx_assembler.go
chris-4chain: Removed ✅

pkg/wallet/internal/testabilities/mock_pending_sign_action_cache.go
```diff
@@ -0,0 +1,44 @@
+package testabilities
+
+import (
+	"log/slog"
+
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/wallet/pending"
+)
+
+type MockPendingSignActionCache struct {
```
dorzepowski: ### minor
Now it's called Repository, not Cache

dorzepowski (approved):

PR #587, opened 2025-09-10
task : fix bitails false positive responses in broadcasting

### LINKED ISSUE #582: [Task] Resolve issue with Bitails returning false positive responses in broadcasting

cain4chain:
## Description of Changes

- Add double spend error code handling to bitails broadcast response methods
- Add descriptive error field for ECONNRESET and ECONNREFUSED errors
- Add corresponding test cases
- Handle response type conflicts for "code" field (`int `or `string`)

Example Response
`Bitails response Status: 201
Bitails response Body: [{"txid":"411193c81c7effbb42a5583cd0aacd5ef02daa94174491279d87c99410dc0bad"},{"txid":"c79082eee4433ea7bbea40abca8aeed0526d07e569e73b114accce04da3209fd"},{"txid":"0f4b4f89590b9296f72cf60aae70892372f2eaf66275c4912803715cc3c3c89d"},{"txid":"342c524c6dc291df790e1cdabae1d95f7d6a82229446d11a772e1817c192f766"},{"txid":"f9b0aadcb2564f028d2e216d0af6da4ceb348aa990cb4c43dedd57f04755b0a2"},{"error":{"code":-26,"message":"double-spend-attempt"}},{"error":{"code":"ECONNRESET"}},{"error":{"errno":-111,"code":"ECONNREFUSED","syscall":"connect","address":"127.0.0.1","port":1113}}]`

## Linked Issues / Tickets

Closes #582

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter


pkg/services/internal/bitails/broadcast.go
```diff
@@ -105,13 +106,27 @@ func (b *Bitails) classifyResponseError(resp broadcastResponse, result *wdk.Post
...
+	case ErrorCodeMissingInputs:
+		result.Result = wdk.PostedTxIDResultMissingInputs
+		result.DoubleSpend = true
+		shouldReturnError = true
 	default:
 		result.Result = wdk.PostedTxIDResultError
 		result.Error = fmt.Errorf("broadcast error code %d: %s", resp.Error.Code, msg)
+		upperMsg := strings.ToUpper(msg)
+		if strings.Contains(upperMsg, ErrorTokenECONNRESET) || strings.Contains(upperMsg, ErrorTokenECONNREFUSED) {
+			if strings.Contains(upperMsg, ErrorTokenECONNREFUSED) {
```
chris-4chain: I'm not sure if this kind of errors should be handled here. It looks like some network problem - and I though it is handled by `resty` and returned (from resty) as an error

pkg/services/internal/bitails/broadcast.go
```diff
@@ -105,13 +106,27 @@ func (b *Bitails) classifyResponseError(resp broadcastResponse, result *wdk.Post
...
 		result.Result = wdk.PostedTxIDResultAlreadyKnown
 		result.AlreadyKnown = true
-	case ErrorCodeMissingInputs:
+	case ErrorCodeDoubleSpend:
 		result.Result = wdk.PostedTxIDResultDoubleSpend
 		result.DoubleSpend = true
 		shouldReturnError = true
+	case ErrorCodeMissingInputs:
+		result.Result = wdk.PostedTxIDResultMissingInputs
+		result.DoubleSpend = true
```
chris-4chain: Why is this still marked as Double spend?

pkg/services/internal/bitails/broadcast.go
cain4chain: Yes, found the issue and is updated now. There was a alternative type issue in the response from `int` to `string`. 
Now we still use the conditional logic but we have the correct value for processing. Let me know what you think of the changes, thanks!

pkg/storage/internal/actions/process.go
```diff
@@ -369,7 +369,16 @@ func (p *process) broadcastTxs(ctx context.Context, txIDs []string, isDelayed bo
 
 	logger.DebugContext(ctx, "Categorizing transactions by status")
 
-	for txID, currentStatus := range knownTxStatusesLookup {
+	for _, txID := range txIDs {
+		currentStatus, ok := knownTxStatusesLookup[txID]
```
chris-4chain, line 373: Many thanks for finding and fixing a place where we wrongly changed the order of provided `txIDs`.

I think, we don't have to worry about `!ok` case - because it should not happen. Check how `getKnownTxStatuses` works.

So, instead of `DebugLog` and proceeding I would return an error.

pkg/services/internal/bitails/broadcast.go
dorzepowski: I see approaches that I would prefer
Option 1:
1. in broadcastError UnmarshalJSON  assign custom error code if the result is one of  those
2. add `case` statement for those new error codes

Option 2:
1. in broadcastError UnmarshalJSON  unmarshal always to some string code (which also could be an enumerated)
2. use in `case` statements those enumerated strings

pkg/services/internal/bitails/broadcast.go
```diff
@@ -105,13 +133,27 @@ func (b *Bitails) classifyResponseError(resp broadcastResponse, result *wdk.Post
...
+		// result.DoubleSpend = true
+		shouldReturnError = true
+	case ErrorCodeMissingInputs:
+		result.Result = wdk.PostedTxIDResultMissingInputs
 		shouldReturnError = true
 	default:
 		result.Result = wdk.PostedTxIDResultError
 		result.Error = fmt.Errorf("broadcast error code %d: %s", resp.Error.Code, msg)
+		upperMsg := strings.ToUpper(msg)
+		if strings.Contains(upperMsg, ErrorTokenECONNRESET) || strings.Contains(upperMsg, ErrorTokenECONNREFUSED) {
```
dorzepowski: It would be good, to add example in tests for those cases

chris-4chain (approved):

pkg/services/internal/bitails/broadcast.go
cain4chain: This is how it was setup on TS- wallet-toolbox for missing inputs. But since we now define missing inputs error in our code I think its safe to remove this. 

I will do some more manual testing for missing inputs though in another task just to cover some edge cases.

mgosek-4chain (approved):

jakubmkowalski (approved):

PR #593, opened 2025-09-11
Feat(#591) : add monitor task unfail failed txs

### LINKED ISSUE #591: [FEATURE] Monitor: Create monitor to check failed transactions
## Desired Solution

## Suggested Implementation

## Alternatives Considered

## Additional Context

## Categorize Feature

cain4chain:
## Description of Changes

This PR will contain monitor method for recovering failed transaction using the following:

 - Setting provenTxReq status to 'unfail' when 'invalid' will attempt to find a merklePath, and if successful will set the tx status to unmined and make outputs spendable related to those transactions.
 
 Reference to typescript implementation [here](https://github.com/bsv-blockchain/wallet-toolbox/blob/540d05855439bd200f51940ea21de8d5ec328c38/src/monitor/tasks/TaskUnFail.ts#L1) 
 
Closes #591 

## Testing Procedure

A new example has been added for listFailedActions. Here we are able to set all transactions that are set as "failed" and update the bsv known txes status field to "unfail". This will then allow the transactions to be processed by the unfail monitor method.

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter


pkg/defs/spec_ops.go
```diff
@@ -0,0 +1,9 @@
+package defs
+
+// ListActionsSpecOpFailedActionsLabel indicates that listActions should return only actions with status 'failed'.
+const ListActionsSpecOpFailedActionsLabel = "97d4eb1e49215e3374cc2c1939a7c43a55e95c7427bf2d45ed63e3b4e0c88153"
```
dorzepowski, line 4: this file should be in some internal package, as we don't want to expose those spec ops things.

pkg/storage/internal/actions/list_actions.go
```diff
@@ -62,6 +63,27 @@ func (l *listActions) ListActions(ctx context.Context, auth wdk.AuthID, args *wd
 		return nil, fmt.Errorf("failed to map inputs/outputs/labels: %w", err)
 	}
 
+	var hasFailedSpecOp, hasUnfailControl bool
```
dorzepowski: Instead of modifying this file, I would rather suggest to:
1. check at the *Storage provider* if the args contains spec op label 
2. if so, then remap args into dedicated wdk.ListFailedActionsArgs struct (this remaping should remove the spec ops label)
3. call the dedicated Provider.ListFailedActions
4. then Provider.ListFailedActions should map and validate the args and call the dedicated method actions.ListFailedActions (it could be method of listActions struct)

pkg/storage/internal/actions/process_unfail.go
```diff
@@ -0,0 +1,103 @@
...
+// If not found: set KnownTx back to 'invalid'.
+func (p *process) Unfail(ctx context.Context) error {
+	log := p.logger.With("action", "unfail")
+	log.InfoContext(ctx, "Attempting to process 'unfail' transactions")
+
+	paging := queryopts.Paging{Limit: unfailItemsPerPage, Sort: "asc"}
+
+	processed := 0
+	for range unfailMaxPages {
+		itemsPage, err := p.knownTxRepo.FindKnownTxIDsByStatuses(
```
dorzepowski, line 40: it would be good to use `queryopts.WithUntil` and query each page with the `creation_time` *until* the time of starting this process, so the paging will be deterministic in case of big numbers of transactions created in system.

mgosek-4chain (approved):

pkg/storage/internal/actions/actions.go
```diff
@@ -76,3 +76,13 @@ func New(
 		getBeef:               newGetBeef(logger, repos.KnownTx, services),
 	}
 }
+
+// UnFail is used by monitor to re-verify failed tx statuses
+func (a *Actions) UnFail(ctx context.Context) error {
+	return a.process.Unfail(ctx)
```
dorzepowski:
```suggestion
	return a.process.UnFail(ctx)
```

pkg/storage/internal/actions/actions.go
```diff
@@ -76,3 +76,13 @@ func New(
...
+}
+
+// ListFailedActions delegates to listActions.ListFailedActions
+func (a *Actions) ListFailedActions(ctx context.Context, auth wdk.AuthID, args *wdk.ListFailedActionsArgs) (*wdk.ListActionsResult, error) {
+	return a.listActions.ListFailedActions(ctx, auth, args)
+}
```
dorzepowski: because process and listActions are "wrapped" into Actions struct, their public methods should be available by default on Actions struct, so no need for this proxy methods.

pkg/wallet/internal/mapping/mapping_list_actions.go
```diff
@@ -121,11 +121,15 @@ func mapListActionsAction(action wdk.WalletAction) (sdk.Action, error) {
...
 
+const ActionStatusFailed sdk.ActionStatus = "failed"
+
 // mapActionStatus maps string status to sdk.ActionStatus
 func mapActionStatus(status string) (sdk.ActionStatus, error) {
 	switch status {
 	case "completed":
 		return sdk.ActionStatusCompleted, nil
+	case "failed":
+		return ActionStatusFailed, nil
```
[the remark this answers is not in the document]
cain4chain: Currently missing from go-sdk for actionStatusFailed - will be updated once it has been added in.

dorzepowski (commented):
please see the additional comment on https://github.com/bsv-blockchain/go-wallet-toolbox/pull/593#discussion_r2361975367

pkg/storage/internal/actions/actions.go
dorzepowski: please remove this wrapping also, as the UnFail from process should be exposed by default.

dorzepowski (approved):

PR #615, opened 2025-09-23
Task/update storage format for ts compatibility

cain4chain:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter


pkg/wallet/internal/mapping/mapping_list_outputs.go
```diff
@@ -20,7 +20,7 @@ import (
 func MapListOutputsArgs(args sdk.ListOutputsArgs) wdk.ListOutputsArgs {
 	result := wdk.ListOutputsArgs{
 		Basket:                    primitives.StringUnder300(args.Basket),
-		Tags:                      slices.Map(args.Tags, func(tag string) primitives.StringUnder300 { return primitives.StringUnder300(tag) }),
+		Tags:                      []primitives.StringUnder300{},
```
dorzepowski: Can you please share some details about those changes (what you want to achieve)?

Because if we wan't to send always an empty array, then removing `omitempty` should be enough, because slices.Map is always returning non nil slice

If you would like to be more safe, then I would rather go with custom MarshalJSON on those types method, that would set empty slice if it is nil, instead of just changing the mapper.

cain4chain: Seems that the `omitempty` resulted in null value where TS excepted empty list. The idea was to provide empty list and update if value is present, although sort of backwards approach it was more minimal than custom method, but I will update so we can streamline the mapping as you mentioned. Thanks :)

cain4chain: I have tested again and removing the `omitempty` was enough 👍 , it is updated, thanks

mgosek-4chain (approved):

pkg/wallet/internal/mapping/mapping_list_actions.go
```diff
@@ -17,11 +17,15 @@ import (
 // MapListActionsArgs maps sdk.ListActionsArgs to wdk.ListActionsArgs
 func MapListActionsArgs(args sdk.ListActionsArgs) wdk.ListActionsArgs {
 	result := wdk.ListActionsArgs{
-		Labels: slices.Map(args.Labels, func(label string) primitives.StringUnder300 { return primitives.StringUnder300(label) }),
+		Labels: []primitives.StringUnder300{},
 		Limit:  primitives.PositiveIntegerDefault10Max10000(to.ValueOr(args.Limit, 10)),
 		Offset: primitives.PositiveInteger(to.ValueOr(args.Offset, 0)),
 	}
 
+	if len(args.Labels) > 0 {
```
dorzepowski: I believe that this change is not needed

ac4ch (approved):

dorzepowski (approved):

PR #636, opened 2025-10-06
feat: storage manager - active, backup & conflicting storages

chris-4chain:
## Description of Changes

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter

### DIFF
--- a/pkg/storage/storage_manager.go
+++ b/pkg/storage/storage_manager.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log/slog"
+	"slices"
 
 	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/internal/logging"
 	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/storage/internal/managed"
@@ -149,6 +150,40 @@ func (m *WalletStorageManager) SyncToWriter(ctx context.Context, writer wdk.Wall
 	)
 
 	return
+}
+
+func (m *WalletStorageManager) SetActive(ctx context.Context, storageIdentityKey string) error {
+	if is.BlankString(storageIdentityKey) {
+		return fmt.Errorf("storage identity key must be provided and cannot be empty")
+	}
+
+	if m.activeStorage != nil && m.activeStorage.Settings.StorageIdentityKey == storageIdentityKey {
+		//already active
+		return nil
+	}
+
+	if _, err := m.MakeAvailable(ctx); err != nil {
+		return fmt.Errorf("failed to make storage available: %w", err)
+	}
+
+	newActiveIndex := slices.IndexFunc(m.stores, func(storage *managed.Storage) bool {
+		return storage.Settings.StorageIdentityKey == storageIdentityKey
+	})
+	if newActiveIndex == -1 {
+		return fmt.Errorf("storage with identity key %s not found among managed storages", storageIdentityKey)
+	}
+
+	newActive := m.stores[newActiveIndex]
+	_ = newActive
+	// TODO: add locking mechanism
+
+	if len(m.conflictingActives) > 0 {
+		// TODO
+	} else {
+
+	}
+
+	return nil
 }
 
 func (m *WalletStorageManager) getActiveReader() wdk.WalletStorageProvider {
--- a/pkg/storage/storage_manager.go
+++ b/pkg/storage/storage_manager.go
@@ -36,17 +36,16 @@ func NewWalletStorageManager(identityKey string, logger *slog.Logger, active wdk
 		panic("identity key must be provided and cannot be empty")
 	}
 
+	var stores []*managed.Storage
 	storesNum := len(backups) + to.IfThen(active != nil, 1).ElseThen(0)
-	if storesNum == 0 {
-		panic("at least one storage (active or backup) must be provided")
-	}
-
-	stores := make([]*managed.Storage, 0, storesNum)
-	if active != nil {
-		stores = append(stores, managed.NewManagedStorage(active))
-	}
-	for _, b := range backups {
-		stores = append(stores, managed.NewManagedStorage(b))
+	if storesNum > 0 {
+		stores = make([]*managed.Storage, 0, storesNum)
+		if active != nil {
+			stores = append(stores, managed.NewManagedStorage(active))
+		}
+		for _, b := range backups {
+			stores = append(stores, managed.NewManagedStorage(b))
+		}
 	}
 
 	logger = logging.Child(logger, "StorageManager")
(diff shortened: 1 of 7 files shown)


mgosek-4chain (approved):

cain4chain (approved):

jakubmkowalski (approved):

pkg/wdk/sync_to_writer_options.go
```diff
@@ -43,3 +36,13 @@ func WithMaxSyncItems[T types.Number](items T) SyncToWriterOption {
 		o.MaxSyncItems = must.ConvertToUInt64(items)
 	}
 }
+
+// WithSyncReader specifies the reader storage provider to use when syncing data to a writer storage.
+// If not provided, currently active storage will be used as the reader.
+func WithSyncReader(reader WalletStorageProvider) SyncToWriterOption {
+	return func(o *SyncToWriterOptions) {
```
dorzepowski, line 43: make sure the reader is not nil before returning the function, so we won't override default one with nil

pkg/storage/storage_manager.go
```diff
@@ -18,51 +20,102 @@ var _ wdk.WalletStorage = (*WalletStorageManager)(nil)
...
 	if m.isAvailable {
 		return m.activeStorage.Settings, nil
 	}
-	settings, err := m.activeStorage.MakeAvailableStorage(ctx, m.identityKey)
+
+	if len(m.stores) == 0 {
+		return nil, fmt.Errorf("no storage providers configured")
+	}
+
+	m.activeStorage = m.stores[0] //first storage is the active storage candidate
```
dorzepowski: minor: comment should start with whitespace

pkg/storage/storage_manager.go
```diff
@@ -109,18 +173,127 @@ func (m *WalletStorageManager) SyncToWriter(ctx context.Context, writer wdk.Wall
...
 }
 
+// SetActive Updates backups and switches to new active storage provider from among current backup providers.
+// Also resolves conflicting actives
+func (m *WalletStorageManager) SetActive(ctx context.Context, storageIdentityKey string) error {
+	if is.BlankString(storageIdentityKey) {
+		return fmt.Errorf("storage identity key must be provided and cannot be empty")
+	}
+
+	if m.activeStorage != nil && m.activeStorage.Settings.StorageIdentityKey == storageIdentityKey {
```
dorzepowski, line 183: shouldn't we first call MakeAvailable, and then make this check?
otherwise I suppose that Settings could be nil

pkg/storage/storage_manager.go
```diff
@@ -18,51 +20,102 @@ var _ wdk.WalletStorage = (*WalletStorageManager)(nil)
...
 		return m.activeStorage.Settings, nil
 	}
-	settings, err := m.activeStorage.MakeAvailableStorage(ctx, m.identityKey)
+
+	if len(m.stores) == 0 {
+		return nil, fmt.Errorf("no storage providers configured")
+	}
+
+	m.activeStorage = m.stores[0] //first storage is the active storage candidate
+	_, err := m.activeStorage.MakeAvailableStorage(ctx, m.identityKey)
```
dorzepowski, line 84: I wonder, what if active storage, doesn't really ThinksItIsActive?

pkg/storage/storage_manager.go
```diff
@@ -18,51 +20,102 @@ var _ wdk.WalletStorage = (*WalletStorageManager)(nil)
...
+	for _, store := range m.stores[1:] {
+		_, err := store.MakeAvailableStorage(ctx, m.identityKey)
+		if err != nil {
+			return nil, fmt.Errorf("failed to make available storage: %w", err)
+		}
+
+		if store.ThinksItIsActive() {
+			// swapping active storage
+			tmpBackups = append(tmpBackups, m.activeStorage)
+			m.activeStorage = store
```
dorzepowski, line 102: Maybe I'm wrong, but my memory tells me that TS version doesn't have active storage if there are conflicting storages.

pkg/storage/storage_manager.go
```diff
@@ -18,51 +20,102 @@ var _ wdk.WalletStorage = (*WalletStorageManager)(nil)
...
+		} else {
+			tmpBackups = append(tmpBackups, store)
+		}
+	}
+
+	for _, backup := range tmpBackups {
+		if !backup.ThinksActiveStorageIs(m.activeStorage.Settings.StorageIdentityKey) {
```
dorzepowski, line 109: hmmm... this is also somehow misleading:
if we have 3 storages (A, B, C)
A thinks that A is active and A is activeStorage
B thinks that B is active
C thinks that B is active - then suddenly C will be conflicting active storage, although it isn't active, because it's saying that B is active.

pkg/storage/storage_manager.go
```diff
@@ -18,51 +20,102 @@ var _ wdk.WalletStorage = (*WalletStorageManager)(nil)
...
+	m.conflictingActives = nil
+
+	tmpBackups := make([]*managed.Storage, 0, len(m.stores)-1)
+	for _, store := range m.stores[1:] {
+		_, err := store.MakeAvailableStorage(ctx, m.identityKey)
+		if err != nil {
+			return nil, fmt.Errorf("failed to make available storage: %w", err)
+		}
+
+		if store.ThinksItIsActive() {
```
dorzepowski, line 99: just something to think about:

Maybe the look below could be actually integrated (merged) in this place.

pkg/storage/storage_manager.go
chris-4chain: This check is for another case. When you already called `SetActive` with this storage or previous `MakeAvailable` call made it active. So no action needed. The `activeStorage` doesn't exist without settings. Look, how the constructor and MakeAvailable works.

pkg/storage/storage_manager.go
chris-4chain: This is only an assumption based on what the user provided to the StorageManager. This is adjusted later anyway.

pkg/storage/storage_manager.go
chris-4chain: https://github.com/bsv-blockchain/wallet-toolbox/blob/2b3cb455006f5afe16905089d2d02e0bcbbefa45/src/storage/WalletStorageManager.ts#L165-L185

pkg/storage/storage_manager.go
chris-4chain: Please help me and double check if your concern is also applicable to the TS code.
https://github.com/bsv-blockchain/wallet-toolbox/blob/2b3cb455006f5afe16905089d2d02e0bcbbefa45/src/storage/WalletStorageManager.ts#L165-L185

pkg/storage/storage_manager.go
```diff
@@ -109,18 +173,127 @@ func (m *WalletStorageManager) SyncToWriter(ctx context.Context, writer wdk.Wall
...
+		return nil
+	}
+
+	if _, err := m.MakeAvailable(ctx); err != nil {
+		return fmt.Errorf("failed to make storage available: %w", err)
+	}
+
+	newActiveIndex := stdslices.IndexFunc(m.stores, func(storage *managed.Storage) bool {
+		return storage.Settings.StorageIdentityKey == storageIdentityKey
+	})
```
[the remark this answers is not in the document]
chris-4chain: This is called after `MakeAvailable` so there is no option that `storage.Settings` is nil.

pkg/storage/storage_manager.go
dorzepowski: ok, so the issue is also in ts

dorzepowski (approved):

PR #708, opened 2025-12-15
UserID is not the relevant value here, it’s the public key (identityKey)

sirdeggen:
Check identityKey when checking certs

## Description of Changes

During Certificate Acquisition we don't need to validate UserID, which is just an internal auto-incrementing ID number; rather we ought to be checking the identityKey which is a universally applicable pubKey.

## Linked Issues / Tickets

Unsure, but this fixes:

Running [this example code](https://fast.brc.dev/?snippet=createCertificate) against a go-wallet-toolbox instance locally.

## Testing Procedure

I ran the example code, once with no changes which failed:
```RPC Error: access is denied due to an authorization error```

running after making these changes results in successful issuance.

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter


dzolt-4chain:
closing due to this change being added in  #770

PR #730, opened 2026-01-15
feat: chaintracks integration

dzolt-4chain:
## Description of Changes

Add go-chaintracks integration to wallet.

## Testing Procedure

- [x] All tests pass locally

## Checklist:

- [x] I have made corresponding changes to the documentation
- [ ] I have run the linter


pawellewandowski98 (approved):

examples/services_examples/chaintracksclient_examples/listen_for_reorgs_and_tips/main.go
```diff
@@ -0,0 +1,61 @@
...
+
+	err = svc.Start(ctx, chaintracksclient.Callbacks{
+		OnTip: func(header *chaintracks.BlockHeader) error {
+			logger.Info("new tip received",
+				"height", header.Height,
+				"hash", header.Hash.String(),
+			)
+			return nil
+		},
+	})
```
kuba-4chain, line 56: Technically you could merge those 2 examples: `listen_for_tips` and `listen_for_reorgs` into one example. I think it might be better in this case but it's just my opinion, not a major thing

pkg/services/chaintracksclient/service.go
```diff
@@ -0,0 +1,174 @@
...
+	adapter := &Adapter{
+		logger: logger,
+	}
+
+	for _, opt := range opts {
+		opt(adapter)
+	}
+
+	if adapter.ct == nil {
+		ct, err := cfg.Initialize(context.Background(), "wallet-toolbox", p2pClient)
```
kuba-4chain: Here in the constructor we're passing in `p2pClient`, which is `nil` in the examples, because we want the default one. It's probably going to be `nil` in most cases, so instead of passing it as a param here, I'd move that to an option, e.g. `WithCustomP2PClient()`, or `WithP2PClient()`

go.mod
```diff
@@ -36,54 +38,268 @@ require (
...
+	modernc.org/sqlite v1.44.3 // indirect
+	sigs.k8s.io/json v0.0.0-20250730193827-2d320260d730 // indirect
+	sigs.k8s.io/randfill v1.0.0 // indirect
+	sigs.k8s.io/structured-merge-diff/v6 v6.3.1 // indirect
+	sigs.k8s.io/yaml v1.6.0 // indirect
 )
 
 tool go.uber.org/mock/mockgen
+
+replace github.com/bsv-blockchain/go-teranode-p2p-client => github.com/shruggr/go-teranode-p2p-client v0.0.0-20251219052644-3b9eba1f195f
```
kuba-4chain: Should this `replace` be here?

examples/services_examples/chaintracksclient_examples/listen_for_reorgs_and_tips/main.go
dzolt-4chain: You're right! Merged

pkg/services/chaintracksclient/service.go
dzolt-4chain: Good catch, I added p2pclient as options

kuba-4chain (approved):

PR #733, opened 2026-01-19
feat(#720): notify about selected tx statuses

### LINKED ISSUE #720: [FEATURE] Transactions Status Updates
## Desired Solution
When interacting with wallet-toolbox, transactions are often processed and broadcasted in the background for speed and concurrency. Currently, a client does not have a way of being updated when a transaction has beed processed and broadcasted. Similarly, when a transaction reaches `MINED` status, a client may want to know about that.

## Suggested Implementation
In the TypeScript version of wallet-toolbox, a solution was introduced in a form of a callback method in the Monitor component which is responsible for sending _delayed_ transactions and marking transactions as _proven_ when they're mined. [see here](https://github.com/bsv-blockchain/wallet-toolbox/blob/36ec3700bfdef5dba09fb78b0cd8abda87bdc3b6/src/monitor/Monitor.ts#L57)

A Go version of the toolbox could leverage buffered channels to achieve this, which would benefit speed and concurrency of handling transaction updates.

## Categorize Feature
Feature Request

pawellewandowski98:
## Description of Changes

This PR introduces in-process notification channels for the monitoring subsystem so that important transaction lifecycle events (broadcasted / proven) can be observed and consumed by the infra server or other local consumers. It wires channels end-to-end: monitor daemon -> tasks -> server consumers. It also extends some storage/monitor APIs so tasks can produce richer event payloads (Merkle path, block info, etc.).

## Linked Issues / Tickets

Closes #720 

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter


dzolt-4chain (approved):

pkg/defs/monitor.go
```diff
@@ -32,6 +33,17 @@ func ParseMonitorTaskStr(task string) (MonitorTask, error) {
...
+type MonitorTaskResponse struct {
+	TxID   string
+	Status string
+
+	BlockHash   string
+	BlockHeight uint32
+	MerklePath  *transaction.MerklePath
+	MerkleRoot  string
+}
+
```
kuba-4chain, line 46: `MonitorTaskResponse` is a very general name. There are multiple monitor tasks. This is a response specifically about the transaction status, so maybe a better name would be one of those:
- `TransactionStatus`
- `TransactionStatusUpdate`
- `TransactionStatusMonitorUpdate` - if you really need the `Monitor` in the name

pkg/monitor/internal/tasks/check_for_proofs.go
```diff
@@ -3,31 +3,58 @@ package tasks
...
+			BlockHeight: res.BlockHeight,
+			BlockHash:   res.BlockHash,
+		}
+
+		select {
+		case t.communicationChannel <- msg:
+		case <-ctx.Done():
+			return fmt.Errorf("context done while sending tx status update: %w", ctx.Err())
+		default:
+			t.logger.Warn("TxBroadcasted channel full, dropping event")
```
kuba-4chain, line 55: This should be `TxProven` channel instead of `TxBroadcasted`, right?

pkg/monitor/internal/tasks/check_for_proofs.go
```diff
@@ -3,31 +3,58 @@ package tasks
...
-// UnFailChecker checks failed transactions against the chain and updates their status if they are found.
-type UnFailChecker interface {
-	UnFail(ctx context.Context) error
+	SynchronizeTransactionStatuses(ctx context.Context) ([]wdk.TxSynchronizedStatus, error)
 }
 
 type CheckForProofsTask struct {
-	storage TransactionStatusesSynchronizer
+	storage              TransactionStatusesSynchronizer
+	communicationChannel chan<- defs.MonitorTaskResponse
```
kuba-4chain, line 18: `communicationChannel` is very general. Just call it the `txProvenChannel`, which is how it's called externally. This way it will be easier to understand what's going on

pkg/monitor/internal/tasks/send_waiting.go
```diff
@@ -3,30 +3,62 @@ package tasks
...
-	SendWaitingTransactions(ctx context.Context, minTransactionAge time.Duration) error
+	SendWaitingTransactions(ctx context.Context, minTransactionAge time.Duration) (*wdk.ProcessActionResult, error)
 }
 
 type SendWaitingTask struct {
-	storage  WaitingTransactionsSender
-	firstRun bool
+	storage              WaitingTransactionsSender
+	firstRun             bool
+	communicationChannel chan<- defs.MonitorTaskResponse
```
kuba-4chain, line 20: Same here, call it `txBroadcastedCh` or `txBroadcastedChannel`

pkg/defs/monitor.go
```diff
@@ -32,6 +33,17 @@ func ParseMonitorTaskStr(task string) (MonitorTask, error) {
 	return parseEnumCaseInsensitive(task, CheckForProofsMonitorTask, SendWaitingMonitorTask, FailAbandonedMonitorTask, UnFailMonitorTask)
 }
 
+// MonitorTaskResponse represents the response from a monitoring task
+type MonitorTaskResponse struct {
+	TxID   string
+	Status string
```
kuba-4chain, line 39: This `Status` here is just a string and it could be an Enum. Especially because the "broadcasted" status is just a magic value given in the monitor task (see comment below)

pkg/monitor/internal/tasks/send_waiting.go
```diff
@@ -3,30 +3,62 @@ package tasks
...
 	}
 
+	if t.communicationChannel == nil || results == nil {
+		return nil
+	}
+
+	for _, res := range results.NotDelayedResults {
+		msg := defs.MonitorTaskResponse{
+			TxID:   res.TxID.String(),
+			Status: "broadcasted",
```
kuba-4chain, line 46: This status could be an enum from `pkg/defs/monitor.go`

pkg/monitor/monitor.go
```diff
@@ -30,6 +30,14 @@ type Daemon struct {
...
 	started   bool
 	startLock sync.Mutex
+
+	communicationChannels CommunicationChannels
+}
+
+// CommunicationChannels holds channels for communicating task results back to other components.
+type CommunicationChannels struct {
+	OnTxBroadcasted chan<- defs.MonitorTaskResponse
+	OnTxProven      chan<- defs.MonitorTaskResponse
```
kuba-4chain, line 40: `communicatonChannels` is too general. Call it `EventChannels`

pkg/monitor/options.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/defs"
+)
+
+// DaemonCommunicationOptions holds options for communication channels used by the monitor daemon.
+type DaemonCommunicationOptions struct {
+	onTxBroadcasted chan<- defs.MonitorTaskResponse
+	onTxProven      chan<- defs.MonitorTaskResponse
+}
```
kuba-4chain, line 11: `Event` instead of `Communication`? 🙏

PR #737, opened 2026-01-27
feat: add chaintracks to services and monitor

### LINKED ISSUE #725: [FEATURE] ReOrg Event subscription and handling
## Desired Solution
We want to be able to subscribe to reorg events.

We want to use the external [go-chaintracks](https://github.com/bsv-blockchain/go-chaintracks) package rather than the current local package because it's been built around Teranode. The local package ought to be deprecated to avoid confusion, any desirable features ought to be copied across to go-chaintracks.

Ideally we should be able to directly use the package within go-wallet-toolbo using the new go channel to listen for events. It should also be possible to configure use of a remote go-chaintracks-server for the same event subscription. 

## Suggested Implementation
I think the message ought to be a slice of chain.Hash which lists the orphaned block headers by hash. The mindset ought to be invalidate quickly, revalidate slowly. In other words, we want to immediately remove any merklepaths leading to the orphaned block's roots, and set those transactions into a state where their new Merkle paths will be acquired slowly during the next scheduled task.

## Alternatives Considered
The preference for David's chain tracks vs the internal one was based on Teranode live header sourcing only, if there are big reasons to consider keeping current implementation, I'd be open to the argument.

## Additional Context
Corresponding issue on the [chaintracks](https://github.com/bsv-blockchain/go-chaintracks/issues/29) side should be co-developed.

## Categorize Feature
Enterprise Robustness.

dzolt-4chain:
## Description of Changes

This pr introduces integration of chaintracks with wallet-toolbox to detect and handle blockchain reorganizations.

## Linked Issues / Tickets

Closes #725 

## Testing Procedure

- [x] All tests pass locally

## Checklist:

- [x] I have run the linter

### DIFF
--- a/pkg/internal/storage/repo/known_tx.go
+++ b/pkg/internal/storage/repo/known_tx.go
@@ -465,7 +465,8 @@ func (p *KnownTx) conditionsBySpec(spec *pkgentity.KnownTxReadSpecification) []g
 
 // InvalidateMerkleProofsByBlockHash sets MerklePath, BlockHeight, MerkleRoot, and BlockHash
 // to NULL for all KnownTx records where BlockHash matches any of the provided hashes.
-// Also sets status to 'unmined' so CheckForProofsTask will re-fetch proofs.
+// Also sets status to 'reorg' so CheckForProofsTask will re-fetch proofs.
+// Adds a history note to each affected transaction.
 // Returns the number of affected records.
 func (p *KnownTx) InvalidateMerkleProofsByBlockHash(ctx context.Context, blockHashes []string) (int64, error) {
 	var err error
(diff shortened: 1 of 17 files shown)


pkg/internal/storage/repo/known_tx.go
```diff
@@ -465,6 +463,78 @@ func (p *KnownTx) conditionsBySpec(spec *pkgentity.KnownTxReadSpecification) []g
...
+		res := p.db.WithContext(ctx).
+			Model(&models.KnownTx{}).
+			Where("block_hash IN ?", blockHashes).
+			Updates(map[string]any{
+				"merkle_path":  nil,
+				"block_height": nil,
+				"merkle_root":  nil,
+				"block_hash":   nil,
+				"status":       wdk.ProvenTxStatusReorg,
+			})
```
kuba-4chain, line 511: Maybe it would also make sense to reset the `attempts` counter to 0?

pkg/monitor/options.go
```diff
@@ -33,3 +37,10 @@ func WithProvenTxChannel(ch chan<- defs.TransactionStatusUpdate) func(*DaemonEve
 		o.onTxProven = ch
 	}
 }
+
+// WithReorgChannel sets the channel for receiving reorg events.
+func WithReorgChannel(ch <-chan *chaintracks.ReorgEvent) func(*DaemonEventOptions) {
+	return func(o *DaemonEventOptions) {
+		o.onReorg = ch
+	}
+}
```
kuba-4chain, line 50: I'm not sure this can be specified like that, because it looks like a user of go-toolbox can/should pass their own channel on which it'll send reorg notifications. This needs to be hooked into the monitor internally

pkg/services/services.go
```diff
@@ -295,6 +330,49 @@ func (s *WalletServices) logActiveServices() {
...
+	}
+
+	err := s.chaintracks.Start(ctx, chaintracksclient.Callbacks{
+		OnTip: func(bh *chaintracks.BlockHeader) error {
+			s.logger.Debug("new chain tip received",
+				"height", bh.Height,
+				"hash", bh.Hash.String(),
+			)
+			return nil
+		},
```
kuba-4chain, line 347: I see this is not doing anything with the tip.

Chaintracks is supposed to be a BHS replacement, so the toolbox needs to be able to receive tip information from chaintracks above all, and reorg is just an additional thing. So we need to handle tip announcements here.

Apart from that, Chaintracks should be added as an implementation of the current mechanism based on polling, similarly to how BHS is currently being setup in `pkg/services/services.go` in `New()`:
```go
	if config.BHS.Enabled {
		bhsService := bhs.New(options.RestyClientFactory.New(), logger, config.Chain, config.BHS)
		predefined = append(predefined, Named[Implementation]{
			Name: bhs.ServiceName,
			Item: Implementation{
				FindChainTipHeader:   bhsService.FindChainTipHeader,
				IsValidRootForHeight: bhsService.IsValidRootForHeight,
				CurrentHeight:        bhsService.CurrentHeight,
				ChainHeaderByHeight:  bhsService.ChainHeaderByHeight,
			},
		})
	}

```
So the polling mechanism will be the fallback for the callbacks here. I think there's already a logic implemented in the `CheckForProofs` task that will check if we already have the newest tip and if we do, it won't try to update it again, so at least we don't have to worry about that.

pkg/services/services_reorg.go
```diff
@@ -0,0 +1,65 @@
...
+	"log/slog"
+	"sync"
+
+	"github.com/bsv-blockchain/go-chaintracks/chaintracks"
+)
+
+// reorgBroadcaster allows multiple subscribers to receive reorg events
+type reorgBroadcaster struct {
+	mu          sync.RWMutex
+	subscribers []chan *chaintracks.ReorgEvent
```
kuba-4chain: Use `map[chan *chaintracks.ReorgEvent]struct{}` instead of a slice of channels. 

You can then do Subscribe:
```go
	b.mu.Lock()
	b.subscribers[ch] = struct{}{}
	b.mu.Unlock()
```
and Unsubscribe:
```go
	b.mu.Lock()
	delete(b.subscribers, ch)
	close(ch)
	b.mu.Unlock()
```
instead of the loop. 

Broadcasting stays almost the same.

pkg/services/services.go
dzolt-4chain: Agree, but I wanted to make this change in different PR so this one's scope is only related to chaintracks and reorg specifically

pkg/services/services_reorg.go
dzolt-4chain: You're right! Changed

pkg/internal/storage/repo/known_tx.go
dzolt-4chain: Good thinking, also added attempts count

pkg/monitor/options.go
dzolt-4chain: I've added a more detailed comment to that function. I think if a user doesn't create infra server from script in this repo and they wire everything by themselves, they should use this option to create chaintracks from activeServices and then pass the channel to this function

kuba-4chain (approved):

PR #747, opened 2026-02-06
feat: add option to define reference and method to list transactions

### LINKED ISSUE #739: [FEATURE] Allow for passing custom ID to actions for a later reference
## Desired Solution
When creating or signing an action, a ReferenceID in a form of a method parameter should be allowed to be specified. That same ID should act as a key with which a client can reference or look up a specific transaction. This ID should also be quoted back when receiving transaction status updates for broadcasted and mined transactions.

## Suggested Implementation
Possibly, a `reference` field, already present in the `SignAction()` method could be reused for this purpose. This needs to be verified.

Example:
```go
	createArgs := wallet.CreateActionArgs{
		ID: user_defined_actionID, // this is the extra field
		Description: "marker transaction",
		Labels:      []string{markerID},
		Outputs: []wallet.CreateActionOutput{
			{
				LockingScript:     dataOutput.LockingScript.Bytes(),
				Satoshis:          0,
				Basket:            kind.String(),
				OutputDescription: "OP_RETURN marker output",
				Tags:              []string{kind.String(), markerID},
			},
		},
		Options: &wallet.CreateActionOptions{
			AcceptDelayedBroadcast: to.Ptr(false),
			RandomizeOutputs:       to.Ptr(false),
		},
	}

```

## Categorize Feature
QoL feature, in rare cases solving common issues where transactions can't be queried because the client has no reference to them and issues with race conditions when receiving status updates and correlating them by TxID.

### LINKED ISSUE #742: [FEATURE] Query for transactions, meaningful statuses for transaction states
## Desired Solution
Currently, when broadcasting a transaction (creating an action) through wallet-toolbox and specifying `AcceptDelayedBroadcast: true`, there is no way to learn about the state of that transaction at any moment in the future through the toolbox. The only thing a client has is a `txid` returned from the action, with which a client usually goes to external providers (e.g. WhatsOnChain) to learn about the transaction state.

The only methods for querying the toolbox are `ListOutputs()` and `ListActions()` (also certificates and failed actions, but that's of no help in this use case). Both of those do not allow for querying by `txid` (or a [reference ID](https://github.com/bsv-blockchain/go-wallet-toolbox/issues/739)). Also, none of them return an information about the block height, hash and merkle proof when the transaction is mined.

After the addition of [status update callback channels](https://github.com/bsv-blockchain/go-wallet-toolbox/issues/720), a client can learn about transaction state, but there are cases in which a message can be dropped (channel full, lost connection, client not listening, etc).

So, a solution is proposed to introduce a new query for Transactions that would return the transactions status and block information with merkle proofs when a transaction is mined. 

## Suggested Implementation
Transactions should be queryable at least by `TxID` (and [reference ID](https://github.com/bsv-blockchain/go-wallet-toolbox/issues/739) once it's implemented)

The transaction status structure needs to be unified across all places it's returned from, that is:
* the query
* the update channel in `BackgroundBroadcaster`
* the `BroadcastedTxChannel` in Monitor
* the `ProvenTxChannel` in Monitor

The status field should be meaningful to a client that is not familiar with internal wallet-toolbox transaction statuses. It should include a finite set of states that a transaction can be in, e.g.:
* `Unknown` - wallet-toolbox never heard about this transaction
* `Waiting` - transaction was created but was not yet broadcasted
* `Broadcasted` - transaction was broadcasted but is not yet included in the block
* `Mined` - transaction is mined and block info for that transaction is available

When an error happens at any point on the transaction's journey that cannot be solved automatically by the wallet-toolbox, it should also be reported together with relevant information, e.g.:
* `DoubleSpend` - should include the txid(s) that are marked as double spending
* `ServiceError` - should include an error message from the service
* `InvalidTx` - should include a reason why the transaction is invalid (error message)

## Categorize Feature
This issue is critical to creating scalable and fast (through concurrency) enterprise and personal wallets.

pawellewandowski98:
## Description of Changes

## Linked Issues / Tickets

Closes #739
Closes #742 

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter


dzolt-4chain (approved):

kuba-4chain (commented):
All good, nice work.

Just one thing, you mentioned you wanted to move the `TransactionStatusUpdate` struct to a different place and name it differently. I think it would be a good idea to do so.

Also, the statuses need to be unified. Currently `unproven` and `success` mean the same thing and there are some other statuses which are unhandled. It would be really nice to have a mapping to user-friendly statuses, such as the ones I mentioned in the issue.

kuba-4chain (approved):

PR #750, opened 2026-02-09
feat: change broadcast beef to ef

### LINKED ISSUE #726: [FEATURE] Broadcasting - Minimal Checks to Avoid False Negatives During Validation
## Desired Solution
Transaction broadcasting should not fail on validation of the wallet's own outputs Merkle path data. This might happen if there is a block reorg at the same time as a createAction is taking place, such that inputs are used, but on validation the Merkle Paths fail.

## Suggested Implementation

What seems to be happening at the moment is the incorrect `beef.Verify()` method is called. Something which validates Merkle Paths ONLY -- frankly useless, the original intention was for actual transaction validation as well as Merkle paths. Let's switch to scripts only validation from `spv.VerifyScripts()` in go-sdk to remove possibility of blocking our own txs here.

ARC will validate beef if you send it - so I suggest that we do not actually send BEEF to any of the broadcast services. Most only take rawtx as it is, ARC will check scripts so EF is recommended for now. It's possible we'll drop that format assuming Teranode doesn't benefit from it.

## Alternatives Considered
Blocking new broadcasts processes while there are unprocessed reorg tasks. Seems like that could cause more problems.

## Additional Context
BEEF ingestion was initially supposed to reduce submission of invalid txs by an ARC instance to the nodes by blocking them prior to propagation. unfortunately this means ARC and wallet-toolbox must be perfectly in sync reorg wise, which we found to practically not be the case.

## Categorize Feature
Stability improvement

dzolt-4chain:
## Description of Changes

This PR refactors the transaction broadcasting system from sending full BEEF to sending Extended Format (EF) per transaction. It also adds script verification before broadcasting.

Key Changes

1. Split PostBEEF into separate service methods:
- PostEF - For EF format broadcasting (ARC service)
- PostTX - For raw transaction broadcasting (WoC, Bitails)
- PostFromBEEF - Orchestration method that iterates over txIDs, extracts transactions from BEEF, and calls 
appropriate services

2. Added Scripts Verification:
- New ScriptsVerifier interface for validating transaction scripts before broadcast
- Scripts verification replaces BEEF verification during broadcast (BEEF verification now only happens during CreateAction with external inputs)
- Prevents broadcasting transactions with invalid signatures

3. EF Conversion per Transaction:
- New TxToEFHex function generates EF for a specific transaction (not the whole BEEF)
- BindBumpsAndTransactions links merkle proofs to transactions in BEEF
- ValidateSingleLeafTx ensures BEEF contains only one unmined leaf transaction (prevents disconnected tx chains)

4. BEEF Verification Changes:
- BEEF is now verified only during CreateAction when external inputs are provided (validates merkle paths from external sources) and InternalizeAction.
- During ProcessAction/broadcast, only script verification is performed (not full BEEF verification)
- This avoids merkle path validation failures during block reorgs, since ARC validates scripts anyway

## Linked Issues / Tickets

Closes #726 

## Testing Procedure

- [x] All tests pass locally

## Checklist:

- [x] I have run the linter


pkg/monitor/monitor.go
```diff
@@ -300,18 +300,18 @@ func (d *Daemon) handleNewTipEvents(ctx context.Context) {
 		)
 
 		go func(h *chaintracks.BlockHeader) {
-			results, err := d.storage.ProcessNewTip(ctx, header.Height, header.Hash.String())
+			results, err := d.storage.ProcessNewTip(context.Background(), h.Height, h.Hash.String())
```
kuba-4chain: Minor: I think it would make sense to pass `ctx` to the `Start()` monitor method and pass it down here instead of `context.Background()` (same for processReorg. That way we remove the possibility of hanging IO operations

pkg/services/internal/whatsonchain/service.go
```diff
@@ -210,28 +209,10 @@ func (woc *WhatsOnChain) FindChainTipHeader(ctx context.Context) (*wdk.ChainBloc
...
-	for i, txid := range txIDs {
-		result := woc.processSingleTx(ctx, txid, rawTxs[i])
-		txResults = append(txResults, result)
-	}
-
-	return &wdk.PostedBEEF{TxIDResults: txResults}, nil
+// PostTX broadcasts a single raw transaction to WhatsOnChain
+func (woc *WhatsOnChain) PostTX(ctx context.Context, rawTx []byte, txID string) (*wdk.PostedTxID, error) {
+	result := woc.processSingleTx(ctx, txID, rawTx)
+	return &result, nil
```
kuba-4chain: You don't need to pass `txID` because you get it in `processSingleTx()`. Bitails service does the same so you can remove `txID` from the `PostTX` interface method entirely

pkg/storage/internal/actions/internalize.go
```diff
@@ -92,6 +95,25 @@ func (in *internalize) Internalize(ctx context.Context, userID int, args *wdk.In
 		return nil, fmt.Errorf("provided beef is not valid")
 	}
 
+	// verify scripts for all unmined transactions in BEEF
+	for txIDHash, beefTx := range beef.Transactions {
+		// no raw tx available or skip already mined txs
+		if beefTx.Transaction == nil || beefTx.Transaction.MerklePath != nil {
+			continue
+		}
+
```
kuba-4chain, line 113: When Internalizing, the toolbox does not allow to pass BeefV2 with "known txs" (I don't know if that's good or bad), so there can't be a situation when `beefTx.Transaction == nil`. If this happens it should throw an error instead of just skipping that tx

pkg/storage/internal/actions/process.go
```diff
@@ -444,18 +448,28 @@ func (p *process) broadcastTxs(ctx context.Context, txIDs []string, isDelayed bo
...
-	if ok, err := p.beefVerifier.VerifyBeef(ctx, beef, false); err != nil {
-		return nil, fmt.Errorf("failed to verify beef: %w", err)
-	} else if !ok {
-		return nil, fmt.Errorf("provided beef is not valid")
+	// Verify scripts for each transaction before broadcasting.
+	// We use scripts-only verification (not full BEEF verification) because:
+	// 1. We're sending EF format to ARC, not BEEF
+	// 2. Merkle path validation can fail during block reorgs
+	// 3. ARC will validate scripts anyway
+	for _, txID := range readyToSendTxIDs {
```
kuba-4chain, line 465: Number 3 reason in the comments does not make sense, because it's saying that we validate scripts only because Arc will validate scripts anyways 😆 

I would put there another comment instead:
```go
// 3. Input BEEF was validated when creating action with external inputs or during internalization, so there's no need to validate it again
```

pkg/internal/txutils/hydrate_from_beef.go
```diff
@@ -0,0 +1,51 @@
...
+// that need access to the source transactions' outputs.
+func HydrateTransactionFromBEEF(tx *transaction.Transaction, beef *transaction.Beef) error {
+	for _, input := range tx.Inputs {
+		err := hydrateInput(input, beef, 0)
+		if err != nil {
+			return fmt.Errorf("failed to hydrate input %s of tx %s: %w", input.SourceTXID.String(), tx.TxID().String(), err)
+		}
+	}
+	return nil
+}
```
kuba-4chain: ❗Important ❗ 

What this function does is that it hydrates all the parents recursively. The EF format only needs the parents of itself, so only one level deep.

RawTx:
- Inputs
    - PrevHash
    - PrevOutputIdx
    - UnlockingScript
    - ...
- Outputs
    - LockingScript
    - ...

EF:
- Inputs
    - PrevHash
    - PrevOutputsIdx
    - UnlockingScript
    - ...
    - **PreviousOutputs** (that are used as inputs here)
        - **Satoshis**
        - **LockingScript**
- Outputs
    - LockingScript
    - ...

BEEF:
- Inputs
    - PrevHash
    - PrevOutputsIdx
    - UnlockingScript
    - ...
    - PreviousOutputs (that are used as inputs here)
        - Satoshis
        - LockingScript
        - **PreviousOutputs** (that are used as inputs here)
            - **Satoshis**
            - **LockingScript**
            - **PreviousOutputs** (that are used as inputs here)
                - **Satoshis**
                - **LockingScript**
                - ... (until a mined parent is included)

pkg/storage/internal/actions/process.go
```diff
@@ -444,18 +448,28 @@ func (p *process) broadcastTxs(ctx context.Context, txIDs []string, isDelayed bo
...
+			return nil, fmt.Errorf("transaction %s not found in beef", txID)
+		}
+		if err := txutils.HydrateTransactionFromBEEF(tx, beef); err != nil {
+			return nil, fmt.Errorf("failed to hydrate transaction %s for verification: %w", txID, err)
+		}
+		if ok, err := p.scriptsVerifier.VerifyScripts(ctx, tx); err != nil {
+			return nil, fmt.Errorf("failed to verify scripts for tx %s: %w", txID, err)
+		} else if !ok {
+			return nil, fmt.Errorf("scripts validation failed for tx %s", txID)
+		}
```
kuba-4chain, line 475: You don't need to `HydrateTransactionFromBEEF()` for every tx. It only needs to be done once, for the "last" transaction (the one with no children)

pkg/storage/internal/actions/create_process_inputs.go
```diff
@@ -134,10 +138,32 @@ func (proc *inputsProcessor) processInputs() (*processedInputsResult, error) {
...
+		tx := beefTx.Transaction
+		if err := txutils.HydrateTransactionFromBEEF(tx, proc.beef); err != nil {
+			return nil, fmt.Errorf("failed to hydrate tx %s for script verification: %w", txIDHash, err)
+		}
+
+		if ok, err := proc.scriptsVerifier.VerifyScripts(proc.ctx, tx); err != nil {
+			return nil, fmt.Errorf("script verification failed for tx %s : %w", txIDHash, err)
+		} else if !ok {
+			return nil, fmt.Errorf("scripts are not valid for tx %s", txIDHash)
+		}
```
kuba-4chain, line 166: Same here, `HydrateTransactionFromBEEF()` should only be done once, not in a loop

pkg/storage/internal/actions/internalize.go
```diff
@@ -92,6 +95,25 @@ func (in *internalize) Internalize(ctx context.Context, userID int, args *wdk.In
...
+		if err := txutils.HydrateTransactionFromBEEF(tx, beef); err != nil {
+			return nil, fmt.Errorf("failed to hydrate tx %s for script verification: %w", txIDHash, err)
+		}
+
+		if ok, err := in.scriptsVerifier.VerifyScripts(ctx, tx); err != nil {
+			return nil, fmt.Errorf("script verification failed for tx %s : %w", txIDHash, err)
+		} else if !ok {
+			return nil, fmt.Errorf("scripts are not valid for tx %s", txIDHash)
+		}
+	}
```
kuba-4chain, line 119: Same here, `HydrateTransactionFromBEEF()` should only be done once, not in a loop

kuba-4chain (approved):

PR #778, opened 2026-03-06
fix: mark all known outputs as spent when used as inputs

shruggr:
## Summary
- Removes the `output.Change` guard in `buildInputsDefinition()` so all known output IDs are collected for `changeOutputIDs`
- Without this fix, non-change outputs (e.g. ordinals) used as inputs in `createAction` are never marked as spent (`spendable=false`, `spent_by=txID`)
- Aligns with the TypeScript wallet-toolbox behavior which marks all known outputs as spent regardless of the `Change` flag

## Test plan
- [x] All existing tests pass (`go test ./...`)
- [x] End-to-end test: use a non-change output (ordinal) as input in createAction, verify it gets marked as not spendable

### DIFF
--- a/pkg/storage/internal/actions/create_process_inputs.go
+++ b/pkg/storage/internal/actions/create_process_inputs.go
@@ -180,9 +180,7 @@ func (proc *inputsProcessor) buildInputsDefinition() (*processedInputsResult, er
 
 		var newXInput *xinputDefinition
 		if output != nil {
-			if output.Change {
-				changeOutputIDs = append(changeOutputIDs, output.ID)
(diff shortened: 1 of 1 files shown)


pawellewandowski98:
Thanks for reporting the bug, I tested the solution and db error occurs when creating tx -> `violates foreign key constraint`. 

Fixed in  #779

shruggr:
Hmm. I think it's only half way there then. Before the fix, when I spent an output, it remained spendable

pawellewandowski98:
Fixed in #779

PR #791, opened 2026-03-16
fix: add no-op Destroy handler to RPCStorageProvider

shruggr:
## Summary
- Adds a no-op `Destroy()` method to `RPCStorageProvider`
- Remote clients (`StorageClient`) call `destroy()` when cleaning up their local wallet instance
- The TypeScript `StorageServer` already ignores this call (logs it as `IGNORED` and returns success)
- The Go `RPCStorageProvider` was missing the handler, causing `"method 'destroy' not found"` RPC errors on every client disconnect

## Test plan
- [ ] `go build ./pkg/storage/rpcserver/...` passes
- [ ] Server no longer logs RPC errors when clients call destroy


mrz1836 (approved):
LGTM

PR #803, opened 2026-03-27
fix: restore spent outputs when broadcast returns failed status

shruggr:
## Summary

- When ARC returns `doubleSpend` or `invalidTx`, `updateSingleTx` sets the transaction to `failed` but doesn't restore spent inputs — leaving them permanently stuck with `spendable=0` and deleted UTXO entries
- Calls `RecreateSpentOutputs` after setting failed status, matching the TypeScript wallet-toolbox behavior where transitioning to `failed` restores all inputs inline
- Follows up on PR #785 (restore all spent outputs on abort) and PR #779 (mark all inputs as spent)

## Test plan

- [x] Existing storage test suite passes (verified locally, all green)
- [x] Verify with a real wallet that a double-spend rejection correctly restores the spent inputs


pawellewandowski98 (approved):

sirdeggen (approved):

mrz1836 (changes requested):
Fix failing tests in CI

shruggr:
Fixed

PR #804, opened 2026-03-27
feat: remove custom reference and return labels in notification channels

pawellewandowski98:
## Description of Changes

This PR streamlines the transaction creation process by transitioning from user-provided custom references to system-generated unique identifiers. To maintain the ability to correlate transactions with external systems, it enhances the notification channels to include transaction labels, which now serve as the primary mechanism for custom metadata and correlation.

## Linked Issues / Tickets

## Testing Procedure

- [ ] All tests pass locally

## Checklist:

- [ ] I have run the linter


dzolt-4chain (approved):

pkg/storage/internal/actions/process.go
```diff
@@ -637,6 +644,8 @@ func (p *process) updateSingleTx(
 		return sendWithResult, reviewActionResult, err
 	}
 
+	reviewActionResult.Labels = labels
+
```
kuba-4chain, line 648: You can pass the labels to the `singleTxBroadcastResult()` and add it to `reviewActionResult` there, because if there's an error, the `reviewActionResult` is still returned but will be missing the labels.

pkg/storage/provider.go
```diff
@@ -1045,6 +1049,11 @@ func (p *Provider) ListTransactions(ctx context.Context, auth wdk.AuthID, args w
 
 	totalCount := uint64(len(knownTxs))
 
+	txLabelsMap, err := p.repo.GetLabelsForTxIDs(ctx, txIDs)
```
kuba-4chain, line 1052: Here you could use `GetLabelsForTransactions()` because you have database IDs and it's quicker because it's not doing any joins.

kuba-4chain: I was wrong

PR #816, opened 2026-04-02
Add AddWalletStorageProvider to WalletStorageManager

shruggr:
## Summary
- Adds `AddWalletStorageProvider(ctx, provider)` method to `WalletStorageManager`
- Adds `WithStorageManager(mgr)` option for wallet construction

## Motivation
The remote-wallet pattern requires:
1. Creating a wallet before the `StorageClient` (client needs wallet for BRC-103/104 auth)
2. Adding storage providers to the manager after wallet construction

Without these changes, there's no way to use remote storage in Go because of the circular dependency: wallet needs storage, storage client needs wallet.

The TypeScript wallet-toolbox supports both patterns — `WalletStorageManager.addWalletStorageProvider` and accepting a `WalletStorageManager` directly in the `Wallet` constructor. These changes bring the Go version to parity.

### AddWalletStorageProvider
- Wraps provider in `ManagedStorage`, calls `MakeAvailableStorage`
- Appends to the stores list
- Resets `isAvailable` and calls `MakeAvailable` to re-partition active/backups

### WithStorageManager
- New wallet option that accepts a pre-built `wdk.WalletStorage`
- When set, the wallet uses it directly instead of creating its own manager
- The `activeStorageFactory` nil check only applies when no manager is provided

## Usage
```go
// 1. Create empty storage manager
mgr := storage.NewWalletStorageManager(identityKey, logger, nil)

// 2. Create wallet with the manager
wallet, _ := walletcore.New(network, privKey, nil,
    walletcore.WithStorageManager(mgr),
    walletcore.WithServices(services),
)

// 3. Create remote storage client (needs wallet for auth)
client, cleanup, _ := storage.NewClient(remoteURL, wallet)

// 4. Add remote storage to the manager
mgr.AddWalletStorageProvider(ctx, client)
```

## Test plan
- [x] Verify existing tests pass
- [ ] Verify remote wallet pattern works end-to-end

### DIFF
--- a/pkg/storage/storage_manager.go
+++ b/pkg/storage/storage_manager.go
@@ -59,6 +59,23 @@ func NewWalletStorageManager(identityKey string, logger *slog.Logger, active wdk
 	}
 }
 
+// AddWalletStorageProvider adds a new storage provider to the manager after construction.
+// This enables the remote-wallet pattern where the wallet is created with an empty storage manager,
+// then the storage client (which needs the wallet for auth) is added dynamically.
+// After adding, the manager re-partitions all stores by calling MakeAvailable.
+func (m *WalletStorageManager) AddWalletStorageProvider(ctx context.Context, provider wdk.WalletStorageProvider) error {
+	store := managed.NewManagedStorage(provider)
+	if _, err := store.MakeAvailableStorage(ctx, m.identityKey); err != nil {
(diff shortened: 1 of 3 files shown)


mrz1836 (approved):
LGTM

PR #824, opened 2026-04-03
fix: internalizeAction merge path FK violation on NULL basket outputs

### LINKED ISSUE #817: internalizeAction merge path fails with FK violation on NULL basket_name outputs
## Problem

When `internalizeAction` is called for a transaction that already exists for the same user (merge path), and the existing output has `basket_name = NULL` (e.g., an outgoing payment output), the Go implementation returns a foreign key constraint error:

```
failed to process internalizeAction: failed to upsert outputs (isMerge): failed to save output: db transaction failed: failed to save output: violates foreign key constraint
```

The TypeScript implementation handles this correctly by performing a **partial update** that assigns the change basket to the output.

## Root Cause

Two issues in `pkg/storage/internal/actions/internalize.go`:

1. **`upsertExistingTx()` uses GORM `Save()` (full row write)** instead of partial column updates. This writes all fields including `basket_name` and nested `UserUTXO`, triggering the FK constraint `(user_id, basket_name) → bsv_output_baskets(user_id, name)`.

2. **No basket auto-creation in the merge path.** The TypeScript version calls `findOrInsertOutputBasket()` to ensure baskets exist before assignment. The Go codebase has `cachedBasketMaker` (`pkg/internal/storage/repo/cached_basket_maker.go`) which does exactly this, but it's not wired into the internalize path.

## How TypeScript Handles This

`mergeWalletPaymentForOutput()` in `internalizeAction.ts` (line 494-509):
- Calls `storage.updateOutput(outputId, { basketId, type, change, ... })` — partial update, only specified columns
- Uses `basketId` (integer FK) resolved via `findOrInsertOutputBasket()` which auto-creates missing baskets

## Reproduction

1. Create a transaction via `createAction` (sendBsv to own address)
2. Transaction completes with output `basket_name = NULL` (outgoing payment)
3. Owner sync detects the output belongs to the same user
4. Calls `internalizeAction` with the same txid → merge path → FK violation

## Suggested Fix

1. Wire `cachedBasketMaker` into the internalize merge path to ensure baskets exist before writing
2. Switch merge path from `SaveOutputs` (full GORM Save) to partial updates — extend `OutputUpdateSpecification` to support `BasketName`, `Change`, `Type`, `ProvidedBy`, `Purpose`, `SenderIdentityKey`, `DerivationPrefix`, `DerivationSuffix`

shruggr:
## Summary
- Fixes FK constraint violation when `internalizeAction` merges an output that has `basket_name = NULL` (e.g., outgoing payment to self detected by owner sync)
- Adds `BasketName` to `UserUTXO` model in `SaveOutputs` — was previously unset, defaulting to empty string
- Adds `FindOrCreateBasket` to ensure baskets exist before saving outputs, matching TS `findOrInsertOutputBasket` behavior

## Test plan
- [x] Existing tests pass (`go test ./...`)
- [ ] Create outgoing tx (sendBsv to own address), then internalize same tx for same user — must succeed
- [ ] Deploy to production, verify inscribe flow through yours-wallet works end-to-end

Fixes #817


mrz1836 (approved):
LGTM

PR #838, opened 2026-04-07
fix: validate merkle proofs against chaintracks before storage

shruggr:
## Summary

- Validate merkle proof roots against chaintracks before persisting in both the sync task and internalize path
- Uses existing `MerklePath.VerifyHex()` from go-sdk and `IsValidRootForHeight` from the `Services` interface — both were available but not wired into the storage pipeline
- Invalid proofs are rejected and retried via the existing attempt counting mechanism

## Test plan

- [ ] `go build ./...` passes
- [ ] `go test ./pkg/storage/ -run "TestSynchronizeTxStatuses|TestInternalize"` passes
- [ ] Verify sync cycle accepts valid proofs and rejects invalid ones
- [ ] Verify internalize path rejects transactions with mismatched merkle roots

### DIFF
--- a/pkg/storage/internal/actions/actions.go
+++ b/pkg/storage/internal/actions/actions.go
@@ -59,6 +59,7 @@ func New(
 			beefVerifier,
 			scriptsVerifier,
 			services,
+			services,
 		),
 		process: newProcessAction(
 			ctx,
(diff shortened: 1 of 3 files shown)


shruggr:
Closing — validation belongs in the WalletServices MerklePath() implementations in 1sat-stack, not in the storage layer.

PR #849, opened 2026-04-18
fix: remove omitempty from Satoshis field to preserve zero values

shruggr:
Fix: Remove omitempty from Satoshis field

The omitempty tag caused zero-value satoshi amounts to be stripped from JSON responses, breaking clients that expect the field to always be present. This affected transactions with OP_RETURN outputs (satoshis:0) when using remote storage over JSON-RPC.

Change: Remove omitempty from ValidCreateActionOutput.Satoshis


mrz1836 (approved):
LGTM

PR #858, opened 2026-04-23
feat: add configurable change basket

dzolt-4chain:
## What Changed
Three values that were previously hardcoded are now configurable:

  | Value | Default | What it controls |
|---|---|---|
| NumberOfDesiredUTXOs | 32 | How many UTXOs the wallet targets per user's change basket |
| MinimumDesiredUTXOValue | 1000 sats | Target size of each UTXO - drives how change is split |
| MaxChangeOutputsPerTx | 8 | Max change outputs created in a single transaction (server-wide cap) |

## Why It Was Necessary
Previously NumberOfDesiredUTXOs (32), MinimumDesiredUTXOValue (1000 sats) and MaxChangeOutputsPerTx (8) were hardcoded constants in the library. There was no way to configure them without modifying the source code and cutting a new release.

For high-throughput use cases like a faucet server, 32 desired UTXOs is far too low - the wallet can't keep up with demand and runs out of spendable outputs. The cap of 8 change outputs per transaction also means a large UTXO pool takes a long time to build up. These values need to be tunable per deployment without a code change.

## Testing Performed
6 new tests were added across 3 packages:

| Test | What it verifies |
|---|---|
| TestWithChangeBasket_SetsDefaultsForNewUsers | WithChangeBasket option values are written to DB when a new user is created |
| TestSetDefaultChangeBasket_AffectsOnlySubsequentNewUsers | Runtime setter applies to new users only; existing users retain their DB values |
| TestUpdateChangeBasket_UpdatesSpecificUserInDB | Per-user update changes only the target user; other users are unaffected |
| TestUpdateChangeBasket_ReturnsErrorForUnknownUser | Calling UpdateChangeBasket with an unknown identity key returns an error |
| TestFunderSQLFund/cap:_SetMaxChangeOutputsPerTx | Calling SetMaxChangeOutputsPerTx at runtime is respected on the very next Fund() call |
| TestChangeBasketEnvVars | All three env vars are correctly parsed and loaded into infra.Config |

## Impact / Risk
No breaking changes. All three values default to their original hardcoded values (32 / 1000 / 8), so any deployment that does not set the new options behaves
  identically to before.

  No database migration required. The schema is unchanged - NumberOfDesiredUTXOs and MinimumDesiredUTXOValue were already stored per-user in the output_baskets table.
  Existing rows are untouched.

  Existing users in DB are unaffected. Changing the defaults only applies to users created after the change. Existing users keep their current basket values unless
  UpdateChangeBasket is called explicitly for them.

  Thread safety. MaxChangeOutputsPerTx is stored as an atomic.Uint64 on the funder struct and defaultChangeBasket as an atomic.Pointer on the provider - both are safe to
   read and write concurrently without locks.

  One removed symbol. The package-level var funder.MaxChangeOutputsPerTransaction has been removed. Any code outside this library that was reading or writing that
  variable directly will fail to compile. Based on the codebase this was not used externally, but should be confirmed before merging.


mrz1836 (changes requested):
Please remove the credentials and secrets that leaked into the code in this PR. Also the CI is not passing.

pawellewandowski98 (approved):

mrz1836 (approved):
LGTM

PR #873, opened 2026-05-01
[codex] fix security review findings

sirdeggen:
## Summary
- Replace all-origin/all-method/all-header CORS with intentional storage CORS defaults: all browser origins and private-network preflight are allowed, while methods, request headers, and exposed response headers are limited to Authrite/payment protocol needs.
- Add default 1 MiB request body caps for storage and certifier HTTP surfaces, including 413 handling for oversized certificate signing requests.
- Remove certifier private-key material from startup errors and validate certifier ports.
- Validate certificate issuance field values as base64 before signing, while preserving BRC-29 whitespace compatibility and existing nonce HMAC key ID behavior.
- Fix monitor-disabled startup, propagate infra context into Chaintracks startup, correct logging config error text, and harden example Postgres settings.

## Validation
- `go test ./...`
- `go test ./pkg/server ./pkg/storage ./pkg/infra ./pkg/brc29 ./pkg/certificates ./pkg/wallet`


mrz1836 (approved):
TLDR LGTM

PR #882, opened 2026-05-12
feat: keep inputs in create action in order

dzolt-4chain:
## What Changed
- Updated create.resultInputs(...) in storage create action flow to preserve user-provided input order exactly as passed in args.
- Removed class-based regrouping of provided inputs (unknown -> known) and replaced it with sequential processing over original xinputs order.
- Kept funded storage inputs appended after provided inputs, preserving existing funding behavior.
## Why It Was Necessary
- Go implementation diverged from ts-toolbox behavior by reordering provided inputs based on storage knowledge.
- This created inconsistent VIN ordering across implementations and surprising behavior for callers.
- Aligning with TS ensures deterministic, caller-expected input order and cross-SDK parity.
## Testing Performed
- Ran focused storage test:
  - go test ./pkg/storage -run TestCreateActionWithProvidedUnknownInput -count=1
- Result: pass.
## Impact / Risk
- Behavioral change: VIN assignment for mixed known/unknown provided inputs now follows request order instead of internal grouping.
- Potential side effects: tests or client logic that implicitly expected grouped ordering may need updates.
- Low protocol risk: transaction validity is unaffected; this is ordering/predictability alignment.

### DIFF
--- a/pkg/internal/testabilities/testservices/fixture_arc.go
+++ b/pkg/internal/testabilities/testservices/fixture_arc.go
@@ -279,23 +279,26 @@ type arcQueryFixture struct {
 }
 
 func (a *arcQueryFixture) WillReturnTransactionOnHeight(height int) {
-	tx := a.knownTransaction()
-	tx.status = "MINED"
-	tx.blockHeight = must.ConvertToUInt32(height)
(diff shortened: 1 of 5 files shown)


mrz1836 (approved):
LGTM

PR #889, opened 2026-05-15
fix: validate single leaf tx in BEEF

kuba-4chain:
## The Bug

The current code only iterates inputs of **unmined** transactions when computing in-degree:

```go
// txsNotMined we are not interested in inputs of mined transactions
txsNotMined := seq.Filter(seq2.Values(idToTx), func(tx *transaction.BeefTx) bool {
    return tx.Transaction.MerklePath == nil
})
```

**The scenario:**

Consider a BEEF with:
- `TxA` — mined (has MerklePath), provides output consumed by `TxB`
- `TxB` — mined (has MerklePath), provides output consumed by `TxC` (subject)
- `TxC` — unmined, subject tx

In-degree calculation:
- `TxA` starts at 0. Since only unmined tx inputs are walked, **`TxB`'s input pointing to `TxA` is never counted**. `TxA` stays at in-degree 0.
- `TxB` starts at 0. `TxC`'s input pointing to `TxB` is counted (TxC is unmined). `TxB` gets in-degree 1.
- `TxC` starts at 0. Nobody points to it. Stays at 0.

Leaves (in-degree = 0): **both `TxA` and `TxC`** → `len(subjectTxs) == 2` → validation **incorrectly fails**.

## The Fix

1. Remove the `txsNotMined` filter — iterate over all txs when collecting inputs.
2. Change the panic on unknown input TXIDs to a `continue`/`return` — those are valid external anchors for mined transactions.

The comment `// we are not interested in inputs of mined transactions` was the wrong assumption. The correct invariant is: **the subject tx is the unique transaction in the DAG that no other transaction in the BEEF depends on as an input.**

### DIFF
--- a/pkg/internal/txutils/to_ef_hex.go
+++ b/pkg/internal/txutils/to_ef_hex.go
@@ -23,12 +23,10 @@ func ValidateSingleLeafTx(beef *transaction.Beef) error {
 	// inDegree will contain the number of transactions for which the given tx is a parent
 	inDegree := seq2.CollectToMap(seq2.MapValues(idToTx, func(tx *transaction.BeefTx) int { return 0 }))
 
-	// txsNotMined we are not interested in inputs of mined transactions
-	txsNotMined := seq.Filter(seq2.Values(idToTx), func(tx *transaction.BeefTx) bool {
-		return tx.Transaction.MerklePath == nil
-	})
-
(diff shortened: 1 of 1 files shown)


pawellewandowski98 (approved):

sirdeggen:
For posterity, this checks out. The PR ensures that the check doesn't panic when there's a txid referenced in the inputs of one of the transactions in the BEEF set, it just skips it. This would otherwise panic every time because there's always one anchor parent at least which references something not included in the BEEF.

PR #890, opened 2026-05-15
refactor(txutils): make ValidateSingleLeafTx clearer and faster

sirdeggen:
The previous implementation using the seq/seq2 iterator library was difficult to follow logically. Replaced the inDegree + heavy seq machinery with a simple two-pass "referenced parent txids" set approach.

This directly implements the documented definition of a "leaf" (a tx whose txid does not appear in any inputs of other txs in the BEEF) and is also faster (2-4x on realistic sizes) with far fewer allocations.

Added a micro-benchmark (old vs new) for regression testing.


mrz1836 (approved):
LGTM

kuba-4chain (approved):

PR #898, opened 2026-05-22
fix(assembler): preserve BumpIndex when assembling atomic BEEF

kuba-4chain:
### Problem

The BEEF returned by `SignAction()` encoded `BumpIndex = 0` for every parent (source) transaction, regardless of which BUMP actually proves it. When the BEEF contained 2+ BUMPs and the spent parent was proven by a BUMP at index > 0, the recipient deserialized the BEEF and attached the **wrong merkle path** to the parent. With a single BUMP the bug was silent because index 0 happened to be correct.

### Root cause

Two layers:

1. **Upstream** — `go-sdk` `Beef.MergeRawTx` (`transaction/beef.go`) sets `beefTx.Transaction.MerklePath` and `beefTx.DataFormat = RawTxAndBumpIndex` when a `bumpIndex` is provided, but **never assigns `beefTx.BumpIndex = *bumpIndex`**. The sibling `MergeTransactionWithTxid` does this correctly. The wire-format serializer encodes `tx.BumpIndex` directly, so the bug only surfaces after serialization.
2. **Toolbox trigger** — `mergeSourceTxIntoBEEF` in `pkg/internal/assembler/assembled_transaction.go` re-merged each input's `SourceTransaction` via `MergeRawTx` after `MergeBeef` had already inserted it correctly, clobbering the in-memory `BumpIndex` to 0.

### Fix

- Skip the redundant re-merge when the tx is already present in the dest BEEF with `RawTxAndBumpIndex`. This avoids the upstream bug without bumping the SDK version.
- An upstream patch will be filed separately against `bsv-blockchain/go-sdk` to add the missing assignment in `MergeRawTx`.

### Tests

- New regression test `TestAtomicBEEF_PreservesBumpIndexAcrossMultipleBumps` builds a 2-BUMP BEEF where the parent is proven by `BUMPs[1]`, calls `AtomicBEEF`, round-trips through `NewBeefFromAtomicBytes`, and asserts `BumpIndex == 1` and `MerklePath.BlockHeight == 2000` on the parent.
- `ExpectedBeefTransactionState` now has an optional `BumpIndex *int` so existing and future `AssertBEEFState` / `AssertAtomicBEEFState` callers can assert the field. Existing call sites are unchanged.

### Why existing tests missed it

- `AssertBEEFState` only checked `DataFormat`, not `BumpIndex`.
- All assertions inspected the in-memory `*Beef`, where `MerklePath` looks correct; the bug only manifests after `Bytes()` / `AtomicBytes()` round-trip.
- Fixtures used single-BUMP BEEFs where `0` was coincidentally correct.


dzolt-4chain (approved):

mrz1836 (approved):
LGTM

PR #899, opened 2026-05-25
feat: add conformance tests for get network and get public key

dzolt-4chain:
### `getNetwork` — all 5 vectors passing

  All 5 `wallet.brc100.getnetwork.*` vectors now run against a real wallet instance.

  **Note on network string format:** BRC-100 specifies `"mainnet"`/`"testnet"` on the wire,
   but this implementation returns `defs.BSVNetwork` values (`"main"`/`"test"`) which are
  stored in the database and used throughout the system. Changing the return value would be
   a breaking change requiring broader coordination. The conformance test configures the
  wallet with the correct chain for each vector and asserts the returned value matches the
  internal representation for that chain. This deviation is documented in the test.

  ---

  ### `getPublicKey` — 150/201 vectors passing, 51 skipped

  All 201 `wallet.brc100.getpublickey.*` vectors now run. Args (`protocolID`, `keyID`,
  `counterparty`, `privileged`) are unmarshalled directly into `sdk.GetPublicKeyArgs` using
   the SDK's own JSON decoders, and the derived public key is asserted exactly against the
  vector's expected value.

  **51 vectors skipped:** Vectors using protocol name `"app"` (3 characters) are skipped.
  The Go SDK enforces a minimum of 5 characters for protocol names; the TS reference
  implementation (`@bsv/sdk@2.0.14`) does not apply this restriction, so those vectors were
   generated with `"app"` as a valid protocol. This is a known Go vs TS divergence — not a
  vector authoring error. The affected vectors are those with `protocolID: [1, "app"]`.

### DIFF
--- a/pkg/wallet/brc100_conformance_test.go
+++ b/pkg/wallet/brc100_conformance_test.go
@@ -11,6 +11,7 @@ import (
 
 	brc100vectors "github.com/bsv-blockchain/go-wallet-toolbox/conformance/vectors/wallet/brc100"
 
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/defs"
 	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/wallet/internal/testabilities"
 )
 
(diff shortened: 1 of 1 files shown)


mrz1836 (approved):
LGTM

PR #902, opened 2026-05-27
feat: add conformance tests for list actions and list outputs

dzolt-4chain:
Added TestBRC100Conformance_ListActions (16 vectors) and
  TestBRC100Conformance_ListOutputs (144 vectors).

  - Vector 12 skipped (seekPermission=false): The TS reference
  implementation ignores seekPermission entirely — it is never
  read by WalletPermissionsManager or the storage layer. The Go
  implementation returns a validation error for
  seekPermission=false. The validation guard in
  validate_list_actions_args.go should probably be removed in a follow-up.
  - Vector 14 seeded: Expects totalActions: 1. Rather than
  skipping it, the test seeds the wallet via faucet +
  CreateAction. Two fields from the expected shape are not
  asserted:
    - status: vector says "completed" as a placeholder; test
  environment produces "unproven" (broadcast accepted but not
  mined).
    - labels: vector omits includeLabels: true, so this
  implementation omits them from the response (TS returns them by
   default).

### DIFF
--- a/pkg/wallet/brc100_conformance_test.go
+++ b/pkg/wallet/brc100_conformance_test.go
@@ -12,6 +12,7 @@ import (
 	brc100vectors "github.com/bsv-blockchain/go-wallet-toolbox/conformance/vectors/wallet/brc100"
 
 	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/defs"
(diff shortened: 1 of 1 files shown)


mrz1836 (approved):
LGTM

PR #906, opened 2026-06-03
feat: use logger methods with context

pawellewandowski98:
## What Changed
- Changed logger methods to use the one with context. Thanks to that If the user pass logger with custom handler which can get trace-id from context it can add it to the logs

## Why It Was Necessary
-

## Testing Performed
-

## Impact / Risk
-

## Notifications
- @username


rob-tkn:
This will help us to correlate logs from the wallet-toolbox client with the logs from wallet-toolbox, especially when there are many concurrent operations happening at once. 👍

mrz1836 (approved):
LGTM

PR #912, opened 2026-06-10
Prevent false double-spend verdicts from corrupting wallet UTXO state

sirdeggen:
## What Changed
- **`confirmDoubleSpends` verification step** (`pkg/storage/internal/actions/process_confirm_double_spends.go`): every aggregated `doubleSpend` broadcast verdict is now re-verified against the network (batched `GetStatusForTxIDs`, 3 attempts, 1s apart) before it becomes a terminal failure. Downgrade rules:
  - tx is known to the network (mempool/mined) → treated as **success** (the report was a false positive)
  - at least one broadcaster accepted the tx → retryable **serviceError** (picked up by SendWaiting)
  - bare "missing inputs" hint with no competing txs and no success → retryable **serviceError** (indistinguishable from propagation lag)
  - evidenced conflicts (mempool-conflict / competing txs) with no success and tx unknown → stays terminal `doubleSpend`
  - Wired into both the synchronous (`broadcastTxs`) and background (`BackgroundBroadcast`) paths. Mirrors `confirmDoubleSpend` in the TS wallet-toolbox (`src/storage/methods/attemptToPostReqsToNetwork.ts`).
- **WhatsOnChain: accepted broadcast no longer voided by the follow-up tx-info fetch** (`whatsonchain/broadcast.go`): a 429/500 (or not-yet-indexed) response from `/txs/status` after a successful `/tx/raw` post used to rewrite the whole result to `error`, discarding the success.
- **WhatsOnChain: client-side rate limiter** (`whatsonchain/service.go`, `defs/whats_on_chain.go`): all WoC requests pass through a token-bucket limiter. Defaults to **3 req/s** (WoC's limit without an API key); configurable via `whats_on_chain.requests_per_second` for API-key plans (e.g. 10).
- **Bitails: "missing inputs" classified as a double-spend hint** (`DoubleSpend=true`), consistent with WhatsOnChain — previously it aggregated to terminal `invalidTx`.

## Why It Was Necessary
Load testing of an enterprise wallet (chained txs broadcast seconds apart, hitting WoC 429 rate limits) produced false `double_spend` / `missing_inputs` failures:

1. WoC receives the **raw tx** — for a freshly-chained tx whose parent hasn't reached WoC's mempool yet it answers "missing inputs", which was classified as `DoubleSpend=true`.
2. In aggregation a single double-spend report outranks any number of successes, so a tx **accepted by ARC** was still marked terminally failed.
3. Failing the tx runs `RecreateSpentOutputs`, releasing its inputs back to `spendable=true` with no chain verification — the wallet then builds **real** double spends from already-spent outputs and the corruption cascades until the wallet is unusable, with no repair path.
4. Additionally, a successful WoC broadcast followed by a rate-limited `/txs/status` lookup was counted as a broadcast failure, and the extra per-broadcast status call doubled WoC traffic, feeding the 429 storm.

The TS wallet-toolbox guards against exactly this with `confirmDoubleSpend`; the Go port was missing that step.

## Testing Performed
- New storage-level reproduction harness `pkg/storage/provider_broadcast_double_spend_test.go` (real GORM storage + mocked ARC/WoC over the shared httpmock transport), covering: false double-spend rescued via network status; conflict with unknown status stays retryable and completes on a later `SendWaitingTransactions` retry; missing-inputs-only stays retryable; evidenced mempool conflict still fails terminally (regression guard). All four were red before the fix.
- New WoC unit tests: accepted broadcast survives 429/500 on the follow-up tx-info fetch; rate limiter throttling and limiter defaults (`rate_limit_test.go`, `rate_limit_internal_test.go`).
- Full test suite green (`go test ./...`), `go vet` clean. Test fixtures set a high `RequestsPerSecond` so suites aren't throttled; the limiter itself is covered by dedicated tests.

## Impact / Risk
- A genuinely double-spent tx that some broadcaster accepted converges to terminal failure on a later retry (when no service accepts it anymore) instead of immediately — corruption-free at the cost of delayed finality for that edge case.
- Verification adds up to ~2s + a batched status query in the broadcast path, only when a double-spend verdict needs confirmation.
- Default WoC throughput is now capped at 3 req/s client-side; deployments with an API key should set `whats_on_chain.requests_per_second` to their plan's limit.
- Known remaining gaps (follow-up candidates): no background repair task akin to TS `TaskReviewDoubleSpends`; `unfail` doesn't restore the spent state of inputs; `updateSingleTx` writes aren't guarded against failing an already-mined tx.

## Notifications
- @pawellewandowski98

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -27,6 +27,7 @@ require (
 	go.opentelemetry.io/otel/sdk v1.44.0
 	go.opentelemetry.io/otel/trace v1.44.0
 	go.uber.org/mock v0.6.0
+	golang.org/x/time v0.15.0
 	gopkg.in/yaml.v3 v3.0.1
 	gorm.io/datatypes v1.2.7
 	gorm.io/driver/mysql v1.6.0
@@ -257,7 +258,6 @@ require (
 	golang.org/x/telemetry v0.0.0-20260603145448-488200d49c17 // indirect
 	golang.org/x/term v0.43.0 // indirect
 	golang.org/x/text v0.37.0 // indirect
-	golang.org/x/time v0.15.0 // indirect
 	golang.org/x/tools v0.45.0 // indirect
 	golang.org/x/xerrors v0.0.0-20240903120638-7835f813f4da // indirect
 	gonum.org/v1/gonum v0.17.0 // indirect
--- a/pkg/defs/services.go
+++ b/pkg/defs/services.go
@@ -126,6 +126,7 @@ func DefaultServicesConfig(chain BSVNetwork) WalletServices {
 			},
 			RootForHeightRetryInterval: DefaultRootForHeightRetryInterval,
 			RootForHeightRetries:       DefaultRootForHeightRetries,
+			RequestsPerSecond:          DefaultWhatsOnChainRequestsPerSecond,
 		},
 		Bitails: Bitails{
 			Enabled:                    false, // NOTE: Bitails is disabled by default
--- a/pkg/defs/whats_on_chain.go
+++ b/pkg/defs/whats_on_chain.go
@@ -5,6 +5,10 @@ import (
 	"time"
 )
 
+// DefaultWhatsOnChainRequestsPerSecond is the WhatsOnChain rate limit for requests
+// without an API key.
+const DefaultWhatsOnChainRequestsPerSecond = 3
+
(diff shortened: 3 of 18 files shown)


mrz1836 (approved):
LGTM

PR #915, opened 2026-06-11
fix: mark created outputs as not spendable when tx fails

dzolt-4chain:
Outputs created by a transaction were left with spendable=true and a
null tx_id when the transaction failed (abort, broadcast rejection, or
sync circuit-breaker). This caused ListOutputs to return zero-value
outpoints, which downstream transaction builders would select as inputs,
producing "duplicate input" / "malformed outpoint" broadcast errors.

Add two symmetric repo methods:
- MarkCreatedOutputsAsNotSpendable(transactionID): called on every
  failure path (abort, process broadcast rejection, sync circuit-breaker,
  unfail cascade-to-invalid)
- MarkCreatedOutputsAsSpendableByTxID(txID): called on unfail success
  to restore outputs when a tx is recovered with a merkle path

Five call sites updated: abort.go, process.go, synchronize_tx_statuses.go,
process_unfail.go (both failure and success branches).

### DIFF
--- a/pkg/storage/internal/actions/process.go
+++ b/pkg/storage/internal/actions/process.go
@@ -691,6 +691,10 @@ func (p *process) updateSingleTx(
 				err = fmt.Errorf("failed to restore spent outputs for failed tx %s: %w", txID, err)
 				return sendWithResult, reviewActionResult, err
 			}
+			if err = p.outputRepo.MarkCreatedOutputsAsNotSpendable(ctx, id); err != nil {
+				err = fmt.Errorf("failed to mark created outputs as not spendable for failed tx %s: %w", txID, err)
+				return sendWithResult, reviewActionResult, err
+			}
 		}
 	}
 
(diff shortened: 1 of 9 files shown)


mrz1836 (approved):
LGTM

pkg/storage/internal/actions/process.go
```diff
@@ -691,6 +691,10 @@ func (p *process) updateSingleTx(
 				err = fmt.Errorf("failed to restore spent outputs for failed tx %s: %w", txID, err)
 				return sendWithResult, reviewActionResult, err
 			}
+			if err = p.outputRepo.MarkCreatedOutputsAsNotSpendable(ctx, id); err != nil {
+				err = fmt.Errorf("failed to mark created outputs as not spendable for failed tx %s: %w", txID, err)
+				return sendWithResult, reviewActionResult, err
+			}
```
kuba-4chain, line 697: When the previous `RecreateSpentOutputs` fails, the new outputs are still marked as spendable, causing an invalid state in the toolbox.

I think it should try to mark created outputs as not spendable even if `RecreateSpentOutputs` fails. Maybe even marking new outputs as not spendable should happen first.

PR #916, opened 2026-06-15
fix: prevent UTXO double-reservation under concurrent transactions

dzolt-4chain:
## What Changed
-

## Why It Was Necessary
-

## Testing Performed
-

## Impact / Risk
-

## Notifications
- @username


pkg/internal/storage/funder/sql.go
kuba-4chain: This is now unused anywhere outside tests. Could it be removed?

pkg/internal/storage/funder/sql.go
```diff
@@ -193,7 +220,7 @@ func (f *SQL) loadUTXOPool(ctx context.Context, userID int, basketName string, f
 
 	var allUTXOs []*models.UserUTXO
 	for {
-		utxos, err := f.utxoRepository.FindNotReservedUTXOs(ctx, userID, basketName, page, forbiddenOutputIDs, includeSending)
+		utxos, err := findFn(ctx, page, forbiddenOutputIDs, includeSending)
 		if err != nil {
```
kuba-4chain, line 198: Why add this abstraction here? You could just directly use `FindNotReservedUTXOsForUpdate()`. This is never used with anything else

pkg/internal/storage/repo/transactions.go
```diff
@@ -44,29 +48,71 @@ func (txs *Transactions) CreateTransaction(ctx context.Context, newTx *entity.Ne
 		tracing.EndTracing(span, err)
 	}()
```
kuba-4chain, line 49: This method (without the DB tx) is only used in `Internalize()`, which also needs to happen in a DB tx. So in the future PR (not this one in order to keep the scope small), we could make this to accept the `tx *gorm.DB` by default, so that this always happens inside a DB tx, and remove the `CreateTransactionInTx()`, right?

pkg/internal/storage/repo/utxos.go
```diff
@@ -44,34 +45,78 @@ func (u *UTXOs) FindNotReservedUTXOs(
...
+// buildNotReservedQuery constructs the shared base query for both FindNotReservedUTXOs variants.
+// When forUpdate is true, SELECT FOR UPDATE SKIP LOCKED is appended (Postgres/MySQL only).
+func (u *UTXOs) buildNotReservedQuery(
+	db *gorm.DB,
+	userID int,
+	basketName string,
+	page *queryopts.Paging,
+	forbiddenOutputIDs []uint,
+	includeSending bool,
+	forUpdate bool,
```
kuba-4chain: I don't think you need this helper method for `FindNotReservedUTXOs()` and `FindNotReservedUTXOsForUpdate()` because the `FindNotReservedUTXOs()` is not used anymore. So you could just remove that method and use this `buildNotReservedQuery` directly in the `FindNotReservedUTXOsForUpdate()`, right?

pkg/storage/internal/actions/create.go
```diff
@@ -272,127 +285,146 @@ func (c *create) Create(ctx context.Context, userID int, params CreateActionPara
...
-		if adjSats < 0 {
-			return nil, errfunder.ErrNotEnoughFunds
+	err = c.db.WithContext(ctx).Transaction(func(dbTx *gorm.DB) error {
+		findFn := func(ctx context.Context, page *queryopts.Paging, forbiddenOutputIDs []uint, includeSending bool) ([]*models.UserUTXO, error) {
+			return c.utxoRepo.FindNotReservedUTXOsForUpdate(ctx, dbTx, userID, basket.Name, page, forbiddenOutputIDs, includeSending)
+		}
+		var fundErr error
+		funding, fundErr = c.sqlFunder.FundWithFinder(ctx, targetSat, initialTxSize, outputCount, basket, userID, processedInputs.ChangeOutputIDs, priorityOutputs, includeUTXOsInSendingState, isSweep, existingUTXOs, findFn)
+		if fundErr != nil {
+			return fmt.Errorf("funding failed: %w", fundErr)
```
kuba-4chain: I don't think this abstraction here is needed (see my comment in `sql.go` above). You could just do `FindNotReservedUTXOsForUpdate` directly inside the `FundWithFinder()` (so actually, just modify and use the old `Fund()`, because we don't need the finderFn, do we?

pkg/internal/storage/funder/sql.go
kuba-4chain: Actually, see my comment at the bottom, I think we could reuse this and remove the new `FundWithFinder`

pawellewandowski98 (approved):

kuba-4chain (approved):

PR #919, opened 2026-06-23
feat: use db transactions

pawellewandowski98:
## What Changed
- Introduced the `UnitOfWork` pattern to ensure atomicity across several composite wallet actions.
- Wrapped database mutations within `uow.Do(...)` blocks to ensure they share a single database transaction context. 
- Refactored the following actions to utilize `UnitOfWork`:
  - **`InternalizeAction`** (`internalize.go`): Bundled complex operations (checking for transaction existence, inserting new records, creating baskets, inserting transaction labels, and tracking internal output arrays) strictly into one atomic database transaction.
  - **`abortTx`** (`abort.go`): Rolled up UTXO un-reservation, recreating spent outputs, marking created outputs as unspendable, and marking transaction statuses as `failed` into a single atomic action block.
  - **`reviewKnownTxStatuses`** (`synchronize_tx_statuses.go`): Packaged the review loop handling cascades down to local transactions mapping to externally failed KnownTx records inside individual atomic queries for safety.
  - **`ProcessAction`** (`process.go`): Wrapped transaction status and state progression logic.
  - **`unfailSingle` / `markAsUnminedAndUnproven`** (`process_unfail.go`): Protected cascading state rollbacks and un-fail transitions within their own transactional boundaries.

## Why It Was Necessary
- Several core wallet actions rely on executing multiple database operations across different repositories (e.g., modifying UTXO pools, inserting/updating transaction records, managing labels and baskets).
- Previously, these composite actions did not run in guaranteed database transactions. If an operation crashed or failed mid-execution, it could leave the database in an inconsistent state (such as orphan UTXOs, out-of-sync statuses, or incomplete processing).
- Using the `UnitOfWork` pattern resolves this by providing all-or-nothing rollback safety, ensuring the integrity and stability of the wallet's local storage layer.

## Testing Performed
- Ran the full module test suite (`go test ./...`) for the `go-wallet-toolbox/pkg/storage/...` package.
- Verified that wrapping repositories behind the `Providers` interface maintains accurate behavior for consumer interactions without introducing deadlocks or regressions.

## Impact / Risk
- **Impact**: Provides strong consistency guarantees. Partial state failures inside critical path functions are now impossible.
- **Risk**: Moderate. Wrapping multiple operations in transactions introduces longer-lived database locks (specifically for SQLite). We deliberately kept transaction scopes tight and excluded complex flows like `create` (which relies on external component dependencies like `CountUTXOs`) to minimize connection-pool deadlocks.


mrz1836 (approved):
LGTM

PR #926, opened 2026-07-01
fix: prevent AbortAbandoned from restoring UTXOs of ARC-submitted TXs

dzolt-4chain:
## Summary

- **Root cause:** When a `processAction` context cancels (e.g. 1-minute HTTP timeout)
  after the signed TX has been submitted to ARC but before `updateSingleTx` can record
  the result, the TX is left in `unprocessed` status with no indication it was already
  sent. `AbortAbandoned` then finds it, calls `RecreateSpentOutputs`, and frees the
  UTXOs. A concurrent TX grabs those UTXOs. ARC eventually mines the original. Double spend.

- **Fix 1 — `process.go`:** Call `MarkKnownTxsAsSubmitting` immediately before
  `PostFromBEEF`. This sets `ProvenTxReq = sending, was_broadcast = true` atomically
  before ARC ever sees the TX. Any subsequent context cancel leaves the TX in a state
  that `AbortAbandoned` skips, and `ApplyProofTimeouts` handles via the rebroadcast
  path rather than `invalid`.

- **Fix 2 — `abort_abandoned.go`:** Replace `FindTransactionIDsByStatuses` with
  `FindTransactionIDsForAbort`, which only returns `unprocessed` TXs where the
  corresponding `KnownTx` is still in `ProvenTxStatusUnprocessed` — i.e. they failed
  before BEEF building completed and ARC was never involved. Those are safe to abort.

## Known remaining issue

If the context cancels **during BEEF building** (before `MarkKnownTxsAsSubmitting` is
reached), the TX is left in `unprocessed + ProvenTxReq = unprocessed`. `AbortAbandoned`
will correctly treat it as safe to abort. However, if the client retries before
`AbortAbandoned` fires and that retry also gets cancelled after reaching ARC, the fix
kicks in. If `AbortAbandoned` fires first, the retry fails — the user's operation is
silently dropped and they must create a new action.

The root fix for that case is restructuring `processNewTx` to call `SpendTransaction`
only after BEEF is successfully built, so a BEEF build failure leaves the TX in
`unsigned` (always safe to abort) rather than `unprocessed`. That is a larger change
tracked separately.

### DIFF
--- a/pkg/internal/storage/repo/known_tx.go
+++ b/pkg/internal/storage/repo/known_tx.go
@@ -62,6 +62,32 @@ func (p *KnownTx) UpdateKnownTxStatus(ctx context.Context, txID string, status w
 	return updateKnownTxStatus(p.db.WithContext(ctx), txID, status, skipForStatuses, txNotes)
 }
 
+func (p *KnownTx) MarkKnownTxsAsSubmitting(ctx context.Context, txIDs []string) error {
+	var err error
+	ctx, span := tracing.StartTracing(ctx, "Repository-KnownTx-MarkKnownTxsAsSubmitting")
+	defer func() {
+		tracing.EndTracing(span, err)
+	}()
+
+	if len(txIDs) == 0 {
+		return nil
+	}
+
+	err = p.db.WithContext(ctx).
+		Model(&models.KnownTx{}).
+		Where("tx_id IN ?", txIDs).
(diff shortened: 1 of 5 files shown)


kuba-4chain (approved):

mrz1836 (approved):
LGTM

PR #931, opened 2026-07-12
fix: JSON wire-format compatibility with TypeScript storage servers

### LINKED ISSUE #776: [BUG] InputBEEF in ValidCreateActionArgs marshals as base64 — incompatible with TypeScript storage servers
**Describe the bug**
When using `storage.NewClient()` to connect a Go service to a TypeScript storage server (e.g. `storage.babbage.systems`), `CreateAction` calls that include `InputBEEF` fail silently. 

Root cause: `ValidCreateActionArgs.InputBEEF` is typed as `primitives.BEEF` which is `[]byte`. Go's `encoding/json` marshals `[]byte` as a base64 string. TypeScript storage servers expect a JSON number array `[0, 1, 2, ...]`. The storage server rejects or ignores the malformed field.

**To Reproduce**
1. Use `storage.NewClient(url, wallet)` pointing at any TypeScript storage server
2. Call `wallet.CreateAction()` with `InputBEEF` populated (i.e. spending an existing output)
3. The request body contains `"inputBEEF": "AAEC..."` (base64) instead of `"inputBEEF": [0, 1, 2, ...]`
4. Storage server rejects it, causing downstream SPV failures when the resulting BEEF is verified

**Expected behavior**
`InputBEEF` should serialize as a JSON number array to match the TypeScript wire format.

**Fix**
Change the field type in `pkg/wdk/storage_create_action_args.go`:

```go
// Before
InputBEEF primitives.BEEF `json:"inputBEEF,omitempty"`

// After  
InputBEEF primitives.ExplicitByteArray `json:"inputBEEF,omitempty"`

### PR DESCRIPTION
**Pull request by shruggr**
Fixes #776.

Two fields in the remote-storage wire format serialize in a shape TypeScript storage servers reject, so any `CreateAction` against a TS storage server that supplies explicit inputs fails.

### `primitives.BEEF` marshals as base64, not a number array
`BEEF` is `[]byte`, which Go's `encoding/json` marshals as a base64 string. The TS wallet-toolbox reads `inputBEEF` with `makeReader`, which requires a number array, so a `createAction` carrying `inputBEEF` fails with:

```
bin must be Uint8Array or number[]
```

Adds `MarshalJSON` (emits `[0,1,255]`, reusing the existing `ExplicitByteArray` encoder) and `UnmarshalJSON` (accepts both a number array and a base64 string).

### `OutPoint` has no JSON tags
`OutPoint{TxID, Vout}` serializes as `{"TxID":…,"Vout":…}`; the TS server reads `outpoint.txid`/`outpoint.vout`, so the txid arrives `undefined` and validation fails with:

```
inputBEEF ... must be valid and contain proof data for possibly known undefined
```

Adds `json:"txid"` / `json:"vout"` tags.

### Tests
Round-trip tests for `BEEF` (number array, base64 input, out-of-range rejection).

Reproduced against a live TypeScript storage server: a Go service calling `CreateAction` with explicit inputs + `inputBEEF` fails on both counts before this change and succeeds after.

### DIFF
--- a/pkg/storage/v1adapter/handler.go
+++ b/pkg/storage/v1adapter/handler.go
@@ -312,7 +312,7 @@ func (h *Handler) createAction(w http.ResponseWriter, r *http.Request) {
 	}
 
 	var args wdk.ValidCreateActionArgs
-	if err := json.Unmarshal(argsRaw, &args); err != nil { //nolint:musttag // wdk.OutPoint (nested) has no JSON tags; decoded by wdk package conventions.
+	if err := json.Unmarshal(argsRaw, &args); err != nil {
 		h.writeError(w, http.StatusBadRequest, "invalid args for createAction")
 		return
 	}
--- a/pkg/wdk/outpoint.go
+++ b/pkg/wdk/outpoint.go
@@ -5,9 +5,9 @@ import "fmt"
 // OutPoint identifies a unique transaction output by its txid and index vout
 type OutPoint struct {
 	// TxID Transaction double sha256 hash as big endian hex string
-	TxID string
+	TxID string `json:"txid"`
 	// Vout Zero based output index within the transaction
-	Vout uint32
+	Vout uint32 `json:"vout"`
 }
 
 func (o OutPoint) String() string {
--- a/pkg/wdk/primitives/beef.go
+++ b/pkg/wdk/primitives/beef.go
@@ -7,6 +7,8 @@ import (
 )
 
 // BEEF An array of integers, each ranging from 0 to 255, indicating transaction data in BEEF (BRC-62) format.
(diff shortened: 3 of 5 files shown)


mrz1836 (changes requested):
CI is failing, linter issue.

sirdeggen:
@mrz1836 The CI/linter failure you flagged has been addressed in `cc19064` (remove obsolete `musttag` nolint after OutPoint gained JSON tags, reuse `ExplicitByteArray` for BEEF number-array unmarshaling, add wire-format regression tests).

All checks are green on the latest HEAD (`cc19064`), including ✨ Lint Code, test matrix, postgres-tests, CodeQL, and SonarCloud Quality Gate.

Ready for re-review when you have a moment.

PR #935, opened 2026-07-18
feat: DR v1 Track P — bounded target-aware funder, contention retry, WAL, hardening (#933)

sirdeggen:
## Summary

Decision Record v1 — **Track P** (funder performance & remaining hardening). Follows Track S (#934): that wave closed the correctness findings from the UTXO-selection review; this wave closes the performance track and the remaining upheld findings.

Spec and plan are on the branch: `docs/superpowers/specs/2026-07-17-dr-v1-track-p-design.md`, `docs/superpowers/plans/2026-07-17-dr-v1-track-p.md`.

### Headline: bounded target-aware funder (P1-3)

The funder previously loaded — and on Postgres/MySQL FOR-UPDATE-locked — the **entire** eligible (user, basket) UTXO pool on every `createAction`, then best-fit selected 1–3 rows in memory. This was the dominant TPS ceiling and hard-serialized same-user funding to one in-flight call (surfacing as false `ErrNotEnoughFunds`).

It now issues per-allocation micro-queries that mirror the old `selectBest` semantics exactly per step (exact → smallest-sufficient via `ORDER BY satoshis ASC LIMIT 1`; largest-insufficient via `DESC LIMIT 16`; walked tier by tier, mined → unproven → sending), locking only rows actually considered:

| BenchmarkSQLFund | before | after | delta |
|---|---|---|---|
| pool_1000 | 4.40 ms/op · 23,021 allocs | 0.175 ms/op · 918 allocs | ~25x |
| pool_10000 | 56.1 ms/op · 231,690 allocs | 0.156 ms/op · 917 allocs | ~360x |

Cost is now flat in pool size. Same-user concurrent funding works; EXPLAIN confirms the new composite index serves all five predicates of both query shapes.

### Also in this PR

- **fix(funder):** `MinimumDesiredUTXOValue=0` caused an integer divide-by-zero panic inside the funding transaction (user-reachable via `UpdateChangeBasket`). Guarded at three layers (P2-4).
- **perf(storage):** composite `idx_user_utxos_selection` (user_id, basket_name, reserved_by_id, utxo_status, satoshis) + single-column `reserved_by_id` index; AutoMigrate, additive only (P2-1).
- **fix(repo):** `UserUTXOs.Update` could silently overwrite `reserved_by_id` with no precondition; now a CAS (`WHERE reserved_by_id IS NULL` + RowsAffected) that errors on contention (P2-5).
- **feat(storage):** public `wdk.ErrUTXOContention` sentinel + bounded retry (3 attempts, 25ms·attempt ±50% crypto-rand jitter, ctx-aware) around the funding transaction (P2-6).
- **fix(funder):** exhaustive (sweep) paging gets a unique `output_id` tiebreaker + dedup — OFFSET+SKIP LOCKED could re-return rows and trip spurious contention rollbacks (P2-2).
- **perf(storage):** file-based SQLite DSNs default to `_journal_mode=WAL&_busy_timeout=5000`; operator DSN params win; in-memory untouched (P2-3). Faucet example moved its database into a mounted `data/` directory (single-file bind mount loses committed-but-not-checkpointed WAL transactions).
- **test:** `BenchmarkSQLFund` baseline (P2-11); Postgres change-path concurrency test — 4 change UTXOs, 8 concurrent CreateActions, exactly 4 succeed, distinct reservations, invariants hold (P1-4 residual); direct `reserveUTXOs` contention-branch test (P2-9); jitter bounds test.
- **docs:** MySQL minimum 8.0.1 (SKIP LOCKED), connection-pool default 5 with production guidance, WAL operational notes (P2-7, NIT-1).

### Release notes / operational impact

1. **Schema:** AutoMigrate builds two new indexes on `bsv_user_utxos` at first startup after upgrade — expect build time on large tables.
2. **SQLite behavior change:** WAL is now the default for file DSNs. `-wal`/`-shm` sibling files appear next to the database — persist/mount the directory, not the single file. `journal_mode` persists inside the db file across reopens. Explicit `_journal_mode`/`_busy_timeout` DSN params disable the defaulting.
3. **New public API:** `wdk.ErrUTXOContention`. CRUD `UserUTXO.Update` setting `ReservedByID` on an already-reserved row now errors instead of silently overwriting.
4. **Latency profile:** `CreateAction` under UTXO contention retries up to 2 extra times (worst ~112ms added) before surfacing an error; emits a "retrying funding after UTXO contention" warn per retry.
5. **Faucet example:** database path moved to `./data/storage.sqlite` — existing example deployments must move their file or start fresh.

### Accepted-risk notes (deliberate, documented in-code/ledger)

- Up to 15 fetched-but-unconsumed batch rows stay locked until commit — bounded residual, strictly smaller than the old whole-pool lock; surfaces as false `ErrNotEnoughFunds` under extreme same-user contention (not retried — retry is contention-only, so genuine insufficient-funds isn't taxed).
- Non-sweep funding now has a theoretical ~32k-input ceiling from `NOT IN` bind params (sweep path unaffected).
- Follow-up candidate: append `output_id` as 6th column of the composite index for a pure ordered index walk.

### Verification

Every task implemented + adversarially reviewed (core funder task reviewed under EXPLAIN ANALYZE on a 200k-row replica). Full SQLite suite, full Postgres suite (`TEST_DB_MODE=postgres -p 1 -race` in CI; locally multiple full runs), `golangci-lint` clean repo-wide. Concurrency tests stable across repeated runs (`-count=5`).

Still open from DR v1 (next waves, unchanged by this PR): W2-2 cascade scoping, W2-3 reorg demote, W3a-2 abort row leases, MySQL testmode/CI.


pkg/internal/storage/repo/user_utxo.go
```diff
@@ -82,12 +82,21 @@ func (r *UserUTXOs) Update(ctx context.Context, spec *entity.UserUTXOUpdateSpeci
...
+	if spec.ReservedByID != nil {
+		query = query.Where(table.ReservedByID.IsNull())
+	}
+
+	var info gen.ResultInfo
+	info, err = query.Updates(updates)
 	if err != nil {
 		return fmt.Errorf("failed to update user utxo: %w", err)
 	}
+	if spec.ReservedByID != nil && info.RowsAffected == 0 {
```
sirdeggen, line 104: **[suggestion]** The CAS path treats every zero-`RowsAffected` update that sets `ReservedByID` as `ErrUTXOContention`. That is correct for "already reserved," but the same branch also fires when `OutputID` does not exist (or does not match any row). Callers that `errors.Is(..., wdk.ErrUTXOContention)` — including any future retry wrapper — cannot distinguish "row missing" from "lost a race," and may waste retries on a permanent not-found condition.

**Suggestion:** When `RowsAffected == 0`, either (a) re-read the row and return a distinct not-found / not-owned error if absent, or (b) document explicitly that missing rows are surfaced as contention by house rule, and keep retries scoped only to create-path writers that cannot hit a missing ID.

pkg/internal/storage/repo/utxos.go
```diff
@@ -83,6 +87,97 @@ func (u *UTXOs) FindNotReservedUTXOsForUpdate(
...
+
+	query := tx.WithContext(ctx).Scopes(
+		scopes.UserID(userID),
+		scopes.BasketName(basketName),
+		notReserved(),
+		outputNotIn(excludedOutputIDs),
+	).Where("utxo_status = ?", string(status)).
+		Where("satoshis < ?", maxSatoshis).
+		Order("satoshis DESC, output_id DESC").
+		Limit(limit)
```
sirdeggen, line 174: **[suggestion]** `FindLargestInsufficientUTXOsForUpdate` passes `limit` straight into GORM’s `Limit(limit)` with no lower-bound check. In GORM v2, `Limit(0)` is a known footgun (often interpreted as "no limit"). The funder always passes `insufficientBatchSize = 16` today, so production is safe, but the repository method is part of the shared interface and a future caller (or test) with `limit <= 0` could reintroduce whole-tier locking.

**Suggestion:** Reject or clamp `limit` (`if limit <= 0 { return nil, fmt.Errorf(...) }` or `limit = max(limit, 1)` and cap to a hard max). Mirror the same guard if other bounded helpers are added later.

pkg/internal/storage/database/dialectors.go
```diff
@@ -33,6 +35,46 @@ func sqliteDialector(cfg defs.Database) gorm.Dialector {
...
+// handing the bare path to sqlite3_open_v2. So appending params to a bare
+// path (no "file:" prefix) works exactly like appending them to a "file:" DSN.
+//
+// Operator-provided params always win: this only fills in defaults when the
+// DSN is silent on both pragmas, so any explicit _journal_mode or
+// _busy_timeout (or an in-memory DSN, where WAL is meaningless and disk-only)
+// is left untouched.
+func sqliteDSNWithDefaults(dsn string) string {
+	switch {
+	case strings.Contains(dsn, ":memory:"),
```
sirdeggen: **[suggestion]** `sqliteDSNWithDefaults` leaves the DSN untouched if **either** `_journal_mode` **or** `_busy_timeout` is already present. Operators who already pin only `_busy_timeout=…` (a common pattern when tuning lock waits) will not receive the new WAL default on upgrade; they keep whatever journal mode is already stored in the file (often DELETE). Docs mention "either," but the upgrade behavior is easy to miss and can leave production file DBs without the intended concurrency win.

**Suggestion:** Prefer independent defaulting: append only the missing pragma(s) (`_journal_mode` if absent, `_busy_timeout` if absent), still never overriding explicit values. If the either-or policy is intentional, call it out more loudly in release notes (e.g. "any existing `_busy_timeout` disables WAL auto-enable").

pkg/storage/internal/actions/create.go
```diff
@@ -37,8 +39,53 @@ import (
...
+// number (see backoffWithJitter). It is a package-level var, rather than a const, so tests
+// can shrink it to keep retry tests fast without adding public configuration.
+var fundingRetryBaseBackoff = 25 * time.Millisecond
+
+// retryOnContention runs fn up to maxFundingAttempts times, retrying only when fn returns an
+// error that wraps wdk.ErrUTXOContention (i.e. a concurrent transaction reserved one of the
+// UTXOs this attempt selected). Any other error, or success, returns immediately after the
+// first attempt. Between retries it waits backoffWithJitter(attempt, random), aborting early
+// if ctx is canceled while waiting.
+func retryOnContention(ctx context.Context, logger *slog.Logger, random wdk.Randomizer, fn func() error) error {
```
sirdeggen, line 63: **[suggestion]** `retryOnContention` retries any path that surfaces `wdk.ErrUTXOContention`, including permanent provided-input double-claims (`markReservedOutputsAsNotSpendable` RowsAffected mismatch). Losers of concurrent same-provided-outpoint races now pay up to two backoffs (~25ms·attempt ±50%) before failing, and an intermediate attempt can also re-run `Fund` and briefly SKIP LOCK / re-select change UTXOs. During that window the change pool can be drained by others, so the final error may flip from contention to `ErrNotEnoughFunds`, which is funds-safe but harder to diagnose.

**Suggestion:** Acceptable as a deliberate tradeoff for change-path contention recovery; if latency/diagnostics matter, consider not retrying when `ReservedOutputIDs`/`SpentOutputIDs` include user-provided known outputs that failed the spent_by CAS, or map that branch to a non-retried sentinel. At minimum, log which contention site failed (reserve vs claim) on the final error.

pkg/internal/storage/funder/sql_bench_test.go
```diff
@@ -0,0 +1,97 @@
...
+	"testing"
+
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/entity"
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/internal/fixtures/testusers"
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/internal/satoshi"
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/internal/storage/funder/testabilities"
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/wdk"
+)
+
+// BenchmarkSQLFund is the baseline benchmark for the current whole-pool-loading
```
sirdeggen: **[nit]** Benchmark comment still describes "the current whole-pool-loading funder.SQL.Fund implementation, captured before Task 7." After T7 the benchmark measures the bounded funder, so the comment is stale and misleading for future perf comparisons.

**Suggestion:** Rewrite the comment to say it now tracks post–Track-P bounded selection (and optionally note that T5 numbers remain the historical baseline in the SDD ledger).

pkg/internal/storage/funder/sql.go
```diff
@@ -101,36 +132,121 @@ func (f *SQL) Fund(
...
+			// result and wdk.ErrNotEnoughFunds.
+			return nil
 		}
 
-		utxo := pool.selectBest(uint64(remaining))
-		if utxo == nil {
-			return nil, wdk.ErrNotEnoughFunds
+		allocated := false
+		for _, tier := range tiers {
+			utxo, err := f.utxoRepository.FindSmallestSufficientUTXOForUpdate(ctx, tx, userID, basketName, tier, uint64(remaining), excluded)
```
sirdeggen: **[nit]** After `remaining <= 0` is ruled out, the code uses `uint64(remaining)` on a signed `satoshi.Value`. Safe given the guard, but inconsistent with the package’s own `MustUInt64` / `UInt64` helpers and slightly easier to break if the control flow changes later.

**Suggestion:** Use `remaining.MustUInt64()` (or a local checked conversion) at both the sufficient and insufficient call sites (lines 193 and 206).

PR #937, opened 2026-07-18
fix: persist KnownTx notify opaque payload for sync (#851)

### LINKED ISSUE #851: UpsertKnownTxForSync: notify field hardcoded, opaque payload lost on round-trip
## Problem

`pkg/internal/storage/repo/syncrepo/sync_knowntx.go` at approximately line 199 hardcodes `Notify: "{}"` with a TODO comment, regardless of the value in the incoming `ProvenTxReq.notify` field. The TypeScript client uses this opaque blob to track which transactions to notify when proofs arrive.

## RPC-Observable Difference

Any `ProvenTxReq` that round-trips through the Go server loses its notify payload.

## Suggested Fix

Add a `notify` string column to the KnownTx model. Persist the incoming value unchanged. Return it unchanged on reads.

## Severity

Low — single-device impact minimal. Tracked for completeness.

sirdeggen:
## Summary
- Add `notify` text column on `KnownTx` (GORM AutoMigrate) to store the opaque ProvenTxReq.notify blob used by the TypeScript wallet.
- Wire notify through entity, chunk ingest (`upsertProvenTxReqs`), `UpsertKnownTxForSync`, and `FindKnownTxsForSync` so the payload round-trips unchanged.
- Empty notify defaults to `"{}"` for backward compatibility with the previous hardcoded value.

Fixes #851

## Test plan
- [x] `go test ./pkg/internal/storage/repo/syncrepo/ -run TestKnownTxNotify -count=1`
- [x] `go test ./pkg/internal/storage/repo/syncrepo/ -count=1`
- [ ] CI green on PR

### Local coverage
- Upsert persists opaque notify payload
- Empty notify defaults to `{}`
- FindKnownTxsForSync returns stored notify
- Newer updated_at update replaces notify payload

### DIFF
--- a/pkg/entity/known_tx.go
+++ b/pkg/entity/known_tx.go
@@ -19,6 +19,9 @@ type KnownTx struct {
 	Notified            bool
 	WasBroadcast        bool
 	RebroadcastAttempts uint64
+	// Notify is an opaque payload (typically JSON) from ProvenTxReq.notify.
+	// Persisted and returned unchanged for sync round-trip; no Go-side semantics.
+	Notify string
 
 	RawTx     []byte
 	InputBEEF []byte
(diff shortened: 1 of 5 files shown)


sirdeggen:

PR #938, opened 2026-07-18
fix: listOutputs includeLabels populates output labels

### LINKED ISSUE #820: listOutputs: includeLabels parameter accepted but not implemented
## Problem

`listOutputs` accepts `includeLabels` as a parameter but does not implement it. There is a TODO comment in the code:

```go
// TODO: Handle args.IncludeLabels
```

When a caller requests `includeLabels: true`, the response includes outputs with empty label arrays instead of the actual labels.

## RPC-Observable Difference

**TypeScript:** Returns populated `labels` array on each output when `includeLabels: true`
**Go:** Always returns empty labels regardless of the parameter value

## Impact

Callers that depend on output labels for filtering, display, or classification get incomplete data.

## Location

The TODO is in the listOutputs mapping/implementation code. The TypeScript implementation fetches labels via join queries on the `TxLabel`/`TxLabelMap` tables.

sirdeggen:
## Summary
- `listOutputs` already fetched labels when `includeLabels` was true, but still had a stale TODO and **swallowed** label-fetch errors (leaving empty labels).
- Refactored label loading into `loadLabelsIfNeeded`, removed the TODO, and return errors from `GetLabelsForTransactions`.
- Outputs inherit labels from their parent transaction (TxLabel / transaction_labels), matching TypeScript behavior.

Fixes #820

## Test plan
- [x] `go test ./pkg/storage/ -run 'TestListOutputs_IncludeLabels|TestListOutputs_MinimalFilter|TestListOutputs_IncludeTags' -count=1`
- [x] `go test ./pkg/wallet/ -run 'ListOutputs' -count=1`
- [x] `go test ./pkg/storage/internal/actions/ -count=1`
- [ ] Verify `includeLabels: true` returns populated labels on each output
- [ ] Verify `includeLabels: false` / omitted leaves labels empty/nil

### DIFF
--- a/pkg/storage/internal/actions/list_outputs.go
+++ b/pkg/storage/internal/actions/list_outputs.go
@@ -41,7 +41,6 @@ func newListOutputs(logger *slog.Logger, outputsRepo OutputRepo, knownTxRepo Kno
 
 func (l *listOutputs) ListOutputs(ctx context.Context, auth wdk.AuthID, args *wdk.ListOutputsArgs) (*wdk.ListOutputsResult, error) {
 	// TODO: Handle args.KnownTxids
-	// TODO: Handle args.IncludeLabels
 
 	userID := *auth.UserID
 	var err error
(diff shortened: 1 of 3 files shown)


sirdeggen:

PR #940, opened 2026-07-18
fix: listOutputs knownTxids BEEF optimization (#821)

### LINKED ISSUE #821: listOutputs: knownTxids parameter accepted but not implemented
## Problem

`listOutputs` accepts `knownTxids` as a parameter but does not use it for BEEF optimization. There is a TODO comment in the code:

```go
// TODO: Handle args.KnownTxids
```

## RPC-Observable Difference

When `includeTransactions: true`, the TypeScript implementation uses `knownTxids` to exclude transactions the caller already has from the returned BEEF, reducing response size. The Go implementation ignores this parameter and always returns full BEEF.

## Impact

Responses are larger than necessary. For callers with many known transactions, this is a performance issue but not a correctness issue.

sirdeggen:
## Summary
- Confirmed storage `listOutputs` already passes `args.KnownTxids` into `GetBEEFForTxIDs` via `entity.WithKnownTxIDs`, matching TS `mergeTxidOnly` behavior when `includeTransactions` is true.
- Removed obsolete `TODO: Handle args.KnownTxids` and `TODO: Handle args.IncludeLabels` (both are implemented).
- Added `TestListOutputs_KnownTxids` demonstrating known parents become `TxIDOnly`, ancestor recursion stops, and BEEF payload size shrinks.

## Test plan
- [x] `go test ./pkg/storage/ -run 'TestListOutputs_' -count=1`
- [ ] Reviewer: spot-check BEEF with/without `knownTxids` on a funded wallet

Fixes #821

### DIFF
--- a/pkg/storage/internal/actions/list_outputs.go
+++ b/pkg/storage/internal/actions/list_outputs.go
@@ -40,9 +40,6 @@ func newListOutputs(logger *slog.Logger, outputsRepo OutputRepo, knownTxRepo Kno
 }
 
 func (l *listOutputs) ListOutputs(ctx context.Context, auth wdk.AuthID, args *wdk.ListOutputsArgs) (*wdk.ListOutputsResult, error) {
-	// TODO: Handle args.KnownTxids
-	// TODO: Handle args.IncludeLabels
-
 	userID := *auth.UserID
 	var err error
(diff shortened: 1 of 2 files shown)


sirdeggen:

PR #956, opened 2026-07-18
[00/11] ci: skip postgres suite on docs/plan-only PRs

sirdeggen:
## Summary
Plan-only / docs-only PRs were still running the full `postgres-tests` workflow (~15 minutes) even though Fortress already short-circuits its heavy jobs via `paths-check`.

## Change
- Add a paths filter job: run the suite only when the PR changes `.go`, `go.mod`, `go.sum`, or this workflow.
- Keep the required check name `test-postgres` as a lightweight always-on rollup so branch protection stays green when the suite is skipped.
- Non-PR events (push to main) still run the full suite.

## Test plan
- [ ] This PR itself (workflow change) should **run** the postgres suite (workflow file is in the filter).
- [x] A pure `plans/*.md` PR should show `test-postgres` green in seconds without starting Postgres.

### DIFF
--- a/.github/workflows/postgres-tests.yml
+++ b/.github/workflows/postgres-tests.yml
@@ -7,10 +7,8 @@ on:
     branches: [main, master]
   pull_request:
 
-permissions:
-  contents: read
-  pull-requests: read # paths filter via gh api on pull_request
-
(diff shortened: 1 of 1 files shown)


mrz1836 (approved):
LGTM - but might be missing security hardening to this workflow

PR #929, opened 2026-07-09
fix: immediately abort tx adn release UTXOs on pre-broadcast failure

dzolt-4chain:
## What Changed
-

## Why It Was Necessary
-

## Testing Performed
-

## Impact / Risk
-

## Notifications
- @username

PR #961, opened 2026-07-21
feat: add `aborted` transaction status to distinguish pre-broadcast aborts from broadcast rejections (#959)

### LINKED ISSUE #959: Ambiguous status: failed broadcast or cancelled ctx?
We are currently unable to determine the difference between:

1. A createAction attempt which never made it to broadcast endpoints because ctx is cancelled.
2. A createAction attempt which hit broadcast endpoints and was explicitly rejected

They're both marked as status: failed today.

The proposal is to differentiate them by adding a tx status of aborted to example #1

sirdeggen:
## Problem (#959)

A transaction can end in the terminal status `failed` for two opposite reasons that were **indistinguishable**:

1. **Aborted before broadcast** — ctx cancellation (upstream HTTP timeout), a token-swap detail fetch timing out, the user calling `AbortAction`, or the Monitor `fail_abandoned` sweep. The tx **never reached a broadcast endpoint**; its inputs are released immediately (per #929), so it is safe to **rebuild and retry**.
2. **Broadcast rejection** — the tx **was posted** and rejected (double spend / invalidTx). Retrying is pointless; this is **permanent**.

Both wrote `wdk.TxStatusFailed` onto the per-user `Transaction` row, which carries no reason/kind column, and the abort path deliberately doesn't touch the shared `KnownTx` — so the Enterprise Wallet's idempotency-label lookup could not tell a retryable abort from a permanent rejection.

## Solution

Add a distinct terminal status **`wdk.TxStatusAborted = "aborted"`**, written **only** by the two pre-broadcast abort writers; every broadcast-rejection writer stays on `failed`. That single divergence lets the caller read the raw status and branch retryable-vs-permanent.

- **Emit `aborted`:** `abort.go` `abortTx` (user `AbortAction` + Monitor `fail_abandoned` sweep) and `process.go` `abortTxByStringID` (ctx-cancel / build-abort).
- **Stay `failed`:** `process.go` double-spend/invalidTx, `known_tx.go` `FailKnownTxAsDoubleSpend`, `synchronize_tx_statuses.go` terminal cascade, `process_unfail.go` `markAsInvalid`.
- **Read path:** the failed-actions listing bucket (`ListFailedActions` / the `unfail`-label spec-op) is widened to `{failed, aborted}` so aborted txs surface where they used to (they were `failed` before), carrying their **raw** status string. `markActionsForUnfail` is guarded to skip non-`failed` actions, so listing an aborted row with `unfail=true` can't spuriously flip its shared `KnownTx`.
- **Standardized surface:** `TxStatus.ToStandardizedStatus(aborted) → TxUpdateStatusFailed` (terminal, **not** `InvalidTx`), and `ListTransactions` treats `aborted` as terminal.
- **BRC-100:** `mapActionStatus` maps `aborted` via a local temp const (the go-sdk `ActionStatus` union lacks it — same pattern already used for `failed`), JSON-transport only.

## Design decisions

- **Read path:** raw status via `ListFailedActions`, not a new BRC-100 `ActionStatus`.
- **TS parity:** Go-only in this PR; `ts-stack` `TransactionStatus` + a BRC-40 conformance vector are a tracked follow-up (see below).
- **Scope:** all pre-broadcast aborts become `aborted` (they share the `abortTx` choke point and are all retryable).

## Not in scope / follow-ups

- **Mid-`PostFromBEEF` ctx-cancel** leaves the tx in `submitting/sending` (it may have reached the network — unsafe to call `aborted`). Unchanged.
- **`ts-stack` parity follow-up:** add `aborted` to the TS `TransactionStatus` union + a `parity_class:'intended'` conformance vector, promote to `required` once both impls pass. `TxStatus` crosses the BRC-40 sync wire verbatim; a Go wallet emitting `aborted` to an older TS peer relies on that peer tolerating an unknown status.
- **No DB migration** — `Transaction.Status` is a free-form string column. Historical `failed`-that-were-aborts cannot be retroactively reclassified (accepted).

## Testing

- `TxStatus` mapping table with a coverage guard (`ToStandardizedStatus`/`ToUTXOStatus`).
- Abort paths write `aborted`; re-abort refused via the public API.
- **Regression locks:** broadcast rejection stays `failed`; a never-broadcast aborted tx is not swept to `failed` by the failure-review reconciliation; `ListTransactions` reports a `nosend`-then-aborted tx as terminal.
- `ListFailedActions` returns both `failed` and `aborted` with distinct raw statuses; the unfail guard is proven (test fails if the guard is removed).
- `mapActionStatus` round-trips `aborted`.
- Full suite: `go test ./pkg/...` green (49 packages), `golangci-lint run` 0 issues.

Design spec and task-by-task plan are included under `docs/superpowers/`.

Fixes #959


pkg/storage/provider.go
```diff
@@ -1284,7 +1284,7 @@ func (p *Provider) ListTransactions(ctx context.Context, auth wdk.AuthID, args w
 	transactions := make([]wdk.CurrentTxStatus, 0, len(knownTxs))
 	for _, ktx := range knownTxs {
 		status := ktx.Status.ToStandardizedStatus()
-		if txStatusMap[ktx.TxID] == wdk.TxStatusFailed {
+		if s := txStatusMap[ktx.TxID]; s == wdk.TxStatusFailed || s == wdk.TxStatusAborted {
 			status = wdk.TxUpdateStatusFailed
 		}
```
kuba-4chain, line 1289: `ListTransactions` is the method used to ensure idempotency of operations, because this is the only one returning the transaction data. The `Aborted` "standardized" status needs to be surfaced specifically here for the end user to retry the operation from scratch.

pkg/wdk/tx_status.go
```diff
@@ -50,6 +57,8 @@ func (s TxStatus) ToStandardizedStatus() StandardizedTxStatus {
 		return TxUpdateStatusWaiting
 	case TxStatusFailed:
 		return TxUpdateStatusInvalidTx
+	case TxStatusAborted:
+		return TxUpdateStatusFailed
```
kuba-4chain, line 61: The `aborted` status needs to be added to Standardized statuses and surfaced in `ListTransactions` (see comment above)

PR #969, opened 2026-07-25
Throughput demo: 250 TPS sustained on TSTN, self-healing fuel pool

sirdeggen:
## Summary

- Takes the throughput demo dashboard from ~12 createAction/s to **250 TPS sustained for 3h18m** on a private TSTN stack, with the FuelKeeper holding the UTXO pool the whole time (mint tracked burn within 2%).
- Measured overnight: **3,099,344 event transactions, 100% network acceptance** (every one `unproven`, zero failed/aborted/stranded), 99.96% createAction success in steady state. Report: `docs/throughput-overnight-run-report.html`.
- Adds TTN/TSTN network support, aligns Arcade SSE statuses, and fixes several storage-layer defects that only appear under sustained load.

The BRC-104 auth concurrency fixes this work depends on shipped upstream in [go-sdk#338](https://github.com/bsv-blockchain/go-sdk/pull/338), released as **go-sdk v1.3.2**. This branch depends on that release; the vendored `third_party/go-sdk` tree and its `go.mod` replace directive are gone.

## Changes

**Networks / Arcade**
- TTN and TSTN networks with per-network service defaults.
- SSE status names aligned with Arcade (`ACCEPTED_BY_NETWORK`, `SEEN_MULTIPLE_NODES`), MINED frames apply their BUMP merkle paths.
- Arcade registered as a `GetStatusForTxIDs` provider via `GET /tx/{txid}`, so status sync works on networks where WhatsOnChain/Bitails are unavailable (previously: `no services registered`).

**Funding correctness (storage)**
- A `UserUTXO` row advertising an already-spent output was fatal: selection is deterministic, so every attempt re-picked it and funding failed permanently against a full basket. Now reported as a retryable `StaleUTXOIndexError`, healed outside the rolled-back transaction, and no longer created (`spent_by` guard on change promotion).
- Change is indexed at the `sending` tier when queued for delayed broadcast, instead of waiting for network acceptance — acceptance runs far slower than creation, which starved the fuel pool.

**Server cost per action**
- Broadcast BEEFs carry only direct sources; script verification checks the subject transaction rather than re-validating our own ancestors' merkle roots against a gullible header client (that walk was 34% of storage CPU). Proof validation is unchanged for internalize and external inputs.
- Fan-outs and stream actions request txid-only results, trimming storage-managed inputs to txid stubs — the fan-out ancestry walk otherwise grew unbounded and eventually killed all minting.

**Capacity ceilings**
- Background broadcaster 10 → 128 workers (was ~33 tx/s, and outputs only become spendable on acceptance), buffer 1k → 50k.
- Connection pools on the Arcade client and on the storage client (`DefaultTransport` pools 2 idle connections per host), plus a per-request timeout so a dropped response cannot block its caller.
- Partial index for the funder's claim query: sequential scan + 19 MB external sort (~130 ms/action) → ~0.07 ms.
- Fuel target sized to the refill horizon rather than a 300s mined-confirmation assumption, so the keeper can reach low water and stop competing with the stream.

**Dashboard**
- Bounded concurrent wallet gate with context-aware waits, live status overlay, bounded stop with a draining state, keeper fair-share + parallel minting, honest per-second tick rates (late ticks previously overstated TPS by their lag), pprof endpoint on the infra binary.

**Demo stack**
- Postgres in the throughput compose file runs with `synchronous_commit=off` (plus checkpoint/shared-buffer tuning). `fsync` and `full_page_writes` stay **on**: the commit-flush wait is what the demo is bound on, and disabling the rest trades crash-recovery ordering guarantees for a marginal gain.

## Test plan

- [ ] `go build ./...` and `go test ./...` (green locally against go-sdk v1.3.2, including new regression tests for the stale-index retry, bounded stop, and keeper fair-share)
- [ ] Bring up the TSTN stack, fund the operator, start the stream at 250 TPS: fuel pool should oscillate between water marks rather than drain, network accept should hold at 1.0
- [ ] Confirm CI is green on this branch


mrz1836 (changes requested):
Gitleaks issues... if those are legitimate keys that should be included, they should be added to a gitleaks ignore file.

PR #972, opened 2026-07-27
Report transactions aborted before signing in ListTransactions

sirdeggen:
## What Changed
- `ListTransactions` now also reports user transactions that reached a terminal local status (`aborted`, `failed`) without leaving a `KnownTx` row behind, extracted into a `localOnlyTerminalTxs` helper.
- Two regression tests: one for the abort-before-signing case, one table-driven test pinning `aborted` and `failed` to distinct standardized statuses.

## Why It Was Necessary
Completes the post-merge review on #961. #971 made `aborted` a distinct `StandardizedTxStatus` and overrode it in `ListTransactions`, which fixed reporting **for transactions that have a `KnownTx` row** — but that row only appears once a transaction has been signed and spent (`upsertKnownTx`, inside `SpendTransaction`).

`AbortAction`, the `fail_abandoned` sweep and a canceled `ProcessAction` can all mark a `Transaction` `aborted` before that ever happens. Those transactions had nothing for the `KnownTx` query to return, so they were dropped from the result entirely. Measured on `main` before the change — a `CreateAction` followed by `AbortAction`, then `ListTransactions`:

```
total=1  len=1
  tx=57cb5176... status=broadcasted     <- fixture funding tx only; the aborted tx is absent
```

`ListTransactions` is the method callers use for idempotency, and it is the only one returning transaction data. "Absent" is indistinguishable from "never created", yet the two demand opposite responses: an aborted action has had its inputs released and must be rebuilt, whereas a missing one has to be created for the first time. After the change the aborted transaction is reported with `status=aborted`.

## Testing Performed
- `TestListTransactions_AbortedBeforeSigningIsReported` — creates a transaction, aborts it while still unsigned, asserts it is present and reported as `aborted`, matched on `Reference` since no `TxID` was ever assigned.
- `TestListTransactions_FailedAndAbortedAreDistinguishable` — drives a `KnownTx`-less transaction into each terminal status and asserts `aborted` reports `aborted` and `failed` reports `failed`.
- Both were confirmed to fail against `provider.go` from `main` and pass with the change, so they are genuine regression tests rather than assertions of current behavior.
- `./pkg/storage/...`, `./pkg/wdk/...` and `./pkg/internal/storage/...` pass; `go vet` and `gofmt` clean.

## Impact / Risk
Additive to the result set: transactions that were previously omitted are now returned. Nothing that was already reported changes, and no stored data or migration is involved.

Two deliberate constraints keep the blast radius small:
- Only terminal local statuses are surfaced. In-flight states (`unsigned`, `nosend`) are left out, so the result stays the set of transactions a caller can act on.
- When `args.Status` is set the caller is filtering on a `ProvenTxReqStatus`, which no `KnownTx`-less transaction can satisfy by definition, so nothing is added and the filter's meaning is preserved.

Callers that treat `failed` as the single terminal bucket need to handle `aborted` as well — already true as of #971, but more visible now that these transactions actually appear.

Worth a reviewer's eye: the helper deliberately mirrors the existing `KnownTx`-backed override (`failed` → `failed`) instead of calling `TxStatus.ToStandardizedStatus()`, which maps `failed` → `invalidTx`. Both halves of one result set have to label the same local status identically, but that divergence still exists in `wdk` and may deserve reconciling separately.

`TxID` is empty for a transaction aborted before signing, since none was ever assigned; callers match those on `Reference`.

## Notifications
- @kuba-4chain

### DIFF
--- a/pkg/storage/provider.go
+++ b/pkg/storage/provider.go
@@ -1260,9 +1260,12 @@ func (p *Provider) ListTransactions(ctx context.Context, auth wdk.AuthID, args w
 	}
 
 	if len(txIDs) == 0 {
+		// No signed transaction, but a locally-terminal one (aborted before it was
+		// ever signed) still has to be reported - see localOnlyTerminalTxs.
+		localOnly := localOnlyTerminalTxs(userTxs, nil, args)
 		return &wdk.ListTransactionsResult{
-			TotalTransactions: 0,
-			Transactions:      []wdk.CurrentTxStatus{},
+			TotalTransactions: primitives.PositiveInteger(uint64(len(localOnly))),
+			Transactions:      localOnly,
 		}, nil
 	}
 
@@ -1285,6 +1288,7 @@ func (p *Provider) ListTransactions(ctx context.Context, auth wdk.AuthID, args w
 	}
 
 	transactions := make([]wdk.CurrentTxStatus, 0, len(knownTxs))
+	reported := make(map[string]struct{}, len(knownTxs))
 	for _, ktx := range knownTxs {
 		status := ktx.Status.ToStandardizedStatus()
 		switch txStatusMap[ktx.TxID] { //nolint:exhaustive // only Failed/Aborted override the KnownTx-derived status; all other TxStatus values keep it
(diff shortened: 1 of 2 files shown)


mrz1836 (approved):
LGTM

pkg/storage/provider.go
```diff
@@ -1322,14 +1326,80 @@ func (p *Provider) ListTransactions(ctx context.Context, auth wdk.AuthID, args w
...
+// deliberately left out so the reported set stays the transactions the caller can act
+// on. When args.Status is set the caller is filtering on a ProvenTxReqStatus, which by
+// definition no KnownTx-less transaction can satisfy, so nothing is added.
+//
+// TxID is empty for a transaction aborted before signing, since none was ever assigned;
+// callers match those on Reference.
+func localOnlyTerminalTxs(userTxs []*entity.Transaction, reported map[string]struct{}, args wdk.ListTransactionsArgs) []wdk.CurrentTxStatus {
+	if args.Status != nil {
+		return nil
+	}
```
kuba-4chain, line 1367: The comment in `localOnlyTerminalTxs` says "filtering on a ProvenTxReqStatus, which by definition no KnownTx-less transaction can satisfy", but a user can also query for `aborted` and `failed` which are `StandardizedTxStatus` values that only local-only transactions can satisfy.

So maybe the guard should look like this?
```go
   func localOnlyTerminalTxs(userTxs []*entity.Transaction, reported map[string]struct{}, args wdk.ListTransactionsArgs) []wdk.CurrentTxStatus {
       if args.Status != nil &&
           *args.Status != wdk.TxUpdateStatusAborted &&
           *args.Status != wdk.TxUpdateStatusFailed {
           return nil
       }
       // ... and inside the loop, filter by args.Status if it's set:
       switch tx.Status {
       case wdk.TxStatusAborted:
           status = wdk.TxUpdateStatusAborted
           if args.Status != nil && *args.Status != wdk.TxUpdateStatusAborted {
               continue
           }
       case wdk.TxStatusFailed:
           status = wdk.TxUpdateStatusFailed
           if args.Status != nil && *args.Status != wdk.TxUpdateStatusFailed {
               continue
           }
       default:
           continue
       }

```

PR #973, opened 2026-07-28
fix(storage): post unconfirmed parents before children in delayed broadcast

galt-tr:
## Problem

The storage delayed-broadcast pool (`pkg/storage/internal/service/background_broadcaster.go`)
drained one channel with N concurrent workers (up to 256 for the throughput strategy) and **no
dependency ordering**. A transaction and the child that spends its still-unconfirmed output are
queued by separate, concurrent `createAction`/`processAction` calls, so a child was frequently
POSTed upstream **before its parent**.

Arcade forwards transactions to Teranode in the order it receives them, so an out-of-order child
reached Teranode before the parent's output was in the UTXO set and was rejected
(`failed to validate transaction`), cascading to every descendant. Observed on the tstn scaling
stack: ~half of fan-out/stream transactions rejected, fuel starvation, and a storage view that
optimistically recorded them `unproven` while the network had rejected them.

## Fix

Make the broadcaster dependency-aware. Before posting an item it inspects the item's beef; if the
tx spends an **unconfirmed** direct parent (present as `transaction.RawTx`, i.e. no merkle proof)
that has not been posted yet, the child is **parked** (keyed by the parent txid) instead of posted.
When a parent's POST succeeds, its parked children are released and re-queued. A configurable
`Sizing.MaxParentWait` deadline + sweeper force-posts a child if its parent never arrives, so a
never-posted parent cannot hold children forever.

Gating only fires on `RawTx` parents, so mined (`RawTxAndBumpIndex`), txid-only, and
confirmed/external sources are untouched — independent transactions still broadcast fully
concurrently. `NewBackgroundBroadcaster`'s signature is unchanged; `Sizing` gains an optional
`MaxParentWait` (default 30s).

## Verification

- New `TestBackgroundBroadcaster_PostsUnconfirmedParentBeforeChild`: a child enqueued before its
  unconfirmed parent is posted after it. Existing broadcaster tests pass; race detector clean.
- Live on the tstn scaling stack: Arcade rejections drop from ~50% to **0** at both 100 and 1000
  TPS, and the fuel pool builds instead of starving.

### DIFF
--- a/pkg/storage/internal/service/background_broadcaster.go
+++ b/pkg/storage/internal/service/background_broadcaster.go
@@ -5,7 +5,9 @@ import (
 	"fmt"
 	"log/slog"
 	"sync"
+	"time"
 
+	"github.com/bsv-blockchain/go-sdk/chainhash"
 	"github.com/bsv-blockchain/go-sdk/transaction"
 
 	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/logging"
@@ -32,6 +34,18 @@ const (
 	// BackgroundBroadcasterChannelSize is the default buffer size for the
 	// broadcast channel. Add() drops to the cron fallback once it is full.
 	BackgroundBroadcasterChannelSize = 1000
+
+	// defaultMaxParentWait bounds how long a child waits for its unconfirmed
+	// parent to be posted to the broadcaster before it is posted anyway. It is a
+	// safety net against a parent that is never broadcast here (e.g. an external
+	// unconfirmed source): the child is not held forever. In steady state the
+	// parent posts within milliseconds and the child is released immediately, so
+	// this deadline is rarely reached.
+	defaultMaxParentWait = 30 * time.Second
+
+	// parentWaitSweepInterval is how often parked children are re-checked for an
+	// expired parent-wait deadline.
+	parentWaitSweepInterval = time.Second
 )
 
 // Sizing configures the broadcaster's capacity. Zero values select the
@@ -39,6 +53,9 @@ const (
 type Sizing struct {
 	Workers     int
 	ChannelSize int
+	// MaxParentWait bounds how long a child is held waiting for its unconfirmed
+	// parent to be posted before being force-posted. Zero selects the default.
+	MaxParentWait time.Duration
 }
 
 func (s Sizing) workers() int {
@@ -53,6 +70,13 @@ func (s Sizing) channelSize() int {
 		return BackgroundBroadcasterChannelSize
 	}
 	return s.ChannelSize
+}
+
+func (s Sizing) maxParentWait() time.Duration {
+	if s.MaxParentWait <= 0 {
+		return defaultMaxParentWait
+	}
+	return s.MaxParentWait
 }
 
 type broadcaster interface {
@@ -71,12 +95,24 @@ type BackgroundBroadcaster struct {
 	// optional notification channel
 	txBroadcastedChannel chan<- wdk.CurrentTxStatus
 
+	// Dependency ordering: Arcade forwards to Teranode in the order it receives
+	// transactions, so a child that spends an unconfirmed parent must be POSTed
+	// after that parent — otherwise Teranode rejects the child (its input isn't
+	// in the UTXO set yet). Because parent and child arrive as independent
+	// concurrent items drained by many workers, we gate here: a child is held
+	// until every unconfirmed parent it spends has been posted.
+	depMu   sync.Mutex
+	posted  map[string]struct{}          // txids whose Arcade POST returned successfully
+	waiting map[string][]broadcastItem   // children parked under an unposted parent txid
+
 	stopOnce sync.Once
 }
 
 type broadcastItem struct {
 	beef  *transaction.Beef
 	txIDs []string
+	// deadline is when the item may be posted regardless of unposted parents.
+	deadline time.Time
 }
 
 func NewBackgroundBroadcaster(ctx context.Context, parentLogger *slog.Logger, broadcastHandler broadcaster, txBroadcastedChannel chan<- wdk.CurrentTxStatus, sizing Sizing) *BackgroundBroadcaster {
@@ -90,6 +126,8 @@ func NewBackgroundBroadcaster(ctx context.Context, parentLogger *slog.Logger, br
 		logger:               logger,
 		broadcastHandler:     broadcastHandler,
 		txBroadcastedChannel: txBroadcastedChannel,
+		posted:               make(map[string]struct{}),
+		waiting:              make(map[string][]broadcastItem),
 	}
 }
 
@@ -98,6 +136,8 @@ func (bb *BackgroundBroadcaster) Start() {
 		bb.wg.Add(1)
 		go bb.worker()
 	}
+	bb.wg.Add(1)
+	go bb.parentWaitSweeper()
 }
 
 func (bb *BackgroundBroadcaster) Stop() {
@@ -110,8 +150,9 @@ func (bb *BackgroundBroadcaster) Stop() {
 
 func (bb *BackgroundBroadcaster) Add(beef *transaction.Beef, txIDs []string) (added bool) {
 	bb.logger.InfoContext(bb.ctx, "Adding new beef to delayed broadcast", "txIDs", txIDs)
+	item := broadcastItem{beef: beef, txIDs: txIDs, deadline: time.Now().Add(bb.sizing.maxParentWait())}
 	select {
-	case bb.broadcastChannel <- broadcastItem{beef: beef, txIDs: txIDs}:
+	case bb.broadcastChannel <- item:
 		return true
 	default:
 		return false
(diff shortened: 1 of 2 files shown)


mrz1836 (dismissed):
CI is failing

pkg/storage/internal/service/background_broadcaster.go
```diff
@@ -110,8 +150,9 @@ func (bb *BackgroundBroadcaster) Stop() {
 
 func (bb *BackgroundBroadcaster) Add(beef *transaction.Beef, txIDs []string) (added bool) {
 	bb.logger.InfoContext(bb.ctx, "Adding new beef to delayed broadcast", "txIDs", txIDs)
+	item := broadcastItem{beef: beef, txIDs: txIDs, deadline: time.Now().Add(bb.sizing.maxParentWait())}
```
sirdeggen: **[bug]** `deadline` is set when the item is enqueued (`Add`), not when it is parked waiting for a parent. `process` skips parent gating entirely once `time.Now().After(item.deadline)`. Under backlog — large channel, posts slower than creates, the exact high-TPS regime this PR is meant to fix — a child can sit in `broadcastChannel` for longer than `MaxParentWait` (default 30s). When a worker finally dequeues it, it force-posts without waiting for the parent, reintroducing the out-of-order POST the PR is fixing. With throughput sizing (`workers * 400` buffer, up to 256 workers, ~300ms posts), multi-tens-of-seconds queue latency is realistic.

**Suggestion:** Only apply the parent-wait deadline while parked. For example: always run `firstUnpostedParentLocked` on first process; set `deadline = time.Now().Add(maxParentWait)` when inserting into `waiting`; have the sweeper (or a `force` flag it sets) be the sole path that bypasses gating. Alternatively start the deadline only on park and leave `deadline` zero until then.

pkg/storage/internal/service/background_broadcaster.go
```diff
@@ -129,11 +170,139 @@ func (bb *BackgroundBroadcaster) worker() {
...
+				continue
+			}
+			parentBeefTx := item.beef.Transactions[*in.SourceTXID]
+			if parentBeefTx == nil || parentBeefTx.DataFormat != transaction.RawTx {
+				continue // absent, txid-only, or mined (bump) parent → no gating
+			}
+			parentID := in.SourceTXID.String()
+			if _, ok := bb.posted[parentID]; ok {
+				continue
+			}
```
sirdeggen: **[bug]** `firstUnpostedParentLocked` gates on any unconfirmed `RawTx` parent that is not in `posted`, even when that parent is a subject of the same broadcast item (`item.txIDs`). `processDelayedTransactions` enqueues one item per batch: `Add(beef, readyToSendTxIDs)`. If a delayed multi-tx `SendWith` (or any batch) includes both a parent and a child that spends it, the whole item parks under the parent txid. That parent is only present in this item, so it is never marked posted, and the batch sits in `waiting` until the sweeper force-posts after `MaxParentWait`. That is a multi-second (default 30s) self-deadlock for an otherwise immediately broadcastable batch.

**Suggestion:** When resolving parents, ignore parent txids that appear in `item.txIDs` (they will be POSTed in the same Arcade request). Build a small set from `item.txIDs` once per call for O(1) membership checks. Add a regression test that `Add`s one beef with `[parentID, childID]` and asserts both post promptly without waiting for the deadline.

pkg/storage/internal/service/background_broadcaster.go
```diff
@@ -129,11 +170,139 @@ func (bb *BackgroundBroadcaster) worker() {
...
+	return ""
+}
+
+// markPostedAndRelease records the posted txids and re-queues any children that
+// were waiting on them.
+func (bb *BackgroundBroadcaster) markPostedAndRelease(txIDs []string) {
+	bb.depMu.Lock()
+	var released []broadcastItem
+	for _, txID := range txIDs {
+		bb.posted[txID] = struct{}{}
```
sirdeggen: **[bug]** `posted` is append-only. Every successfully delayed-broadcast txid is retained for the lifetime of the `BackgroundBroadcaster` (i.e. the storage provider process). There is no eviction, TTL, or size bound. At high delayed-broadcast rates this is an unbounded memory growth path: keys are 64-char hex strings plus map overhead, and throughput deployments keep this process up indefinitely. The map only needs to answer “has this parent already been POSTed by this broadcaster so a later child need not wait?” — not a permanent ledger.

**Suggestion:** Bound the map. Practical options: (1) store `postedAt time.Time` and periodically prune entries older than a generous window (e.g. several hours / unconfirmed lifetime), accepting rare false waits that the deadline already covers; (2) LRU/cap with eviction; (3) only track parents that currently have waiters or were posted within `MaxParentWait` if you also fix the enqueue-time deadline issue so late children are not the common case. At minimum document the growth and add a prune on the sweeper tick.

sirdeggen: **[suggestion]** `posted` only records successes from this broadcaster’s delayed path. Parents posted via the non-delayed path, `SendWaitingTransactions` (`broadcastTxs(..., isDelayed=false)`), or a previous process lifetime never appear in `posted`. Children that still carry those parents as `RawTx` in BEEF will park for the full `MaxParentWait` before force-posting. The PR text acknowledges external parents, but the same false-wait applies to in-process non-delayed parents and to cron fallback after a full channel — common when the delayed pool is saturated. Combined with the enqueue-time deadline issue, channel-full parent (cron posts it) + delayed child can still mis-order or stall.

**Suggestion:** Consider treating a parent as already satisfied if it is not expected to be posted by this pool (hard without more signals), or register intent on every `Add` and only wait when the parent has been enqueued / is in-flight / is posted. Shorter default `MaxParentWait` for throughput sizing may also help if full “knows about parent” tracking is out of scope. Worth a comment in `firstUnpostedParentLocked` that `posted` is delayed-pool-local, not global network state.

pkg/storage/internal/service/background_broadcaster.go
```diff
@@ -129,11 +170,139 @@ func (bb *BackgroundBroadcaster) worker() {
...
+	bb.requeue(released)
+}
+
+// requeue puts released/expired items back on the broadcast channel without
+// blocking the caller (a full channel would otherwise deadlock a worker).
+func (bb *BackgroundBroadcaster) requeue(items []broadcastItem) {
+	if len(items) == 0 {
+		return
+	}
+	bb.wg.Add(1)
```
sirdeggen: **[nit]** `requeue` spawns a goroutine per release batch and uses `wg.Add` so `Stop` waits. That is correct and avoids worker deadlock on a full channel. Under a storm of parent completions this can create many short-lived goroutines; acceptable, but a single dedicated requeue loop or non-blocking send with a small side buffer would be simpler to reason about. Not blocking.

**Suggestion:** Optional follow-up: one long-lived requeue helper fed by a lock-free queue, or try non-blocking send and only spawn a goroutine on the full-channel path.

sirdeggen:
Pushed two commits addressing all review feedback, plus a CI fix.

**Review threads**

- **`wg.Add` racing `Stop`'s `Wait`** — requeueing no longer spawns a goroutine per release batch. `Start` creates one long-lived `requeueDispatcher`; `requeue` appends to a mutex-guarded queue and pokes a 1-slot wake channel. No `wg.Add` after startup.
- **Same-batch parent parked the whole batch** — `Add` runs `parentsFirst`, a stable topological sort of `item.txIDs` (`PostFromBEEF` posts in slice order), and gating skips parents that are subjects of the same item. Both suggested remedies, so an in-batch parent is genuinely posted first rather than merely assumed to be.
- **Unbounded `posted` map** — `posted` is now a `txidSet`: a `txid → postedAt` map plus an insertion-ordered slice the sweeper front-prunes past `Sizing.PostedRetention` (default 2h). Pruning costs only what it removes, not a scan.
- **Deadline started at enqueue, not at park** — items carry `parkedAt`, zero until the first park; only the sweeper's `force` flag bypasses gating. A child queued behind a backlog can no longer force-post ahead of its parent.
- **`posted` is delayed-pool-local** — implemented rather than documented, because it was breaking CI (below). An `enqueued` set records what this pool was actually asked to broadcast. A parent that is enqueued here gets the full `MaxParentWait`; a parent this pool has never seen gets only `UnknownParentWait` (default 500ms), enough to cover the concurrent-enqueue race the gating exists for, without stalling on a POST that is never coming.

**CI fix**

`TestProcessAction_DelayedBroadcast`, `TestProcessAction_DelayedBroadcastForManyTransactions` and `TestSendWaitingTransactions_DelayedBroadcastCountsOnlyActualPost` fail on 817ed12 (verified by checking that commit out). Any beef carrying an unconfirmed raw parent parked for the full 30s, so transactions stayed `unsent` past the assertions' 5s budget. The unknown-parent grace fixes all three.

**Verification**

- `go test ./...` green repo-wide; `-race` clean on the broadcaster package.
- `golangci-lint` reports 0 issues on `pkg/storage/internal/service/...`.
- Four new tests: same-batch parent/child posts without waiting, a queued child still waits for its parent, a child of a never-enqueued parent does not stall, a child of a pruned parent is still posted.

`Sizing` now also carries `UnknownParentWait`, `PostedRetention` and `SweepInterval`, all defaulted; `NewBackgroundBroadcaster`'s signature is still unchanged. Note the PR description above still describes only the original `MaxParentWait` design.

---
_Generated by [Claude Code](https://claude.ai/code)_

sirdeggen:
CI status on `7bb26bf`: the only failure is `🔒 Security Scans`, and it is a pre-existing base-branch failure, not something this PR introduces.

- The job reports `NANCY_STATUS: failure` with `GOVULN_STATUS: success` and `GITLEAKS_OUTCOME: success` — Nancy found a vulnerability in a dependency.
- The same job fails identically on `main`'s own push run for `f14c1c2` ([run 30377696936](https://github.com/bsv-blockchain/go-wallet-toolbox/actions/runs/30377696936)), where `🎯 All Tests Passed` is red solely because of security scans and every other job is green.
- This PR changes no dependencies — `go.mod` and `go.sum` are untouched; the diff is `background_broadcaster.go`, its test, and the two workflow files from the `main` merge.

Fixing it means bumping the offending dependency or adding its CVE to `NANCY_EXCLUDES`, which belongs on `main` rather than in this PR. Happy to do either if that is wanted here.

For the record, the earlier `🎯 All Tests Passed` failure on `61cec61` was also not a real one: the `main` merge landed 83 seconds after that push and `cancel-in-progress` cancelled the in-flight jobs, which the gate counts as failed.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #976, opened 2026-07-29
chore(security): exclude the gopacket sFlow/Diameter CVEs from Nancy

sirdeggen:
## What Changed

- Added `CVE-2026-54332` and `CVE-2026-54345` to `NANCY_EXCLUDES` in `.github/env/90-project.env`.
- Nothing else. `MAGE_X_CVE_EXCLUDES` is untouched — govulncheck passes and needs no exclusion.

## Why It Was Necessary

`🔒 Security Scans` is failing on **every** branch, `main` included, with `NANCY_STATUS: failure` while `GOVULN_STATUS: success` and `GITLEAKS_OUTCOME: success`. It is currently red on:

- `main`'s own push run for `f14c1c2` ([run 30377696936](https://github.com/bsv-blockchain/go-wallet-toolbox/actions/runs/30377696936))
- #974, the dependabot `gomod-all` PR
- #973

The last green `main` run was `e25fb238` on Jul 27 17:39; the first failures appear Jul 28 ~15:00. No dependency changed in that window, which is the signature of a newly published advisory rather than a bad merge. Two were published on 2026-07-28 against `github.com/gopacket/gopacket`:

| CVE | Decoder | Issue |
|---|---|---|
| CVE-2026-54332 | sFlow `ExtendedGatewayFlow` | unbounded allocation — 104-byte UDP datagram to ~16 GiB |
| CVE-2026-54345 | Diameter AVP | `uint32` underflow on vendor header size to ~4 GiB allocation |

Neither is reachable from this codebase:

- The dependency here is the **unmaintained original**, `github.com/google/gopacket v1.1.19`, pulled in transitively by `teranode`, `go-chaintracks`, `go-p2p-message-bus`, and `go-teranode-p2p-client`.
- The only package of it that enters the build is `github.com/google/gopacket/routing` (confirmed against `go list -json -deps ./...`). The vulnerable sFlow and Diameter decoders live in `layers`, which nothing imports.
- `v1.1.19` is the final release of that fork, so there is no fixed version to bump to. Exclusion is the only lever short of `libp2p` migrating to the maintained fork.

This mirrors the existing `CVE-2023-26248` entry, which already excludes this same dependency for the same reason.

## Testing Performed

- Traced the dependency with `go mod graph` and `go list -json -deps ./...` to confirm only the `routing` package is built.
- Cross-referenced the GitHub Advisory Database for Go advisories published in the failure window.
- **Not verified locally**: my environment's egress policy blocks `ossindex.sonatype.org` (403 on CONNECT), so I could not run `nancy sleuth` to confirm these are the exact IDs it reports. This PR's own `🔒 Security Scans` job is the verification — if it goes green, the identification was right. If it stays red, the CVE IDs need correcting from the `nancy-scan-results` artifact and I will follow up.

## Impact / Risk

- CI-only change; no product code, no dependency versions.
- Risk is limited to suppressing two CVEs in a dependency whose vulnerable code is not in the build. If `google/gopacket` ever gains a reachable use of `layers`, these entries should be revisited.
- Unblocks the required-checks gate on every open PR.

## Notifications

- @sirdeggen

---
_Generated by [Claude Code](https://claude.ai/code/session_01M5F4YPqWxPJp4f5veuLv4T)_


sirdeggen:
**This PR does not fix the problem — its own `🔒 Security Scans` job still fails.** Converting to draft so it is not merged on my reasoning; every other check on it is green, including the full test matrix.

As stated when I opened it, I could not run Nancy locally (egress policy blocks `ossindex.sonatype.org`), so `CVE-2026-54332` / `CVE-2026-54345` were an inference from the advisory publication window. The run disproves it: with both excluded, Nancy still exits non-zero.

I also could not retrieve the actual Nancy output to correct the IDs. For the record, what is blocked:

- `ossindex.sonatype.org` — 403 at the egress proxy, so no local `nancy sleuth`.
- The `nancy-scan-results` artifact and the raw job-log endpoint — both redirect to blob storage, also blocked.
- The job log via API is size-capped and starts after the Nancy step.
- The failure annotation only says "see job summary for details"; step summaries have no API.

Two hypotheses remain, and they need different fixes:

1. **A newly published CVE** in the tree that I have not identified. Fix: exclude the correct ID (or bump, if a fixed version exists).
2. **Nancy is erroring rather than finding anything.** The step only treats the literal string `rate limited by OSS Index` as inconclusive; any other failure — expired/absent `OSS_INDEX_USER`/`OSS_INDEX_TOKEN`, a 401/403, an API change — exits non-zero and is reported as "detected vulnerabilities". This would equally explain simultaneous failure across `main`, #973, #974 with no dependency change, and would explain why excluding CVEs changes nothing.

Hypothesis 2 is worth checking first, since it is a one-line look at the job summary. If it is real, the durable fix is to widen the inconclusive-error handling so an OSS Index outage or auth failure cannot mark every branch as vulnerable.

Anyone with the job summary in front of them: paste the Nancy output block here and I will push the correct fix.

---
_Generated by [Claude Code](https://claude.ai/code)_

mrz1836 (approved):
LGTM

PR #978, opened 2026-07-31
feat: fix double-spend revert tx

dzolt-4chain:
## What Changed

- Stopped restoring a transaction's spent inputs back to spendable when it is terminally failed as a double-spend or invalid tx. The three places that did this (process.updateSingleTx on broadcast failure, KnownTx.FailKnownTxAsDoubleSpend on externally-rejected txs, and synchronizeTxStatuses.reviewKnownTxStatuses on terminal review) now leave the spent inputs claimed instead of calling RecreateSpentOutputs/recreateSpentOutputs.
- Outputs created by the failed tx are still marked non-spendable, since those never existed on-chain if the tx failed - only the restoration of the spent inputs was removed.
- Added a regression assertion to TestBroadcastConflict_MempoolConflict_TxUnknown_StaysFailed proving a double-spend-failed tx's inputs stay claimed and non-spendable.
- Updated two ListFailedActions tests that had been (unintentionally) relying on the old revert-to-spendable behavior to fund a second test action with the same input; they now fund independently.

## Why It Was Necessary

- WhatsOnChain reports "missing inputs" far more often than genuine double-spends occur (e.g. a parent tx in a chain hasn't reached its mempool yet), and the toolbox's broadcast aggregation folds "missing inputs" into the same DoubleSpend=true signal as a real txn-mempool-conflict.
- A "confirmed" double-spend verdict only proves a competing tx exists - it does not prove the original spend is invalid. Restoring the input to spendable let the wallet re-spend it while the original tx was potentially still valid/pending, producing a real double-spend on top of the reported one and corrupting the wallet's UTXO state.
- Better to lose access to some valid inputs than to risk generating a real double-spend by resurrecting one that was only ever a false-positive rejection.

## Testing Performed

- go build ./..., go vet ./..., gofmt -l on all changed files - clean.
- go test ./... - full suite passes.
- Added regression test in pkg/storage/provider_broadcast_double_spend_test.go (confirmed RED before the fix, GREEN after).
- Verified TestListFailedActionsWithUnfail_ToleratesTransactionWithoutKnownTxRow and TestListFailedActions_IncludesAbortedWithDistinctStatus pass after decoupling their test fixtures from the old behavior.

## Impact / Risk

- Behavior change: inputs spent by a tx that ends up double-spend/invalid-failed are no longer released back to the wallet's spendable UTXO set - they remain permanently claimed (effectively lost) rather than resurrected. This trades occasional loss of a genuinely-valid input (on a false-positive rejection) for eliminating a class of real double-spend/wallet-corruption bugs.
- No API or schema changes; scope is limited to the three internal terminal-failure code paths in the storage layer.
- Pre-broadcast abort and the unfail/merkle-path-timeout cascade were intentionally left unchanged (different risk profile - see code comments); worth revisiting separately if the same false-positive pattern turns up there.

## Notifications
- @username

### DIFF
--- a/pkg/internal/storage/repo/known_tx.go
+++ b/pkg/internal/storage/repo/known_tx.go
@@ -153,21 +153,11 @@ func (p *KnownTx) FailKnownTxAsDoubleSpend(ctx context.Context, txID string, ski
 			return fmt.Errorf("failed to update transaction status as failed: %w", txErr)
 		}
 
-		// NOTE: There can be multiple transactions with the same tx_id, so we need to restore all of them.
-		var transactionIDs []uint
-		txErr = tx.Model(&models.Transaction{}).
-			Where(p.query.Transaction.TxID.Eq(txID)).
-			Pluck(p.query.Transaction.ID.ColumnName().String(), &transactionIDs).Error
-		if txErr != nil {
-			return fmt.Errorf("failed to find transaction IDs: %w", txErr)
-		}
-
-		query := genquery.Use(tx)
-		for _, transactionID := range transactionIDs {
-			if txErr = recreateSpentOutputs(ctx, query, transactionID); txErr != nil {
-				return fmt.Errorf("failed to restore spent outputs for transaction %d: %w", transactionID, txErr)
-			}
-		}
+		// Spent inputs are intentionally NOT restored to spendable here: a confirmed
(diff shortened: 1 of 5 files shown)


mrz1836 (approved):
LGTM

PR #986, opened 2026-08-11
feat: prefetch known tx during BEEF building

pawellewandowski98:
## What Changed
- Refactored `GetBEEFForTxIDs` in `pkg/internal/storage/repo/known_tx_get_beef.go` to batch-fetch transaction records using a single SQL `IN` query.
- Passed a `preFetched` cache map down to `recursiveBuildValidBEEF` to allow the recursive builder to bypass individual database queries when the transaction data has already been retrieved in the batch fetch.

## Why It Was Necessary
- Distributed traces showed massive delays (e.g., ~10 seconds) when building BEEF payloads for unconfirmed transactions with numerous inputs.
- The previous implementation suffered from an N+1 query problem, hitting the database individually for every single parent transaction and redundantly deserializing massive overlapping historical `input_beef` payloads.
- The new batch query approach prevents sequential network/database blocking and significantly reduces CPU parsing overhead during transaction broadcast.

## Testing Performed
- Ran the existing repository test suite (`go test ./pkg/internal/storage/repo/...`) to verify the modified recursive fallback logic behaves exactly as expected.
- Created and ran a custom benchmark utilizing an in-memory SQLite database across 50 simulated inputs.
- **Results:** The batch-fetching approach proved **10x faster** (1.76ms vs 17.6ms per operation) and reduced memory allocations by nearly **63%** (2,427 vs 6,564 allocs/op) even without factoring in the massive network latency savings that will be realized on a real PostgreSQL database.

## Impact / Risk
- **Impact:** Significant reduction in latency and CPU usage during complex transaction creation and broadcasting (particularly for transactions with many inputs).
- **Risk:** Low risk. If an input is somehow missing from the batch pre-fetch, the code gracefully falls back to the original individual DB lookup (`query.First`) and subsequently the external service lookup (`TxGetterFcn`), ensuring functional parity.

### DIFF
--- a/pkg/internal/storage/repo/known_tx_get_beef.go
+++ b/pkg/internal/storage/repo/known_tx_get_beef.go
@@ -31,7 +31,7 @@ func (p *KnownTx) GetBEEFForTxID(ctx context.Context, txID string, opts ...entit
 		beef = options.MergeToBEEF
 	}
 
-	err = p.recursiveBuildValidBEEF(ctx, 0, beef, txID, options)
+	err = p.recursiveBuildValidBEEF(ctx, 0, beef, txID, options, nil)
 	if err != nil {
 		return nil, fmt.Errorf("failed to build valid BEEF: %w", err)
 	}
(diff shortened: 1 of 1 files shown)


dzolt-4chain (approved):

mrz1836 (approved):
LGTM

PR #474, opened 2025-08-13
feat: add wallet services integration and GetHeight method implementation

jakubmkowalski:
Closes #475
This pull request introduces support for wallet services in the wallet creation and usage flow, and implements the `GetHeight` method to retrieve the current blockchain height. It also adds comprehensive tests for the new functionality and originator validation. The changes are grouped into three main themes: wallet service integration, new feature implementation, and testing.

**Wallet service integration:**

* Added an optional `services` field to the `Wallet` struct and support for injecting wallet services via the new `WithServices` option in both the wallet builder and the wallet constructor. (`pkg/wallet/internal/testabilities/fixture_wallet_builder.go`, `pkg/wallet/internal/wallet_opts/wallet_opts.go`, `pkg/wallet/wallet.go`) [[1]](diffhunk://#diff-46c9ad57016c699e7246a8becb512f71d90028724a9ff56c9e0136e52df4c3caR4-R13) [[2]](diffhunk://#diff-46c9ad57016c699e7246a8becb512f71d90028724a9ff56c9e0136e52df4c3caR33-R41) [[3]](diffhunk://#diff-46c9ad57016c699e7246a8becb512f71d90028724a9ff56c9e0136e52df4c3caR50-R54) [[4]](diffhunk://#diff-46c9ad57016c699e7246a8becb512f71d90028724a9ff56c9e0136e52df4c3caL62-R79) [[5]](diffhunk://#diff-72c23512e5a524b79bc536aa7aa09765e9f27e5ebb2ce5b24860cd6827ad627dL3-R10) [[6]](diffhunk://#diff-ea88400ee9978d3861f790f89a8efae2e437c92fa450cf110dbde7a2a0fcbc42R10) [[7]](diffhunk://#diff-ea88400ee9978d3861f790f89a8efae2e437c92fa450cf110dbde7a2a0fcbc42R27) [[8]](diffhunk://#diff-ea88400ee9978d3861f790f89a8efae2e437c92fa450cf110dbde7a2a0fcbc42R61-R67) [[9]](diffhunk://#diff-ea88400ee9978d3861f790f89a8efae2e437c92fa450cf110dbde7a2a0fcbc42R99-R107)

**New feature implementation:**

* Implemented the `GetHeight` method on the `Wallet` struct to return the current blockchain height using the configured wallet services, with validation for the `originator` parameter and error handling if services are not configured. (`pkg/wallet/wallet.go`)

**Testing:**

[...]

PR #990, opened 2026-08-13
docs: add UTXO lifecycle page and correct stale certificate note

galt-tr:
## What Changed

- **New `docs/utxo-lifecycle.md`** — traces a transaction through the BRC-100 surface as implemented here:
  - Swim lanes for `CreateAction`, `ProcessAction`/broadcast, and `AbortAction`, plus a state diagram for the release compensation machine.
  - The layer map down to the seam that matters: `pkg/storage/internal/actions/repos.interface.go` and its `UnitOfWork`.
  - **The two-phase UTXO reservation** written out — `reserveUTXOs` CAS on `reserved_by_id IS NULL` → `ErrUTXOContention` (retried) versus `markReservedOutputsAsNotSpendable` guarded on `spent_by IS NULL` → `StaleUTXOIndexError` / `ErrProvidedInputConflict` (not retried), and why only the first is retried.
  - Broadcast outcome mapping, including why service error keeps change unspendable and why the guarded transaction write happens before the failure cascade.
  - The three status vocabularies (`TxStatus`, `ProvenTxReqStatus`, `UTXOStatus`) and where they diverge from the TypeScript implementation.
  - Coin selection, basket roles, and the four registered monitor tasks.
  - A parity-gap table covering the tracked `plans/issue-*.md` work plus two differences **not yet tracked as issues** (below).

- **`docs/wallet.md`** — replaced the stale note claiming the certificate APIs are placeholders.

## Why It Was Necessary

`docs/wallet.md` stated that `AcquireCertificate`, `ListCertificates`, `ProveCertificate`, `RelinquishCertificate` and `Discover*` were "placeholders and not yet implemented". All six are implemented in `pkg/wallet/wallet.go` — acquisition by both `issuance` and `direct` protocols, and both discovery methods with overlay lookup and caching. The note has been replaced with an accurate section, and a *genuine* remaining gap is documented in its place: the privileged keyring is still unwired, with nine `PrivilegedKeyManager` TODOs across the crypto and key-linkage methods.

Beyond that, the UTXO lifecycle here had no single description. `docs/storage.md` covers components and configuration; `plans/high-throughput-utxo-management.md` covers the throughput design. Neither explains what an ordinary `CreateAction` does to the database, and nothing described the reservation model, the compensation machine, or the guarded-write discipline that the code depends on.

## Testing Performed

- Documentation only — no Go code changed, so no build, vet, or lint delta.
- All 4 mermaid blocks on the new page parsed with `mermaid.parse()` under jsdom → 0 failures. GitHub renders mermaid natively; there is no docs pipeline in this repo to gate it.
- Every `file:line` reference and every behavioral claim was read against source rather than inferred. The two previously-undocumented differences below were each confirmed at the exact line, including their in-code rationale comments.

## Impact / Risk

No runtime impact — markdown only.

The page records two differences from the TypeScript implementation that were **not previously tracked as issues** and may deserve them:

1. **`GetNetwork` returns `main`/`test`**, where BRC-100 requires `mainnet`/`testnet`. `pkg/wallet/brc100_conformance_test.go` documents this and configures its vectors around the internal representation, so conformance passes without the wire value matching the spec.
2. **The V1 storage client returns empty results instead of failing.** `findOutputBasketsAuth` and `findOutputsAuth` in `pkg/storage/client.go` return empty collections with a `nil` error — a caller cannot distinguish that from a genuine empty result. `SetActive` and `ProcessSyncChunk` at least return errors.

It also records a behavioral divergence that is *deliberate on both sides* and unspecified by BRC-100: on a failed broadcast this implementation leaves spent inputs spent (a false-positive verdict must not cause a real double spend), while TypeScript releases them and then re-marks only those the chain positively confirms are gone. Documented as a difference, not as a defect.

A companion page giving the cross-implementation view is opened against `bsv-blockchain/ts-stack` (PR #474); this page links to it.

## Notifications

-

---
_Generated by [Claude Code](https://claude.ai/code/session_01Lug8S7eE3uxTFRtr5pU9kw)_

### DIFF
--- a/docs/utxo-lifecycle.md
+++ b/docs/utxo-lifecycle.md
@@ -1,0 +1,302 @@
+## UTXO LIFECYCLE: BSV Wallet Toolbox
+
+This page traces a transaction through the BRC-100 wallet interface as implemented here:
+which layer does what, which repository calls each method makes, and what state the
+transaction and its outputs are left in.
+
+The cross-implementation view — the same lifecycle described against the BRC-100
+specification with the TypeScript implementation side by side — lives in
+[`ts-stack/docs/architecture/wallet-utxo-lifecycle.md`](https://github.com/bsv-blockchain/ts-stack/blob/main/docs/architecture/wallet-utxo-lifecycle.md).
+This page is the Go-specific half: the repository layer, the two-phase UTXO reservation,
+and the release compensation machine, none of which have TypeScript counterparts.
+
+### Layers
+
+| Layer | Package |
+|---|---|
+| Wallet | `pkg/wallet/wallet.go` — BRC-100 surface, asserts `sdk.Interface` |
+| Actions | `pkg/wallet/internal/actions/` — create/sign orchestration, release compensation |
+| Manager | `pkg/storage/storage_manager.go` — auth injection, active/backup routing |
+| Provider | `pkg/storage/provider.go` — implements `wdk.WalletStorageProvider` |
+| Storage actions | `pkg/storage/internal/actions/` — create, process, abort, internalize, list |
+| Funder | `pkg/internal/storage/funder/` — coin selection |
+| Repo | `pkg/internal/storage/repo/` — SQL, reservations, guarded status writes |
+| Monitor | `pkg/monitor/` — background convergence |
+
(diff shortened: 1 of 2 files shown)


mrz1836 (approved):
LGTM

PR #996, opened 2026-08-14
perf: use knownTxIds, optimize beef

pawellewandowski98:
# perf(beef): stop rebuilding ancestries the wallet already holds

> **Depends on** the go-sdk PR *"perf(transaction): make BEEF merging linear"*.

## Why

Production traces showed `Wallet-ListOutputs` at **2m21s** with its in-process storage call
taking 2.16s, and `Task-send_waiting` spending 9.89s + 6.32s inside
`Repository-KnownTx-GetBEEFForTxIDs`. DB work was negligible — `ListAndCountOutputs` was
3.93ms. Stored `input_beef` blobs had reached **337 KB**.

The go-sdk PR fixes the merge algorithms. This PR stops feeding them work that never needed
doing, and fixes three bugs found on the way.

### 1. `ListOutputs` never told storage what the wallet already had

`MapListOutputsArgs` never populated `KnownTxids`, so `entity.WithKnownTxIDs(...)` always
received an empty set, `IsKnownTxID` was always false, and storage rebuilt the **full
ancestry of every output on every call** — which the wallet then re-merged into its beef
party. `wdk.ListOutputsArgs.KnownTxids` already existed, was already validated, and was
already honoured by storage; only the wallet-side wiring was missing.

Now populated from the party graph via the same `WalletParty.GetKnownTxIDs()` helper
`CreateAction` uses. Storage replies with txid-only stubs, and the wallet resolves them back
before returning — **callers receive exactly the same data as before**.

### 2. `CreateAction` had two unreachable branches

```go
if a.wdkArgs.Options.KnownTxids == nil {   // never true
```

`slices.Map` returns `make([]R, 0)` for a nil input — a **non-nil empty slice** — so the
auto-advertise block never ran and `CreateAction` never told storage anything either.
Verified by probe: `KnownTxids nil=false len=0` on every call.

The resolve half was equally dead: `result.Tx` is only populated when `ReturnTXIDOnly` is
false, so `if ReturnTXIDOnly { resolve }` could never fire. Both fixed — the check is now on
length, and the resolve is unconditional so stubs are never handed to callers.

Effect on the storage→wallet payload for a chain of self-spends:

| round | before | after |
|---|---|---|
| 1 | 458 B | 458 B |
| 2 | 923 B | **39 B** |
| 3 | 1,387 B | **39 B** |
| 4–12 | 1,387–1,613 B | **39 B** |

It grew with the chain; now it is flat.

### 3. `DirectSourcesOnly` still parsed the whole stored ancestry

The short-circuit only applied at `depth >= 1`, so a `send_waiting` build still parsed and
merged the subject's entire `input_beef` — up to 337 KB — at depth 0, purely to reach a
handful of immediate parents.

It is now skipped **when every direct parent is a known tx in its own right**, and those
parents are read in one batched query instead of one per input.

The condition matters: the first version of this change skipped the blob unconditionally and
broke `TestCreateWithUnknownInputThenProcess` with *"not known to storage"*. Caller-supplied
inputs exist **only** inside that blob — they are never `known_txes` rows — so dropping it
loses them. `DirectSourcesOnly` previously had no direct test coverage at all; both branches
are now covered.

### 4. Bounding the graph discarded transactions the wallet had just advertised

Found by running this against real data. A wallet whose graph reached **257** transactions
(cap: 256) failed on its second `ListOutputs`:

```
tx with only txid not found in beef party: 980991f1…
```

`MergeBeefFromParty` pruned the over-sized graph **after** the wallet advertised those txids
and **before** it resolved the stubs storage sent back for them. The wallet claimed
knowledge, discarded the evidence, then needed it back.

Latent before this PR — `ListOutputs` never advertised, so responses never contained stubs
and the prune was harmless. Making the advertise real made it fatal.

The bound now runs at **advertise** time (`ValidateTransactions`), so the advertised set is
always the retained set, and never between advertise and resolve.

`DefaultMaxGraphBUMPs` was also added: transaction count alone never bounded merge cost,
since merging a proof is compared against the proofs already held.

### 5. `tracing.Enable` silently dropped the spans of any short-lived process

```go
err = exporter.Shutdown(ctx)   // exporter dies here
err = tp.Shutdown(ctx)          // ...then the batcher flushes through it
```

Spans are batched; `tp.Shutdown()` is what drains the queue *through* the exporter. Closing
the exporter first discards everything still queued. Long-running services never noticed —
their 5-second batch timer had already exported. Any script finishing inside that window
lost **100%** of its spans, which is why wallet-side traces never appeared in Jaeger while
the storage server's did.

Now `ForceFlush` → `tp.Shutdown` → `exporter.Shutdown`.

### 6. Observability for the beef party

`ctx` is threaded through every `BeefParty` method that walks or holds the graph, with spans:

| Span | Attributes |
|---|---|
| `BeefParty-LockWait` / `BeefParty-LockHeld` | `party.graph_txs`, `party.graph_bumps` |
| `BeefParty-MergeBeefFromParty` | `party`, `beef.bytes` |
| `BeefParty-MergeBeef` | `other.txs`, `other.bumps`, graph size |
| `BeefParty-ValidateTransactions`, `-PruneIfOversized` | graph size |
| `BeefParty-GetKnownTxIDsForParty`, `-GetTrimmedBeefForParty` | `party` |
| `BeefParty-AddKnownTxIDsForParty` | `party`, `txids` |
| `BeefParty-MergeTxidOnly`, `-Clone` | — |
| `BeefParty-VerifyReturnedTxIDOnly(Atomic)` | `beef.bytes` |

`WithLock` splits **wait** from **hold**: one wallet shares one graph behind one mutex, so a
slow action is either doing too much work or queued behind another, and only separating the
two tells you which. `AddParty`/`IsParty` are left untraced — map lookups, no graph work.

## Evidence from production data

Running the updated `list_outputs` example against a real wallet:

```
=== server ===
round  ms   outputs  beef_bytes  beef_txs  bumps  blocks  proved_txs  depth  stubs
1      73   120      54660       120       4      4       120         0      0
2      30   120      54660       120       4      4       120         0      0
3      16   120      54660       120       4      4       120         0      0
  -> 120 proved transactions share 4 BUMP(s): 116 fewer proofs than one-per-transaction
```

- **`bumps == blocks` in every row.** Before, that response carried 120 BUMPs — one per
  transaction. This is the mechanism behind the 337 KB blobs.
- **Latency falls as the graph warms:** 73 → 30 → 16 ms, with `beef_bytes` byte-identical.
  Callers get the same data; the saving is on the wire and in the merge.
- **`stubs = 0`** — nothing is handed back unresolved.

## Breaking changes

- **`wdk.BeefParty` methods now take `context.Context`**: `WithLock`, `MergeBeefFromParty`,
  `MergeBeef`, `MergeTxidOnly`, `Clone`, `ValidateTransactions`, `AddKnownTxIDsForParty`,
  `GetKnownTxIDsForParty`, `GetTrimmedBeefForParty`, `PruneIfOversized`. `BeefParty` is
  exported and reachable via `Wallet.GetBeefParty()`, so external callers are affected.
- **`WalletParty.GetKnownTxIDs` takes `context.Context`.**
- **`CreateAction` and `ListOutputs` now send `KnownTxids`** to storage by default, so
  storage responses contain txid-only stubs. The wallet resolves them before returning; the
  data callers receive is unchanged.
- `MergeBeefFromParty` no longer prunes — `ValidateTransactions` does.

## Testing

`go test ./pkg/...` — 51 packages, all green, `go vet` clean.

New coverage, each pinning a bug this PR found:

- `pkg/wdk/beef_party_bound_test.go` — reproduces the 257-vs-256 failure exactly and asserts
  that anything advertised stays resolvable across rounds that cross the cap, plus a
  counter-assertion that pruning still happens so it cannot pass by never pruning.
- `pkg/wallet/wallet_list_outputs_test.go` — a **second** `ListOutputs` call still returns
  full transactions. The existing test made only one call, so the party graph was empty and
  this path had no coverage. Runs on SQLite and remote storage.
- `pkg/wallet/wallet_create_action_known_txids_test.go` — same contract for `CreateAction`:
  every input source present, no txid-only entries.
- `pkg/internal/storage/repo/known_tx_get_beef_direct_sources_test.go` — both
  `DirectSourcesOnly` branches: the blob is skipped when parents are storage-known, and
  still merged when a parent is caller-supplied.
- `pkg/tracing/provider_flush_test.go` — a stub OTLP gRPC collector asserting a short-lived
  process actually delivers its spans. Confirmed to fail on the old shutdown ordering.
- `pkg/wallet/wallet_deep_beef_test.go` — reproduces deep and wide unproved ancestries
  without external infrastructure, and reports BEEF growth per round.
- `pkg/internal/assembler/assembled_transaction_bench_test.go` — regression guard on
  `ToAtomicBEEF`.


mrz1836:
Sorry @pawellewandowski98 I had a PR open i was trying to merge last night. Please fix the merge conflict and then we can get it merged 🙏

mrz1836 (dismissed):
please resolve merge conflict

dzolt-4chain (approved):

sirdeggen:
Resolved the merge conflicts by merging latest `main` (`#995` dep updates + `#997` arcade rejected-tx retry).

Resolution:
- Took `main`'s `golang.org/x/exp` and remaining dependency updates
- Kept this PR's go-sdk pin `v1.3.4-0.20260814070058-13edcd08a74c` (BEEF merge linearization)
- Restored `00-core.env` govulncheck pin to `1.26.5`; `90-project.env` still overrides to `1.26.6`

Local verification: `go test` green for `pkg/wdk`, `pkg/wallet`, `pkg/storage`, `pkg/wallet/internal/{party,actions}`, `pkg/internal/storage/repo`, `pkg/storage/internal/actions`.

PR #997, opened 2026-08-14
fix: arcade rejected tx retry

dzolt-4chain:
## What Changed
- applyExternalRejected's ServiceError branch (no double-spend evidence on an Arcade REJECTED/DOUBLE_SPEND_ATTEMPTED event) no longer no-ops. It now guardedly transitions the KnownTx back to Sending via UpdateKnownTxStatus, skipping the write if the tx concurrently reached a genuinely terminal status (Completed/Invalid/DoubleSpend).
- Added retryExternallyRejectedTx (pkg/storage/internal/actions/process_external_status.go) implementing this, and wired it into applyExternalRejected in place of the old log-and-return no-op.
- Updated TestExternalStatusRejectedWithoutEvidenceStaysRetryable → TestExternalStatusRejectedWithoutEvidenceIsMadeRetryableAgain to assert the tx moves to Sending and a TxSynchronizedStatus is emitted.
- Added TestExternalStatusRejectedWithoutEvidenceRacingCompletionLeavesCompletedTxUntouched, mirroring the existing double-spend race test, to cover the concurrent-completion guard.

## Why It Was Necessary

Once a tx advances past Sending (e.g. to unmined via a SENT_TO_NETWORK/ACCEPTED_BY_NETWORK event), an Arcade REJECTED with no double-spend evidence was logged as "keeping the tx retryable" but left the stored status untouched at unmined. SendWaitingTransactions (process_send_waiting.go) only rescans Unsent/Sending — unmined is unreachable by any retry path, so the tx was silently dead while local state still claimed it was in-flight. We hit this in practice via Arcade's 300-ancestor limit rejecting chained txs after they'd already advanced past Sending. The fix reuses the exact status transition (Sending) that the normal broadcast flow already uses for its own ServiceError verdict, so the tx re-enters the existing retry sweep instead of inventing new state. Confirmed double spends still go through the pre-existing terminal failExternallyRejectedTx path — unchanged.

## Testing Performed

- go build ./..., go vet ./pkg/storage/... ./pkg/monitor/... — clean.
- go test ./pkg/storage/... (full suite) and go test ./pkg/monitor/... — all pass.

## Impact / Risk

- Scoped to the ServiceError branch of applyExternalRejected; the Success (false positive) and DoubleSpend (confirmed) branches are untouched.
- Low risk: reuses an existing, already-tested status transition (Sending) and an existing guarded-update pattern (skip-list on terminal statuses) rather than introducing new state or altering the Transaction-row status.
- Txs previously stuck at unmined with no evidence will now be picked up by the next SendWaitingTransactions sweep and rebroadcast.

### DIFF
--- a/pkg/storage/internal/actions/process_external_status.go
+++ b/pkg/storage/internal/actions/process_external_status.go
@@ -2,6 +2,7 @@ package actions
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log/slog"
 	"strings"
@@ -13,6 +14,7 @@ import (
 
 	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/internal/storage/entity"
 	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/internal/storage/history"
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/internal/storage/repo"
 	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/tracing"
 	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/wdk"
 )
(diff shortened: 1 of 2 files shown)


mrz1836 (approved):
LGTM

PR #998, opened 2026-08-14
feat(examples): add BEEF perf torture harness (perf_torture)

sirdeggen:
## What Changed
Two parts: the perf/torture harness this PR started with, and a production bug fix the harness investigation surfaced.

### 1. `examples/perf_torture/` — BEEF hot-path torture harness
- `perf_torture.go` — mainnet end-to-end run: N createActions (single 0-sat `OP_FALSE OP_RETURN deadbeef` output) broadcast through Arcade (`arcade-v2-us-1.bsvblockchain.tech`) with `change_basket.number_of_desired_utxos=1`, every tx tracked to MINED (per-tx CSV, p50/p95 summary), `net/http/pprof` on `localhost:6060`. Flags: `-txid`/`-vout` (funding internalization), `-n`, `-interval`, `-fee-satkb`, `-list-beef` (replicates the client's listOutputs-with-entire-transactions-per-operation pattern), `-mined-timeout`, `-poll`.
- `profile.sh` — 30s CPU profile capture with `-diff_base` report against the previous capture.
- `beefbench/` + `bench.sh` — in-memory benchmarks of the exact hot paths (chain-depth MergeTransaction, repeated merge, `Beef.Bytes`, compound-bump `ComputeRoot`, same-height `MergeBump`) with benchstat comparison.

Flame graph captured from this harness on mainnet (30s window, ~10 tx/s creation at unmined chain depth 400–700) — the `recursiveBuildValidBEEF → MergeTransaction` recursion towers on the right replicate the production flame graph that motivated this work:

perf_torture mainnet flame graph

### 2. `fix(storage)`: externally rejected txs were kept "retryable" but never retried
Reported by the wallet developer team (hit via Arcade's 300-ancestor limit). An Arcade SSE `REJECTED` that `confirmDoubleSpends` could not confirm as a double spend logged "keeping the tx retryable" but left the tx `unmined` — a status no retry path can reach:
- `send_waiting` only scans `{unsent, sending}`;
- the proof-poll timeout backstop never fired either: a rejected tx is NOT_FOUND on the network, so `filterTxsByConfirmationDepth` dropped it **before** its attempts counter grew, and `ApplyProofTimeouts` could never requeue or terminally fail it.

The tx was dead while the wallet reported it in-flight. Two complementary fixes:
1. `applyExternalRejected` now requeues via a new guarded `RequeueKnownTxForRebroadcast` repo method (`unmined → unsent`, attempts reset, `rebroadcast_attempts+1`, mirroring `ApplyProofTimeouts`; terminal `invalid` once `MaxRebroadcastAttempts` is exhausted). Concurrent completion is protected by a fromStatuses guard inside the UPDATE; already-queued txs are untouched. Each retry of an ancestor-limit rejection lands after more ancestors mined, so it eventually clears.
2. The status-sync task counts network-NOT_FOUND candidates as failed proof attempts, so `ApplyProofTimeouts` eventually reaps any tx absent from the network regardless of how it got wedged (SSE rejection, mempool eviction).

## Why It Was Necessary
A client (settlement workload, 1-desired-UTXO torture config) degraded to ~18 tx/h with ~3 CPU cores pegged in BEEF merging. Analysis of their flame graph + goroutine dump traced two towers: `CreateAction → assembler.ToAtomicBEEF` (~135-deep merge recursion) and listOutputs-with-BEEF per operation (~190-deep). Root cause is the per-createAction full unmined-ancestry rebuild (`recursiveBuildValidBEEF`) amplified by go-sdk v1.3.3 (uncached `Transaction.TxID()`, no visited-set in `MergeTransactionWithTxid`, non-memoized `ComputeRoot`/`GetOffsetLeaf`). A reproducible rig + iteration loop is needed before optimizing (README has the target list). The retryable-wedge fix makes the failure mode the same client hits at Arcade's ancestor limit self-heal instead of stranding txs.

## Testing Performed
- `go build`, `go vet`, `golangci-lint` (0 issues) on changed packages; full `go test ./pkg/storage/... ./pkg/internal/storage/...` green.
- New/updated tests: `TestExternalStatusRejectedWithoutEvidenceIsRequeuedForRebroadcast` (was ...StaysRetryable — semantics intentionally changed), `TestExternalStatusRejectedWhileAlreadyQueuedIsNoOp`, `TestExternalStatusRejectedExhaustsRebroadcastBudget` (incl. review-cascade to failed), `TestSynchronizeTxRequeuesNetworkAbsentTxAfterMaxAttempts`.
- Harness field-tested on mainnet: 3,000+ createActions across runs, proofs attached to MINED by the in-process monitor; flame graphs captured at depth 400–700 (image above). Field notes: create latency grows with unmined depth (84ms → ~190ms by depth ~700); at unmined depth 1000 createAction hard-fails on `maxDepthOfRecursion` (known_tx.go:29) until the next block — recorded in the README as a follow-up target along with the 1000-level error-wrap allocation churn.

## Impact / Risk
- Storage fix changes SSE REJECTED handling semantics: previously-stranded txs now rebroadcast (bounded by `MaxRebroadcastAttempts`, default 0 = unlimited) or fail terminally once the budget is exhausted. POST-path rejection handling is unchanged.
- Example + benchmarks are additive; no CI impact beyond building them. The example spends real mainnet satoshis when funded (documented); keys/config are generated locally and git-ignored.

## Notifications
- Wallet developer team: the "REJECTED kept retryable but never retried" report is fixed here (both the SSE path and the missing sync backstop).


mrz1836 (changes requested):
Fix sonarcloud warning and linter issue please - ty!

sirdeggen:
### CI note: SonarCloud suppression in `examples/perf_torture/profile.sh`

The initial run failed the `new_security_rating` gate on two `shell:S5332` findings ("clear-text protocol") for the `http://` pprof URLs in `profile.sh`. These are suppressed with `NOSONAR(shell:S5332)` and an inline justification, following the existing repo convention (see the `NOSONAR(go:S2068)` usages in `pkg/storage/v1adapter/handler.go`).

Justification: Go's `net/http/pprof` exposes no TLS variant, and the script targets a local dev-only debug endpoint — default `localhost:6060`, overridable via `PPROF_HOST`. It never talks to a production or public host, and the only payload is CPU sample data.

Quality gate is now green on all five conditions (`new_security_rating` back to `1`).

sirdeggen:
Resolved merge conflicts against latest `main` (`#995`, `#997`, `#996`) and addressed the SonarCloud leak-period issues.

**Merge**
- Kept this branch's `RequeueKnownTxForRebroadcast` path (budgeted demotion to `unsent`) over `#997`'s simpler pull-back to `Sending`.
- Kept `main`'s pre-broadcast `aborter` / `release` wiring from `#982`.

**Sonar**
- `go:S1186` — documented why the wallet storage-factory cleanup is a no-op.
- `go:S107` / `go:S3776` in `perf_torture.go` — packed run options into a struct and split the create/list/pace loop.
- `go:S3776` in `broadcastTxs` — extracted partition / script-verify / result-collection helpers.

Local: `go test` green for `pkg/storage/internal/actions`, `pkg/internal/storage/repo`, and the ExternalStatus/ProcessAction/Synchronize/SendWaiting subset of `pkg/storage`.

Issue #343, opened 2025-07-08
[FEATURE] Wallet Examples : add config loader standard solution to example setup


cain4chain:
ref : https://github.com/bsv-blockchain/go-wallet-toolbox/pull/291#discussion_r2185436204

loader solution : https://github.com/bsv-blockchain/go-wallet-toolbox/blob/f55b7debc10a42c964f3e763d570aa59393de6e1/internal/config/loader.go#L35

PR #1003, opened 2026-08-18
fix(beef): complete txid-only ancestry instead of rejecting it

sirdeggen:
## What was happening

`createAction` rejected input BEEFs with a bare `provided beef is not valid` for transactions **storage was holding itself**, whenever the ancestry was still unconfirmed. It self-healed on the next block, because the subject transaction then carried its own merkle proof and the bare-txid ancestors above it stopped mattering.

Storage proved every bare txid was known (`AllKnownTxsExist`) and then handed the still-stubbed BEEF to a validator that accepts only a merkle proof as an anchor. Any raw transaction above a bare txid therefore invalidated the whole BEEF. The `knownTxIds` optimisation (#996) put bare txids on the wire; nothing completed them again before verification.

## The go-sdk defect underneath

`transaction/beef.go` `readBeefTx` allocates every entry as `BeefTx{DataFormat: f, Transaction: &Transaction{}}` and never clears it for `TxIDOnly`. `readBeefTxFull` then wires that empty object into later transactions as `input.SourceTransaction`, with no check on the source's data format.

So a round trip through the wire format turns a bare txid into **a bare txid plus an empty transaction** (version 0, no inputs, no outputs). Three consequences, all observed:

- `TransactionInput.SourceTxOutput` indexes `Outputs` without a bounds check → the storage server **panics** during script verification;
- `Beef.MergeTransaction` walks `SourceTransaction` pointers → the empty transaction is merged as a graph entry of its own, under the txid of ten zero bytes (`f702453d…`). That is the "floating empty transaction" found in every stored BEEF from the stress runs — it is **not** injected by the client;
- anything treating a non-nil `SourceTransaction` as "already hydrated" keeps the empty parent instead of the real one, so hydrating a bare txid does not help on its own.

The defect predates #343; #996 only made it reachable. `pkg/internal/sdkbeef` works around it until an upstream fix lands, and every BEEF parse/merge site now routes through it.

## Storage

- **`hydrateUnanchoredAncestry`** runs before `VerifyBeef` and reads back from storage every source an unproven raw transaction spends that the BEEF cannot supply — bare txids *and* ancestors absent altogether, so client-side truncation heals too. Bare txids nothing spends stay bare, so #996's win survives. Unknown to storage → an error that **names the txid**.
- `missingProofError` joined `msgParts` where `txIDs` belongs, so the txids the error exists to report never appeared in it.
- `xinputDefOnUnknownUTXO` passed `WithStatusesToFilterOut(readyToBeInputProvenTxStatuses...)` — **inverted**: it excluded exactly the usable statuses, so every stubbed unknown UTXO failed with "is not known to storage" immediately after storage confirmed it knows it.
- `hydrateInput` dereferenced a nil `Transaction` for a bump-covered bare txid.

## Wallet

- **`CreateAction` ran the beef-party merge and txid-only resolution after `ProcessAction`**, so a party-graph miss returned an error for a transaction that was already broadcast — a settlement engine reacting to that either rebuilds and double-spends its own inputs or writes off an operation that succeeded. Both steps now log and fall back to the BEEF storage returned. Before #996 this block ran only under `ReturnTXIDOnly`, where `result.Tx` is nil, so it never fired.
- **`BeefParty.Lease`** defers the oversize reset while an action is between advertising its known txids and resolving the reply built from them — the reset used to strand replies in flight ("tx with only txid not found in beef party"). An `EmergencyResetFactor` ceiling keeps sustained concurrency from reintroducing unbounded growth.
- `verifyReturnedTxIDOnly` now detects the ancestry `FindAtomicTransactionByHash` silently skips, instead of shipping a holed BEEF whose hole only surfaces later as a storage-side rejection.
- **`AutoKnownTxids` was dead code** — declared, documented and settable, never read. Both advertise sites now honour it. Default flipped to `true` so behaviour is unchanged, and there is finally a way to turn the optimisation off (the fastest way to rule this path out when diagnosing a BEEF failure).

## Monitor

- `requeueExternallyRejectedTx` accepted `unconfirmed` as requeueable. That status means a proof has already arrived, so a **mined** transaction could be pulled back to rebroadcast and, once the rebroadcast budget ran out, failed terminally as `invalidTx`.

## Tests

- `pkg/storage/provider_create_action_known_txids_test.go` — reproduces the exact failure and pins the named-txid rejection for a genuinely unknown ancestor.
- `pkg/internal/sdkbeef/sanitize_test.go` — pins the go-sdk reader defect and the repair.
- `pkg/wdk/beef_party_bound_test.go` — two cases: the reset stays deferred while a window is open, and the ceiling resets anyway.

Full suite and `golangci-lint run ./...` clean locally.


sirdeggen:
Draft until the upstream fix lands: **bsv-blockchain/go-sdk#345** — `fix(transaction): stop attaching an empty tx to a TxIDOnly BEEF entry`.

That PR fixes the reader defect at its source (don't allocate the placeholder for `TxIDOnly`; don't link a source that has no transaction; bounds-check `SourceTxOutput` instead of panicking), with three regression tests that fail on `master` and pass with the change.

Once #345 merges and the pin here moves past it, `pkg/internal/sdkbeef` becomes a no-op and can be deleted, with its call sites reverted to `transaction.NewBeefFromBytes` / `MergeBeefBytes`. Everything else in this PR stands on its own and is unaffected by the upstream outcome.

Worth keeping `sdkbeef` for one release after the pin bump if any peer wallet may still be serialising BEEFs with an older go-sdk — the empty transaction travels inside merged bytes, so it can arrive from outside this process.

sirdeggen:
Upstream fix merged as **bsv-blockchain/go-sdk#345**, released as **v1.3.4**. Out of draft.

`bfcecb2f` bumps the pin (root and `manual_tests`) and retires the workaround:

- `pkg/internal/sdkbeef` deleted; all seven call sites back to `transaction.NewBeefFromBytes` / `MergeBeefBytes` / `NewBeefFromAtomicBytes` directly.
- The tests that pinned the reader defect live upstream now, where the fix does.
- Replaced here by a contract test on the invariant hydration depends on — `TestHydrateBEEFHandlesABareTxIDReadBackFromTheWire` in `pkg/internal/txutils`. A pin that regressed the invariant fails there and names the cause, instead of resurfacing as an opaque failure inside `createAction`.

No peer wallets on older go-sdk builds in this estate, so the empty transaction has no route in from outside the process and there is nothing left for the workaround to defend against.

Full `go test ./...` and `golangci-lint run ./...` clean on the bumped pin. The `provided beef is not valid` regression test passes on the real fix rather than on the workaround, which is the result that actually matters here.

pawellewandowski98 (approved):

PR #1009, opened 2026-08-25
feat: add background broadcaster config, sort tx before broadcast, fallback for beef party

pawellewandowski98:
## What Changed

- **Delayed-broadcast queue is now sizeable by the operator.** New `defs.BackgroundBroadcaster` (`workers`, `channel_size`), wired through `infra` config as a `background_broadcaster:` block and through the provider as `storage.WithBackgroundBroadcaster`. Resolution is per field: an explicit value wins over the sizing derived from the throughput strategy, and a zero falls through to the package defaults — so a deployment not on the throughput strategy can still widen just the queue. Validation caps it at 1024 workers / 100k queue items.
- **Queue observability.** `wallet.broadcaster.overflow_total` counter plus depth/capacity observable gauges, and a windowed overflow warning that the `send_waiting` sweep flushes, so the tail of a one-off burst is never silently dropped.
- **Monitor task deadline derived from the configured interval** instead of `gocron.Job.NextRun()`. `ActiveTask` now carries `Interval`, and `contextWithTimeout` budgets from it.
- **`send_waiting` sweep now runs to a budget.** An EWMA of observed batch durations decides whether another batch fits before the deadline; the remainder is deferred to the next run as one summary line. Page-read failures after earlier pages succeeded no longer sink the whole sweep, and context errors are reported as deferrals rather than joined into the returned error.
- **Parents-first ordering applied everywhere a batch is posted.** `parentsFirst` promoted out of the broadcaster into `txutils.ParentsFirst` and applied to `broadcastTxs` as well; new `retainClaimed` keeps the caller's ordering instead of adopting whatever order the claim query returned; sweep batches are collected oldest-first with a `Paging.ThenBy` tie-break so OFFSET paging is stable.
- **BEEF-party resolution failures now degrade instead of failing the call.** `listOutputs` and `signAction` fall back to the BEEF storage sent, the way `createAction` already did. The `signAction` guard was also corrected — it paired `result.Tx != nil` with `ReturnTXIDOnly`, which the result mapping makes mutually exclusive, so the block had never executed.

## Why It Was Necessary

- The queue size was hard-wired to the throughput strategy's derivation. A deployment running bursty traffic outside that strategy had no way to widen it, and everything that overflowed fell through to the `send_waiting` cron, which drains far more slowly. Overflow was also invisible: nothing counted it, so an undersized queue looked like unexplained latency.
- `NextRun()` is not a reliable budget source. gocron can start a run a few milliseconds *before* its scheduled instant and still report that same instant as the next run — subtracting it from `now` yielded single-digit milliseconds instead of the interval. Every batch in the run then failed instantly on an already-expired context: observed in production as hundreds of identical errors from a sweep that lasted 34ms in total.
- Without a budget, a sweep that overran its slot died mid-batch, and a batch killed after `ClaimKnownTxsForBroadcast` is left marked `sending`/`was_broadcast` without ever having been posted.
- Arcade forwards to Teranode in receive order, so a child posted before the unconfirmed parent it spends is rejected. `broadcastTxs` was posting in whatever order the claim query happened to return — typically by txid hex, which says nothing about dependency order.
- Under the load of the 1000-trade test the client reported ~24 `tx with only txid not found in beef party` errors, all on `listOutputs`. The shared BEEF graph has a hard ceiling that resets it even while leases are open, which leaves an in-flight reply asking for transactions the graph no longer holds. Resolving those stubs is a wire optimisation, not a correctness requirement — storage's own BEEF is a valid reply, and storage completes bare txids it knows when that BEEF comes back as an input. `createAction` already degraded this way; the other two did not.

## Testing Performed

- `go test ./pkg/wallet/... ./pkg/storage/...` — green, including both the SQLite and remote-storage variants of `WalletTestSuite`.
- New unit coverage: `pkg/defs/background_broadcaster_test.go` (validation bounds), `pkg/monitor/task_budget_test.go`, `pkg/storage/internal/actions/broadcaster_overflow_test.go`, `broadcaster_sizing_test.go`, `process_send_waiting_budget_test.go`, `pkg/internal/txutils/parents_first_test.go`, `pkg/storage/provider_send_waiting_test.go`.
- New regression tests for the fallback: `TestListOutputsFallsBackWhenAStubCannotBeResolved` and `TestSignActionFallsBackWhenAStubCannotBeResolved`. Both seed the beef party with a bare txid under a merkle proof, which is what the known-txid list is built from but not what resolving a reply accepts — reproducing the client's failure without racing threads.
- Both fallback tests were verified to discriminate, not just pass: the `listOutputs` one fails on the unpatched code with `an unresolvable stub must not fail the call`, and the `signAction` one fails against a guard-fixed-but-fallback-less build with `an unresolvable stub must not fail an already broadcast action`.
- Confirmed the `signAction` fallback actually fires rather than passing vacuously — the run logs `Could not resolve the txid-only entries in the returned BEEF after broadcast`.

## Impact / Risk

- **Config is backwards compatible.** Both new fields default to zero, which means "not configured" and preserves the sizing every existing deployment gets today.
- **The fallbacks trade strictness for availability.** A wallet that cannot resolve a stub now returns storage's BEEF and logs a warning instead of erroring. Watch for `Could not resolve the txid-only entries in the returned BEEF` in production — a sustained rate means the graph is being reset under load and the knownTxids optimisation is not paying off, which is worth investigating separately even though it no longer breaks callers.
- **Ordering changes affect every delayed broadcast**, not just batched ones. `retainClaimed` is a no-op when the claim accepted everything (the common case), and `ParentsFirst` is stable, so creation order survives wherever there is no dependency to respect.
- **`Paging.ThenBy` touches shared paging code** used beyond the sweep. It is opt-in — an empty `ThenBy` leaves the ordering exactly as before.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -73,6 +73,9 @@ manual_tests/*.sqlite-journal
 manual_tests/*.sqlite-wal
 manual_tests/*.sqlite-shm
 
+# Runtime / container log dumps (e.g. `docker compose logs > token-wallet.log`)
+*.log
+
 # Fuzz Tests
 testdata
 
--- a/.golangci.json
+++ b/.golangci.json
@@ -285,7 +285,8 @@
         ],
         "concurrency": 8,
         "issues-exit-code": 1,
-        "tests": true
+        "tests": true,
+        "timeout": "10m"
     },
     "severity": {
         "default": "warning",
--- a/infra-config-docker.yaml
+++ b/infra-config-docker.yaml
@@ -105,6 +105,18 @@ monitor:
       interval_seconds: 600
       start_immediately: false
 
+# --- Delayed broadcast queue ---
+# createAction hands transactions to this queue; the send_waiting cron is only a
+# fallback for whatever does not fit, and it drains far more slowly.
+#   workers      0 = derive from utxo_management.throughput.target_tps, else 10.
+#                    Throughput is roughly workers / (time of one post).
+#   channel_size 0 = 1000. This buffers bursts; it does not raise sustained
+#                    throughput. 5000 absorbs a burst of a few thousand
+#                    transactions without spilling to the cron.
+background_broadcaster:
+  workers: 0
+  channel_size: 5000
+
 # --- Change Basket Defaults ---
 change_basket:
   max_change_outputs_per_tx: 8
--- a/infra-config.example.yaml
+++ b/infra-config.example.yaml
@@ -1,3 +1,6 @@
+background_broadcaster:
+  channel_size: 0
+  workers: 0
 bsv_network: main
 change_basket:
   max_change_outputs_per_tx: 8
--- a/pkg/defs/background_broadcaster.go
+++ b/pkg/defs/background_broadcaster.go
@@ -1,0 +1,52 @@
+package defs
+
+import "fmt"
+
+const (
+	// MaxBroadcasterWorkers bounds the delayed-broadcast worker pool. Past this
+	// point the database connection pool and the broadcast service, not the
+	// number of goroutines, decide throughput.
+	MaxBroadcasterWorkers = 1024
+
+	// MaxBroadcasterChannelSize bounds the delayed-broadcast queue. Every queued
+	// item holds a parsed BEEF - a structure, not raw bytes - so an oversized
+	// queue is an out-of-memory risk rather than a harmless buffer. The cap is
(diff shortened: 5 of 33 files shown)


dzolt-4chain (approved):

mrz1836 (changes requested):
I am fixing the gitleaks issue :(

PR #1014, opened 2026-09-01
fix(beef): let callers stop shipping ancestry storage already holds

galt-tr:
## The problem

`createAction` persists the caller's `inputBEEF` verbatim onto `bsv_known_txes.input_beef`. A caller spending its own outputs therefore stores a copy of ancestry storage is already holding — and BEEF prunes only back to a *mined* ancestor, so on a chain that is not confirming it never prunes and the blob grows without bound.

Measured on a 2,000-trade load run against teratestnet (which produced 7 blocks in 65 minutes, leaving 2,948 unmined transactions against 1 completed):

| | |
|---|---:|
| mean `input_beef`, start → end of one run | 1,892 B → **1,038,633 B** (549×) |
| `raw_tx` on those same rows | 349 B |
| `bsv_known_txes` | **10 GB** from 10,013 rows |
| throughput decay over the run | 8.03 → 0.27 trades/s |

Mints were unaffected throughout (~200 ms, 6-byte BEEF) because they are the one operation with no provided inputs. Transfers and burns degraded ~15×.

## Why a caller could not simply stop supplying it

**1. The assembler had one place to look.** `toTxInputFromArgs` resolved each provided input's `SourceTransaction` only from the `inputBeef` echoed back in the storage response, and `ToAtomicBEEF` fails outright on a nil source transaction. So the persisted blob had to stay large purely so that lookup would succeed — even though storage already attaches the parent's raw bytes to the input itself (`StorageCreateTransactionSdkInput.SourceTransaction`, read from `raw_tx` when `IncludeInputSourceRawTxs` is set). It now falls back to those bytes.

**2. Hydration pulled the whole graph back in.** `hydrateUnanchoredAncestry` called `GetBEEFForTxIDs` without `WithDirectSourcesOnly`, so it merged every ancestor's stored `input_beef` via `recursiveBuildValidBEEF` — and the result is then persisted, handing the next transaction a larger blob again. The validator needs one generation, and the loop already re-checks and fetches again if a further one turns out to be required. `broadcastTxs` already asks for direct sources only; this makes `createAction` consistent with it.

An absent `inputBeef` also becomes legitimate rather than a parse error, which is what lets a caller decline to ship ancestry at all.

## Compatibility

The `inputBeef` is still **preferred**, so behaviour is unchanged wherever it carries the parent; the fallback engages only where assembly would previously have failed. Where neither route has the parent, the error now names the outpoint instead of surfacing later as the opaque "every signable transaction input must have a source transaction".

## Result

With both changes, a caller passing `InputBEEF: nil` for inputs storage already knows stores txid stubs instead — 72 bytes for a two-input transfer, 39 for a one-input burn — flat against wallet history rather than growing with it.

Re-running the same 2,000 trades, same workload, same 10,013 transactions:

| | before | after |
|---|---:|---:|
| mean `input_beef` | 1,038,633 B | **34 B** |
| `bsv_known_txes` | 10 GB | **7,408 kB** |
| schema total | 10 GB | **45 MB** |
| wall clock | 59m 53s | **14m 09s** |
| trades/s | 0.56 | **2.35** |
| failures | 0 | 0 |

The mean stayed flat for the whole run (4,096 → 17 → 6 → 5 bytes) rather than rising.

One trade worth noting: index scans against `bsv_known_txes` went **up**, 316,105 → 28,752,603. Each now returns a single row with no TOAST to detoast, where before a handful of multi-megabyte reads dominated — a clear net win here, but it is a trade rather than free.

## Tests

Four new cases in `create_action_source_tx_test.go` covering: assembly with an empty `inputBeef`, that the result still produces an atomic BEEF, that the `inputBeef` is preferred when present, and the named-outpoint error. All four verified to fail without the change (`cannot parse inputBeef: EOF`). Full `./pkg/...` suite passes with zero failures.

### DIFF
--- a/pkg/internal/assembler/create_action_source_tx_test.go
+++ b/pkg/internal/assembler/create_action_source_tx_test.go
@@ -1,0 +1,151 @@
+package assembler_test
+
+import (
+	"testing"
+
+	"github.com/bsv-blockchain/go-sdk/chainhash"
+	"github.com/bsv-blockchain/go-sdk/script"
+	"github.com/bsv-blockchain/go-sdk/transaction"
+	sdk "github.com/bsv-blockchain/go-sdk/wallet"
+	"github.com/stretchr/testify/require"
+
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/internal/assembler"
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/internal/fixtures/testusers"
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/wdk"
+	"github.com/bsv-blockchain/go-wallet-toolbox/pkg/wdk/primitives"
+)
+
+// A caller-provided input needs its parent transaction to assemble, and until
+// now the only place the assembler looked was the inputBeef echoed back in the
+// storage response. That made the persisted blob load-bearing for assembly: a
+// caller could not trim the ancestry it ships even when storage already held
+// every parent, because assembly would then fail with
+// "every signable transaction input must have a source transaction".
+//
+// Storage populates StorageCreateTransactionSdkInput.SourceTransaction for
+// exactly these inputs, so the assembler can fall back to it.
+
+// givenParentTransaction returns a spendable parent and its serialised bytes.
+func givenParentTransaction(t *testing.T) (*transaction.Transaction, []byte) {
+	t.Helper()
+
+	lockingScript, err := script.NewFromHex("76a914aabbccddeeff00112233445566778899aabbccdd88ac")
+	require.NoError(t, err)
+
+	parent := transaction.NewTransaction()
+	prev, err := chainhash.NewHashFromHex(
+		"0000000000000000000000000000000000000000000000000000000000000001")
+	require.NoError(t, err)
+	parent.AddInput(&transaction.TransactionInput{
+		SourceTXID:       prev,
+		SourceTxOutIndex: 0,
+		SequenceNumber:   transaction.DefaultSequenceNumber,
+	})
+	parent.AddOutput(&transaction.TransactionOutput{Satoshis: 1000, LockingScript: lockingScript})
+
+	return parent, parent.Bytes()
+}
+
+// givenResultWithProvidedInput builds the storage response for a single
+// caller-provided input, with whatever inputBeef the caller is meant to have
+// shipped back.
+func givenResultWithProvidedInput(
+	t *testing.T,
+	parent *transaction.Transaction,
+	parentRaw []byte,
+	inputBeef []byte,
+	attachSourceTx bool,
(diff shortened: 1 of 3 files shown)


pawellewandowski98 (approved):

galt-tr:
Thanks for the txid check — that's a real gap in what I pushed, and I've rebased onto it rather than around it.

Added a second commit with two further findings from profiling the wallet under load, once the stored-BEEF growth from the first commit stopped masking them.

## Ancestry reads: one query per generation, not one per ancestor

`recursiveBuildValidBEEF` reads any ancestor it wasn't handed with an individual `First()`. Two generations were pre-read in bulk, which covers a build whose subjects are already proved — but not one walking back through unmined history, and on a chain that isn't confirming that's every build.

Profiling the wallet at 2,000 trades: `recursiveBuildValidBEEF` was **24.9% of all CPU, and 80.4% of that was inside `gorm.(*DB).First`**. The wallet sat at 446% CPU while postgres stayed **flat at 27%** for the entire run. The cost was never the query — it was ORM and round-trip overhead paid ~2,870 times per transaction.

`preFetchAncestry` walks breadth-first, one query per generation, bounded by the same `maxDepthOfRecursion` the recursion already enforces and stopping as soon as a generation adds nothing. The rows read are exactly those the recursion would have read; only the statement count changes.

Worth noting so it isn't a surprise in review: **index scan counts don't drop** (28.7M → 33.7M), because each value in an `IN` list still costs a probe. That's consistent with the profile — the probe was never the expensive part.

## Join table indexes

`bsv_output_tags` and `bsv_transaction_labels` are both modelled with a composite primary key led by the *foreign* key, while every lookup filters the other way, by tag or label name. Measured after 2,000 trades: `bsv_output_tags` took **19,571 sequential scans reading 92,989,035 tuples** from an 8,000-row table.

Declared on the models following the existing partial-index precedent on `idx_user_utxos_claim`. Sequential scans fall to 3, tuple reads to zero.

**On its own this did not improve throughput**, and I'd rather say so than imply otherwise — an 8,000-row table in cache is cheap to scan however often you do it. It's here because the cost grows with history and the fix is free, not because it moved the benchmark.

## Measured

2,000 trades, 16 in flight, fresh schema, zero failures:

| arm | trades/s | lifecycle p50 | wall clock |
|---|---:|---:|---:|
| stored BEEF bounded (commit 1) | 2.33 | 12.8 s | 14m 20s |
| + join table indexes | 1.81 | 20.1 s | 18m 23s |
| + ancestry batching | **3.89** | **9.2 s** | **8m 34s** |

Against the original baseline: **0.56 → 3.89 trades/s**, and 59m 53s → 8m 34s.

The middle row is a single run that went the wrong way. It's within the range where a repeat is warranted before concluding anything, and the batching row below it carries the same indexes — I didn't want to quietly drop an inconvenient data point.

Full suite passes at every step.

PR #1015, opened 2026-09-02
perf(beef): Read ancestry one generation per query, and index the join tables

galt-tr:
Follow-up to #1014. Two costs that only became visible once that change stopped the stored-BEEF growth from masking them, both found by profiling the wallet process under sustained load.

## Ancestry reads — one query per generation, not one per ancestor

`recursiveBuildValidBEEF` reads any ancestor it wasn't handed with an individual `First()`. Two generations were pre-read in bulk, which covers a build whose subjects are already proved — but not one walking back through unmined history, and on a chain that isn't confirming that's every build.

Profiling at 2,000 trades: `recursiveBuildValidBEEF` was **24.9% of all CPU, and 80.4% of that was inside `gorm.(*DB).First`**. The wallet process sat at 446% CPU while postgres stayed **flat at 27%** for the whole run. The cost was never the query — it was ORM and round-trip overhead paid roughly **2,870 times per transaction**.

`preFetchAncestry` walks breadth-first, reading each generation in one query. Bounded by the same `maxDepthOfRecursion` the recursion already enforces, stops as soon as a generation adds nothing, and skips deeper generations entirely under `DirectSourcesOnly` since parents are terminal there. The rows read are exactly the ones the recursion would have read; only the statement count changes.

`parentIDsOf` replaces `directParentIDs`, taking one generation rather than the whole accumulated map — otherwise each round rescans everything already fetched.

Flagging for review so it isn't a surprise: **index scan counts do not drop** (28.7M → 33.7M), because each value in an `IN` list still costs a probe. That's consistent with the profile — the probe was never the expensive part.

## Join table indexes

`bsv_output_tags` and `bsv_transaction_labels` are both modelled with a composite primary key led by the *foreign* key, while every lookup filters the other way round, by tag or label name. The primary key can't serve that, so the lookup scans the whole join table — and both tables gain a row per name per row, forever.

Measured after 2,000 trades:

| | sequential scans | tuples read |
|---|---:|---:|
| `bsv_output_tags` (8,000 rows) | 19,571 | 92,989,035 |
| `bsv_transaction_labels` | 28,982 | — (154 per index scan) |

Declared on the models, following the existing partial-index precedent on `idx_user_utxos_claim`. Sequential scans fall to 3 and the 93M tuple reads to zero.

**On its own this did not improve throughput**, and I'd rather say so plainly — an 8,000-row table sitting in cache is cheap to scan however often you do it. It's included because the cost grows with history and the fix is free, not because it moved the benchmark.

## Measured

2,000 trades, 16 in flight, fresh schema per arm, zero failures throughout:

| arm | trades/s | lifecycle p50 | wall clock |
|---|---:|---:|---:|
| #1014 alone | 2.33 | 12.8 s | 14m 20s |
| \+ join table indexes | 1.81 | 20.1 s | 18m 23s |
| \+ ancestry batching | **3.89** | **9.2 s** | **8m 34s** |

Against the pre-#1014 baseline that is **0.56 → 3.89 trades/s**, and 59m 53s → 8m 34s.

The middle row is a single run that went the wrong way. It sits within the range where a repeat is warranted before concluding anything, and the batching row below it carries the same indexes — I didn't want to quietly drop an inconvenient data point.

Full `./pkg/...` suite passes, zero failures, on top of the merged #1014.

### DIFF
--- a/pkg/internal/storage/database/models/output_tags.go
+++ b/pkg/internal/storage/database/models/output_tags.go
@@ -18,14 +18,13 @@ type OutputTag struct {
 	//
 	// The primary key cannot answer that - its leading column is output_id - so
 	// without this the lookup falls back to scanning the whole join table, and
-	// the table gains a row per tag per output forever. Measured on a wallet
-	// after 2,000 trades: 19,571 sequential scans reading 92,989,035 tuples from
-	// 8,000 rows, the single largest source of tuple reads in the schema.
+	// that table gains a row per tag per output for the life of the wallet.
 	//
-	// Column order follows the predicate: equality columns first, then output_id
-	// so the subquery is answered from the index alone. Partial on deleted_at
-	// because the soft delete appends `deleted_at IS NULL` to every read.
-	OutputID  uint   `gorm:"primary_key;index:idx_output_tags_name_user,priority:3,where:deleted_at IS NULL"`
-	TagName   string `gorm:"primary_key;index:idx_output_tags_name_user,priority:1,where:deleted_at IS NULL"`
-	TagUserID int    `gorm:"primary_key;index:idx_output_tags_name_user,priority:2,where:deleted_at IS NULL"`
+	// Column order follows the predicate: the equality columns first, then
+	// output_id so the subquery is answered from the index alone. Partial on
+	// deleted_at because the soft delete appends `deleted_at IS NULL` to every
+	// read.
+	OutputID  uint   `gorm:"primaryKey;index:idx_output_tags_name_user,priority:3,where:deleted_at IS NULL"`
+	TagName   string `gorm:"primaryKey;index:idx_output_tags_name_user,priority:1,where:deleted_at IS NULL"`
+	TagUserID int    `gorm:"primaryKey;index:idx_output_tags_name_user,priority:2,where:deleted_at IS NULL"`
 }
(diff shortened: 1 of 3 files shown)


galt-tr:
All three addressed in the latest commit — all fair points.

**`primary_key` → `primaryKey`.** You're right that this is the convention; I'd carried the existing spelling forward without checking. Those two files were in fact the *only* remaining users of the legacy alias — 11 `primaryKey` against 6 `primary_key`, and all six were here. Now zero.

Since that tag is schema-affecting I didn't want to take it on trust, so I migrated a scratch schema with each version and diffed the result. Byte-identical — same composite primary keys, same indexes, same partial clauses:

```
CREATE UNIQUE INDEX bsv_output_tags_pkey ... (output_id, tag_name, tag_user_id)
CREATE INDEX idx_output_tags_name_user ... (tag_name, tag_user_id, output_id) WHERE (deleted_at IS NULL)
```

**Benchmark numbers in comments.** Agreed, and it's a habit worth correcting — those figures come from one wallet on one network and would quietly become wrong. The comments now carry the invariant only: that the walk costs one round trip per ancestor dominated by per-statement overhead, that pre-reading two generations doesn't cover a chain that isn't confirming, and that the primary key leads with the wrong column for the filter direction. Measurements stay in the PR description where they're dated and attributable.

I kept one non-obvious note in the code, since it's a property of the change rather than a measurement: index scan counts do *not* fall, because each value in an `IN` list still costs a probe. Without that a future reader could reasonably conclude the batching had regressed.

Full suite still passes, zero failures.

pawellewandowski98 (approved):

PR #1029, opened 2026-09-16
feat: cheaper TrustSelf ancestry read and optimize label/tag filters

pawellewandowski98:
## What Changed
- **GoBDK (libsecp256k1) enabled automatically in cgo builds** (new `pkg/internal/signaturebackend`, `pkg/storage/signature_backend.go`):
  - `gobdk.go` (`//go:build cgo && !nobdk && !ios && !android && (darwin || linux) && (amd64 || arm64)`) calls `bdk.InstallSignatureBackend()` in `init()`. go-sdk then signs and verifies secp256k1 through GoBDK, process-wide, before anything signs or verifies.
  - `builtin.go` (every other build, or `-tags nobdk`) does nothing; go-sdk keeps its built-in pure-Go implementation.
  - `pkg/storage` imports the package, so every wallet and storage built from this module gets the behaviour.
  - New `storage.SignatureBackend()` returns `"gobdk-libsecp256k1"` or `"go-sdk-builtin"`, so an application can log which backend its build uses.
  - `go.mod` adds `github.com/bitcoin-sv/bdk/module/gobdk` (indirect, via go-sdk's `transaction/bdk`).
- **Existence-only ancestry pre-fetch under TrustSelf** (`pkg/internal/storage/repo/known_tx_get_beef.go`):
  - `preFetchAncestry` now issues a single `SELECT tx_id … WHERE tx_id IN (…)` over the subjects (`preFetchExistence`) when `TrustSelf=known`, respecting `StatusesToFilterOut`. It no longer walks the ancestry loading `raw_tx`, `input_beef` and `merkle_path` for every generation.
  - The per-row fallback in `recursiveBuildValidBEEF` selects only `tx_id` under TrustSelf.
  - Txids passed in `KnownTxIDs` are removed from the pre-fetch frontier, since the build stubs them before any lookup.
- **"All" label and tag filters use one condition per name** (new `pkg/internal/storage/repo/name_filter.go`):
  - `QueryModeAll` changed from `name IN (…) GROUP BY id HAVING COUNT(DISTINCT name) = n` to one `id IN (…)` subquery per distinct name. "Any" mode is unchanged.
  - Shared helpers `labelledWith` / `taggedWith` replace the three copies in `ListAndCountActions` / `GetLabelsForSelectedActions` (`transactions.go`), `selectedActionsSubquery` (`outputs.go`) and the ListOutputs tag filter (`outputs.go`).
- **New tests:**
  - `pkg/internal/signaturebackend/{spend,gobdk,builtin}_test.go`
  - `pkg/storage/provider_label_query_mode_test.go`
  - `pkg/internal/storage/repo/known_tx_get_beef_trust_self_test.go`
  - `pkg/wallet/wallet_sign_action_beef_completeness_test.go`

## Why It Was Necessary
- Found while load-testing: 1000 mint → transfer → burn trades released at once, profiled with pprof and goroutine dumps.
- **Signature verification:**
  - By default go-sdk verifies `OP_CHECKSIG` through `crypto/ecdsa`'s generic big.Int path. That was **~49% of wallet CPU**.
  - go-sdk ships a GoBDK adapter, but it is opt-in. With GoBDK, a full P2PKH script execution goes from 253 µs to 36 µs, and wallet CPU for the same run fell from 1656 s to 805 s.
  - During development, GoBDK and the built-in verifier returned identical accept/reject verdicts on a differential suite (high-S, hybrid keys, malformed DER, r/s = 0 or ≥ N, off-curve keys) and on 400 rounds of signature mutation.
  - Selecting it from the build means applications built with cgo get it without code, and builds without cgo (static binaries, other platforms) keep compiling unchanged.
- **TrustSelf pre-fetch:**
  - With `TrustSelf=known`, `recursiveBuildValidBEEF` turns every transaction storage holds into a bare txid stub and never descends to its parents. The pre-fetch still downloaded the full unconfirmed ancestry, blobs included, and all of it was then thrown away.
  - With callers that set `TrustSelf` + `ReturnTXIDOnly`, `mergeAllocatedUTXOs` fell from **269 s to 3 s** of CPU in a 12-minute profile.
- **Label/tag filter:**
  - The grouped query must read, sort and group every association of every requested name.
  - Load test checks for `id=<operation>` **and** `kind=token` before each mint, transfer and burn. `kind=token` is on every action, so on a 51k-action wallet each check read 51,006 index rows and sorted them to disk.
  - Goroutine dumps showed ~63% of in-flight wallet requests waiting on these two queries.
  - Measured on that Postgres database, with the exact SQL the code generates, 3 runs each:

    | query | before | after |
    |---|---|---|
    | count (`ListAndCountActions`) | 35.8–40.3 ms | 0.061–0.078 ms |
    | page + labels (`GetLabelsForSelectedActions`) | 33.8–35.0 ms | 0.059–0.078 ms |

  - The new cost no longer grows with table size.
- **End to end,** the 1000-trade run went from ~9 min to 42 s.

## Testing Performed
- `go build ./...` passes in all three build modes: cgo, `CGO_ENABLED=0` and `-tags nobdk`. `go vet` is clean in all three.
- **Full suite on each backend:**
  - `go test ./...` with cgo, i.e. on GoBDK: 63 packages ok.
  - `go test -tags nobdk ./...`, i.e. on the built-in verifier: 63 packages ok.
  - `CGO_ENABLED=0 go test ./...` fails in the SQLite-backed packages because `go-sqlite3` requires cgo. The same packages fail on `main`; this is not caused by this change.
- **`pkg/internal/signaturebackend` tests,** run in each build mode:
  - the reported backend matches the build
  - a valid P2PKH spend verifies and a tampered one is rejected
  - in cgo builds, GoBDK signatures verify under go-sdk's built-in verifier
- **Activation through `pkg/storage`:** a program importing only `pkg/storage` showed GoBDK C calls (`_Cfunc_secp256k1_ecdsa_verify`) in its CPU profile when built with cgo, and none without.
- **`provider_label_query_mode_test.go`** covers labels (with and without `IncludeOutputs`) and tags:
  - one, two and three names, in either order
  - a repeated name, an unknown name
  - a label owned by another user
  - a soft-deleted label or tag
  - "any" mode
  - totals with paging
  - Every case except "repeated name" also passed on the old code.
- The label/tag and existing BRC-114 label tests pass with `TEST_DB_MODE=postgres` against Postgres 17.
- **Mutation check:** making "all" behave like "any" fails 10 cases of the new tests.
- **`known_tx_get_beef_trust_self_test.go`** records the SQL issued, and checks:
  - a TrustSelf build over a 6-deep unproven chain issues one existence read and no blob reads
  - `KnownTxIDs` subjects issue no reads
  - an unknown subject still resolves via services
  - a full (non-TrustSelf) build still reads `raw_tx`
- **`wallet_sign_action_beef_completeness_test.go`:**
  - The signed BEEF returned by `SignAction` is valid with no txid-only entries. This guards against trimming sign-action ancestry by default; that change was tried during this work and broke exactly this.
  - With the `TrustSelf` + `ReturnTXIDOnly` opt-in, the signable tx still gives a sighash for every input, and the action signs and broadcasts.
  - Spending the wallet's own output as a caller-supplied input with no input BEEF works across 12 unconfirmed generations with delayed broadcast, deeper than the 8-round hydration bound.
- Repeated 1000-trade load tests, with the application built with cgo: 0 failed trades, and `storage.SignatureBackend()` reported `gobdk-libsecp256k1`.
- `golangci-lint` was **not** run locally (installed binary is built with go1.26; the module targets 1.27).

## Impact / Risk
- **Default change for cgo builds:** any application that imports `pkg/storage` or `pkg/wallet` and builds with cgo on darwin/linux amd64/arm64 now signs and verifies through GoBDK, process-wide.
  - cgo is on by default for native builds when a C compiler is present.
  - Such binaries link `libstdc++` dynamically, so they won't start on images without it (`scratch`, distroless static, Alpine without the package).
  - To keep the previous behaviour, build with `CGO_ENABLED=0` or `-tags nobdk`.
  - This repository's own Dockerfiles and goreleaser build with `CGO_ENABLED=0`, so its images and releases are unchanged.
- **Signature compatibility:** GoBDK signatures verify under go-sdk's built-in verifier (tested), and the two backends' verdicts matched across the differential and mutation runs.
- **Dependency:** `github.com/bitcoin-sv/bdk/module/gobdk` (with prebuilt native libraries) is now part of the module graph for all consumers.
- **Behaviour change:** in "all" mode a repeated name (`x, y, y`) now matches transactions carrying `x` and `y`. Before, it matched nothing, because `COUNT(DISTINCT) = 3` can't be satisfied. This makes it consistent with `ListTransactions`, which already used one `EXISTS` per label.
- **TrustSelf pre-fetch:** it only affects callers that explicitly opt into `TrustSelf=known` for a BEEF build. The BEEF produced is the same, because found rows were already stubbed without their columns being used, and missing rows still fall back to services. Non-TrustSelf builds are untouched.
- **Query shape:** relies on the existing `idx_transaction_labels_name_user` / `idx_output_tags_name_user` indexes; no schema or migration changes. Postgres plans the new form as index nested loops. A request with many names adds one subquery per name.
- **API:** adds `storage.SignatureBackend()` and the `nobdk` build tag; no other API or config changes.


dzolt-4chain (approved):

pkg/storage/signature_backend.go
```diff
@@ -0,0 +1,12 @@
...
+import "github.com/bsv-blockchain/go-wallet-toolbox/pkg/internal/signaturebackend"
+
+// SignatureBackend names the secp256k1 implementation go-sdk uses in this binary:
+// "gobdk-libsecp256k1" when it was built with cgo on darwin or linux, amd64 or
+// arm64, and "go-sdk-builtin" otherwise or with -tags nobdk. The choice is made
+// at build time and applies process-wide; log it at startup to confirm a
+// deployment got the build it expects.
+func SignatureBackend() string {
+	return signaturebackend.Name
+}
```
kuba-4chain, line 12: This file can be removed, unused anywhere

mrz1836 (approved):
AI-TLDR; LGTM

pkg/storage/signature_backend.go
kuba-4chain: EDIT: it's exported and used by the client. Resolving...

kuba-4chain (approved):

